PR #27, opened 2014-12-27
Xep 0122

annovanvliet:
Support for XEP-0122

Added extensions objects and added consistency checks.


Flowdalic:
Merged into master

annovanvliet:
Excuse for the late response, but I have one comment on your modifications.

The XEP explicitly states that if no validation method is specified, form
processors MUST assume <basic/> validation. I had solved this by creating a
class DefaultValidateElement extending BasicValidateElement.I noticed you
have replaced it by an empty class without any functionality.

I had put this inner class in the provider, while it (i.m.o.) only has a
function within the context of parsing the xml.



2015-01-04 0:39 GMT+01:00 Florian Schmaus <notifications@github.com>:

Flowdalic:
Good point. For some reason I thought that EmptyValidateElement has any value. I suggest https://github.com/Flowdalic/Smack/commit/810a12acfca30ea12e9e7e09591021e0589501cc

Do you agree?

annovanvliet:
This solution is correct according to the specs, but I did override the
toXml of  BasicValidateElement in order to make sure that there would be
the same output as the input. See the test you had to adjust.
 Op 4 jan. 2015 22:25 schreef "Florian Schmaus" <notifications@github.com>:

PR #33, opened 2015-02-03
Refactored the vcard implementation to use a pull parser.

deeringc:
I haven't written any Java in a long time, so let me know if anything needs to change.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
@@ -18,339 +18,414 @@ package org.jivesoftware.smackx.vcardtemp.provider;
 
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.provider.IQProvider;
-import org.jivesoftware.smack.util.StringUtils;
 import org.jivesoftware.smackx.vcardtemp.packet.VCard;
-import org.w3c.dom.Document;
-import org.w3c.dom.Element;
-import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
-import org.w3c.dom.Text;
-import org.xml.sax.SAXException;
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
 
-import javax.xml.XMLConstants;
-import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
-
-import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
  * vCard provider.
- *
+ * 
  * @author Gaston Dombiak
  * @author Derek DeMoro
+ * @author Chris Deering
  */
+
 public class VCardProvider extends IQProvider<VCard> {
     private static final Logger LOGGER = Logger.getLogger(VCardProvider.class.getName());
-    
-    private static final String PREFERRED_ENCODING = "UTF-8";
 
-    private static final DocumentBuilderFactory DOCUMENT_BUILDER_FACTORY;
+    private enum Adr {
+        POSTAL,
+        PARCEL,
+        DOM,
+        INTL,
+        PREF,
+        POBOX,
+        EXTADR,
+        STREET,
+        LOCALITY,
+        REGION,
+        PCODE,
+        CTRY,
+        FF
+    }
 
-    static {
-        DocumentBuilderFactory documentBuilderFactory = DocumentBuilderFactory.newInstance();
-        try {
-            documentBuilderFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
-        } catch (ParserConfigurationException e) {
-            LOGGER.finer("Could not disallow doctype decl: " + e.getMessage());
-            // If we can't disable DTDs, then at least try the following
-            // Xerces 1 - http://xerces.apache.org/xerces-j/features.html#external-general-entities
-            // Xerces 2 - http://xerces.apache.org/xerces2-j/features.html#external-general-entities
-            try {
-                documentBuilderFactory.setFeature( "http://xml.org/sax/features/external-general-entities", false);
-                documentBuilderFactory.setFeature( "http://xml.org/sax/features/external-parameter-entities", false);
-            } catch (ParserConfigurationException e1) {
-                LOGGER.finer("Could not disallow external entities for xerces parser: " + e1.getMessage());
-            }
-        }
-        // Android throws an UnsupportedOperationException when calling setXIncludeAware() and for some dumb reason also
-        // when calling isXIncludeWare(), while it defaults according to the docs to 'false'.
-        boolean isXIncludeAware;
-        try {
-            isXIncludeAware = documentBuilderFactory.isXIncludeAware();
-        } catch (UnsupportedOperationException e) {
-            // Assume we are on Android where isXIncludeAware defaults to 'false'
-            isXIncludeAware = false;
-        }
-        if (isXIncludeAware) {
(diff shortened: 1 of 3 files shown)


smack-extensions/src/test/java/org/jivesoftware/smackx/VCardUnitTest.java
```diff
@@ -35,40 +37,222 @@ public void initSmackTestSuite() {
     }
 
     @Test
+    public void testParseFullVCardIQStanza() throws Throwable {
+
+        // @formatter:off
+        final String request =
+        "<iq id='v1' to='user@igniterealtime.org/mobile' type='result'>"
```
deeringc: One thing I'm not clear about is whether this XEP is case sensitive. Currently all of the tags are matched only in their upper case form, which is how they appear in the spec. Should I refactor to accept them in a case insensitive way?

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+                    }
+
+                    depth++;
                 }
+                else if (eventType == XmlPullParser.END_TAG) {
+
+                    depth--;
 
-                if (event == XmlPullParser.END_TAG && "vCard".equals(parser.getName())) break;
+                    if (lastEventWasStart && 0 == depth && name.length() != 0 && text.length() != 0) {
```
deeringc: This bit is so that we capture "unknown" top level elements that just contain text, and add them as a key, value pair. I believe this was the original behaviour.

Flowdalic:
I've written some dev documentation about smack provider design today and published it at https://github.com/Flowdalic/Smack/blob/master/documentation/developer/provider.md
you may want to take a look

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
 
         try {
-            documentBuilderFactory.setFeature(XMLConstants.FEATURE_SECURE_PROCESSING, true);
-        } catch (ParserConfigurationException e) {
-            LOGGER.info("Could not enable secure processing parsing feature: " + e.getMessage());
-        }
 
-        DOCUMENT_BUILDER_FACTORY = documentBuilderFactory;
-    }
+            boolean done = false;
```
Flowdalic: please use labeled loops instead (see providers.md)

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+            boolean lastEventWasStart = false;
+            int depth = 0;
```
Flowdalic: XmlPullParser already maintains a depth (e.g. `parser.getDepth()`)

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+            String name = "";
```
Flowdalic: Those values should usually be `null` initialized. Otherwise you can't tell the difference between a missing attribute value/text or if it's set to the empty string.

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
-                        sb.append('<').append(parser.getName()).append('>');
-                        break;
-                    case XmlPullParser.END_TAG:
-                        sb.append("</").append(parser.getName()).append('>');
-                        break;
-                    default:
+            while (!done) {
+                int eventType = parser.next();
+
+                if (eventType == XmlPullParser.START_TAG) {
```
Flowdalic: Please use switch/case instead of if/if-else cascades

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+
+                    lastEventWasStart = true;
+
+                    name = parser.getName();
+
+                    if (name.equals("N")) {
```
Flowdalic: Java7 allows switch/case on Strings, please use it here.

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+
+                        vCard.setField(name, text);
+                    }
 
-                event = parser.next();
+                    name = "";
+                    text = "";
+                    lastEventWasStart = false;
+
+                    if (parser.getName().equals("vCard")) {
```
Flowdalic: Abort condition should only consider the current parser depth (see providers.md)

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
-        private final VCard vCard;
-        private final Document document;
+    private void parseAddress(XmlPullParser parser, VCard vCard) throws XmlPullParserException, IOException {
 
-        VCardReader(VCard vCard, Document document) {
-            this.vCard = vCard;
-            this.document = document;
-        }
+        boolean isWork = true;
+        boolean done = false;
```
Flowdalic: use labeled loop instead

smack-extensions/src/main/java/org/jivesoftware/smackx/vcardtemp/provider/VCardProvider.java
```diff
@@ -18,339 +18,335 @@
...
+            int eventType = parser.next();
 
-        private void appendText(StringBuilder result, Node node) {
-            NodeList childNodes = node.getChildNodes();
-            for (int i = 0; i < childNodes.getLength(); i++) {
-                Node nd = childNodes.item(i);
-                String nodeValue = nd.getNodeValue();
-                if (nodeValue != null) {
-                    result.append(nodeValue);
+            if (eventType == XmlPullParser.START_TAG && parser.getName().equals("BINVAL")) {
```
Flowdalic: 2x switch/case here

smack-extensions/src/test/java/org/jivesoftware/smackx/VCardUnitTest.java
Flowdalic: Did it previously accept case insensitive vcards? If not, then I say we leave it as it is. It appears to me that XMPP is case sensitive just as XML is. And vcard is the only XEP I'm aware of that uses uppercase element names.

smack-extensions/src/test/java/org/jivesoftware/smackx/VCardUnitTest.java
```diff
@@ -35,40 +37,222 @@ public void initSmackTestSuite() {
...
+        // @formatter:off
+        final String request =
+                        "<iq id='v1' to='user@igniterealtime.org/mobile' type='result'>"
+                        + "<vCard xmlns='vcard-temp'><EMAIL><USERID>foo@fee.www.bar</USERID></EMAIL></vCard>"
+                        + "</iq>";
+        // @formatter:on
+
+        IQ iqRequest = (IQ) PacketParserUtils.parseStanza(request);
+        assertTrue(iqRequest instanceof VCard);
+        VCard vCard = (VCard) iqRequest;
```
Flowdalic: `VCard vCard = PakcetParserUtils.parseStanza(request)` should be enough

Flowdalic:
Cherry-picked with minor modifications as https://github.com/Flowdalic/Smack/commit/1e0f697aea5455cff2a6d3c8a5a541c0c227770a
Hint: use `git diff --word-diff` to let you show the modifications I've made.

Thanks for your contribution. I'm glad that the DOM based parsing is gone.

PR #34, opened 2015-03-11
Add roomDestroyed to MUC UserStatusListener

meisterfuu:
I could not test it yet, but it is pretty simple and looks ok to me.

### DIFF
--- a/version.gradle
+++ b/version.gradle
@@ -1,7 +1,7 @@
 allprojects {
 	ext {
-		shortVersion = '4.2.0-alpha1'
-		isSnapshot = false
+		shortVersion = '4.2.0-alpha2'
+		isSnapshot = true
 		jxmppVersion = '0.5.0-alpha1'
 		smackMinAndroidSdk = 8
 


meisterfuu:
error: method roomDestroyed in interface UserStatusListener cannot be applied to given types;
                listener.roomDestroyed(mucUser.getDestroy().getJid(), mucUser.getDestroy().getReason());
                        ^
  required: Jid,String
  found: String,String

Fixing this now

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DefaultUserStatusListener.java
```diff
@@ -65,5 +65,8 @@ public void adminGranted() {
 
     public void adminRevoked() {
     }
+    
+    public void roomDestroyed(Jid alternateJID, String reason) {
```
Flowdalic: Why is the parmater name `alternateJID`?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -2181,6 +2182,23 @@ private void checkPresenceCode(
                 listener.nicknameChanged(from, mucUser.getItem().getNick());
             }
         }
+        //The room has been destroyed
+        if (mucUser.getDestroy() != null) {
+            Jid alternateJid = null;
```
Flowdalic: Why not simply

    alternateJid = mucUser.getDestroy().getJid()

?

Flowdalic:
Thanks for you contribution. Please

- squash the commits into one and update this PR
- mention the issue key SMACK-619 somewhere in the body of the commit message

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DefaultUserStatusListener.java
meisterfuu: Same as here:
MultiUserChat.destroy(String reason, String alternateJID)

Flowdalic: Ahh, makes sense then. And I see it's javadoc'ed at the listener, so it's fine.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
meisterfuu: mucUser.getDestroy().getJid() returns a string. I can change that later if you want the parsing done in some other place.

meisterfuu:
Ok, I update the PR later when I get home.
Commiting with the github editor was not the best thing i ever did.^^

meisterfuu:
Squashed the commits

Flowdalic:
I've just pushed https://github.com/igniterealtime/Smack/commit/516e39767994e8cf4a1b21caef6a35ca5bcc6d16

Could you rebase you commit on it, so that we can get rid of the jid parsing code?

Flowdalic:
Also please change

    public abstract void roomDestroyed(Jid alternateJID, String reason);

to

    public abstract void roomDestroyed(MultiUserChat alternativeMuc, String reason);

meisterfuu:
I will do it when I get home later.

meisterfuu:
https://github.com/igniterealtime/Smack/pull/36

Flowdalic:
FYI, no need to close a PR. You could simply update the existing one by force-pushing to its source branch.

meisterfuu:
I messed with the master and wanted a clean clone from the smack master without a "Merge commit". I rarely do other things than Push/Pull with git, still learning :)

PR #36, opened 2015-03-16
Add roomDestroyed to MUC UserStatusListener

meisterfuu:
SMACK-619

New PR(and fork) because work computer messed with the git usernames....

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -2179,6 +2179,18 @@ public class MultiUserChat {
                 listener.nicknameChanged(from, mucUser.getItem().getNick());
             }
         }
+        //The room has been destroyed
+        if (mucUser.getDestroy() != null) {
+            MultiUserChat alternateMUC = multiUserChatManager.getMultiUserChat(mucUser.getDestroy().getJid());
+            for (UserStatusListener listener : userStatusListeners) {
+                listener.roomDestroyed(alternateMUC, mucUser.getDestroy().getReason());
+            }
+
+            // Reset occupant information.
+            occupantsMap.clear();
+            nickname = null;
+            userHasLeft();
+        }
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -2179,6 +2179,18 @@ private void checkPresenceCode(
                 listener.nicknameChanged(from, mucUser.getItem().getNick());
             }
         }
+        //The room has been destroyed
+        if (mucUser.getDestroy() != null) {
+        	MultiUserChat alternateMUC = multiUserChatManager.getMultiUserChat(mucUser.getDestroy().getJid())
```
Flowdalic: looks like a whitespace issue here

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/UserStatusListener.java
```diff
@@ -123,4 +123,12 @@
      */
     public abstract void adminRevoked();
 
+    /**
+     * Called when the room is destroyed.
+     * 
+     * @param alternateMUC an alternate MultiUserChat.
```
Flowdalic: IIRC alternateMUC and reason are both optional. If so, it should be mentioned in the javadoc that they may be null.

meisterfuu:
The coverage decrease seems kinda random to me?!

Flowdalic:
Well, you didn't write a unit test for the `mucUser.getDestroy() != null` condition, so when you add code that isn't covered by the unit test, the overall test coverage decreases. But although it would be nice to have a unit test for that case, not having one is not a show stopper.

What's your JIRA username (so I can assign the issue to you)?

meisterfuu:
I will create an account called "meisterfuu" later. 
Ok, but why does the output say something about "/smack-core/src/main/java/org/jivesoftware/smack/util/ArrayBlockingQueueWithShutdown.java"?

Flowdalic:
Cherry picked with minor modifications as https://github.com/igniterealtime/Smack/commit/05c97c494bf1b5c122c2ef0a7cfdb60a8f3ff828

meisterfuu:
JIRA username is "meisterfuu"

PR #42, opened 2015-03-24
Resurrect integration tests: Chat Test

stawny:


smack-integration-test/src/main/java/org/jivesoftware/smack/ChatTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public ChatTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+        chatManagerOne = ChatManager.getInstanceFor(conOne);
+    }
+
+    @SmackIntegrationTest
+    public void testProperties() throws XmppStringprepException, NotConnectedException, Exception {
+        Chat newChat = chatManagerOne.createChat(conTwo.getUser());
+        PacketCollector collector = conTwo.createPacketCollector(new ThreadFilter(newChat.getThreadID()));
+        JivePropertiesManager.setJavaObjectEnabled(true);
```
Flowdalic: Tests should leave the state unchanged, so if you set this to true, then you need to reset it at the end. Create two methods `setUp()` and `tearDown()`, annoate them with @BeforeClass and @AfterClass, to set and reset the config of JivePropertiesManager.

smack-integration-test/src/main/java/org/jivesoftware/smack/ChatManagerListenerTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jivesoftware.smack.chat.Chat;
+import org.jivesoftware.smack.chat.ChatManager;
+import org.jivesoftware.smack.chat.ChatManagerListener;
+import org.junit.Assert;
+
+
+/**
+ * @author Stawicki Adam
+ */
+public class ChatManagerListenerTest extends AbstractSmackIntegrationTest {
```
Flowdalic: I guess there is no need to put this in an extra class. Just put it in the chat manger listener.

smack-integration-test/src/main/java/org/jivesoftware/smack/ChatManagerListenerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    
+    private final ChatManager chatManagerOne;
+    
+    private boolean invoked;
+
+    public ChatManagerListenerTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+        chatManagerOne = ChatManager.getInstanceFor(conOne);
+        chatManagerOne.addChatListener(new TestChatManagerListener());
```
Flowdalic: should be added in the test method, and eventually removed at the end of that method.

smack-integration-test/src/main/java/org/jivesoftware/smack/ChatManagerListenerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public void testccc() {
+        chatManagerOne.createChat(conTwo.getUser());
+        Assert.assertTrue("TestChatManagerListener wasn't invoked", invoked);
+    }
+    
+    public class TestChatManagerListener implements ChatManagerListener {
+
+        @Override
+        public void chatCreated(Chat chat, boolean createdLocally) {
+            System.out.println("creating!!!!!!!!!\n\n\n\n\n");
```
Flowdalic: never use println in a library or unit tests. but checkstyle already told you so.

smack-integration-test/src/main/java/org/jivesoftware/smack/ChatManagerListenerTest.java
```diff
@@ -0,0 +1,58 @@
...
+        chatManagerOne.addChatListener(new TestChatManagerListener());
+    }
+    
+    @SmackIntegrationTest
+    public void testccc() {
+        chatManagerOne.createChat(conTwo.getUser());
+        Assert.assertTrue("TestChatManagerListener wasn't invoked", invoked);
+    }
+    
+    public class TestChatManagerListener implements ChatManagerListener {
```
Flowdalic: I guess this could be a anonymous class within the method

smack-integration-test/src/main/java/org/jivesoftware/smack/ChatManagerListenerTest.java
```diff
@@ -0,0 +1,58 @@
...
+    private boolean invoked;
+
+    public ChatManagerListenerTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+        chatManagerOne = ChatManager.getInstanceFor(conOne);
+        chatManagerOne.addChatListener(new TestChatManagerListener());
+    }
+    
+    @SmackIntegrationTest
+    public void testccc() {
```
Flowdalic: please give test method meaningful names, as they are displayed when the test framework runs.

Flowdalic:
Picked as 9d0ea3cf70506bef7fcd0e83d5abd71a75661040

PR #43, opened 2015-04-04
Fix invoking beforeClass and afterClass method

stawny:


Flowdalic:
Picked as 75e35e8db62d508f5dc6dd353b8c736d5ad87252

PR #45, opened 2015-06-02
Better cope with clients with bad extensions or bad priorities.

felixdoerre:
There are clients out there that send strange XML, and Servers that relay such strange XML. The library currently crashes the whole XML-stream when such malformed packages occur. This is a very bad behaviour.

The examples I recieved, and for which I provide a patch here are:
* malformed priorities `<presence ...><priority>200</priority>..` or
  empty priorities `<presence ...><priority></priority>..`
* malformed extensions that cause the parser to crash.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.compress.packet.Compress;
 import org.jivesoftware.smack.packet.DefaultExtensionElement;
 import org.jivesoftware.smack.packet.EmptyResultIQ;
@@ -543,9 +544,23 @@ public class PacketParserUtils {
                 case "status":
                     presence.setStatus(parser.nextText());
                     break;
-                case "priority":
-                    int priority = Integer.parseInt(parser.nextText());
-                    presence.setPriority(priority);
+                case "priority" :
+                    try {
+                        int priority = Integer.parseInt(parser
+                                .nextText());
+                        presence.setPriority(priority);
+                    } catch (Throwable e) {
+                        // Some implementations send presence stanzas with a
+                        // '<priority />' element or an invalid presence element, which is a invalid XMPP presence
+                        // stanza according to RFC 6121 4.7.2.3
+                        LOGGER.log(
+                                Level.WARNING,
+                                "Empty, null or malformed text in presence priority element form "
+                                        + presence.getFrom()
+                                        + " with id '"
+                                        + presence.getStanzaId()
+                                        + "' which is invalid.", e);
+                    }
                     break;
                 case "show":
                     String modeText = parser.nextText();
@@ -909,7 +924,17 @@ public class PacketParserUtils {
         // See if a provider is registered to handle the extension.
         ExtensionElementProvider<ExtensionElement> provider = ProviderManager.getExtensionProvider(elementName, namespace);
         if (provider != null) {
+            try {
                 return provider.parse(parser);
+            } catch (IOException e) {
+                throw e;
+            } catch (XmlPullParserException e) {
+                throw e;
+            } catch (SmackException e) {
+                throw e;
+            } catch (Throwable e) {
+                throw new SmackException(e);
+            }
         }
 
         final int initialDepth = parser.getDepth();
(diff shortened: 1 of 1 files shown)


Flowdalic:
Hi Felix, thanks for you contribution, but

RFC 6120 mandates that stream-level errors are non-recoverable (4.9.1.1) and the connection should be terminated after such an error. In particular, XML parse errors are stream-level errors.

Furthermore ignoring malformed extension or data is the worst thing you can do when it comes to a network library. The library should not do it in the way as suggested in your commit. The only way to ignore such errors Smack will *ever* provide is the already existing parsing exception callback, which ignores the whole faulty stanza if the callback does not throw (and thus violates RFC 6120 when used).

felixdoerre:
I am not sure, if RFC 6120 (4.9.1.1) refers to such errors, because these are not XML-errors, but XML-content errors. When the XML-stream is malformed I am totally with you.

For presence stanzas I see your point.

But faulty "extensions" should not cause the hole package to be invalid. These are extensions, and a broken extension in a message should not make the whole message unparsable. Clients should at least be able to display the plain message instead of totally discarding the message. (And instead of closing the stream). Why should this behaviour be bad?

Flowdalic:
What if the extension carried a semantic, e.g. a security label? You can't assume that everything will be right and the stream still is consistent if you simply ignore a extension that threw an exception on parsing.

felixdoerre:
I know, _in general_ catching `Throwable` is absolutely ugly, but when invoking "Plugins" or "Extensions" that are meant to be isolated from the core code, this could be helpful to increase that isolation.

My problem is: A jabber server sends slightly broken offline messages:
```
<message ... type='chat'>
 <body>...</body>
 <delay xmlns='urn:xmpp:delay' stamp='2015-06-04T21:38:16.302+00:00' ...></delay>
</message>
```
These messages cause Smack to fail and crash the connection. So, when I logon, I am directly disconnected. Luckily, such messages disappear after a few reconnects. But that way, I loose messages. Ok, with the ErrorCallback, the connection can be kept open, but the problem persists: The messages are still lost. That should _never_ happen. I do not want to loose messages. How can I use Smack to achieve this?

Regarding the security labels: what is the behaviour of clients, not implementing this extension? They ignore it! So why should a client that is not able to understand the extension behave differently? Any standard specifying things such as those security extensions, has to cope with clients that do not understand the extension. When an extension requires such a behaviour, any client that implements this extension has to cope with it in that way that the extension will always be "parsed" but that be internally flagged as "malformed", so malformed "security labels" can cause the message to be ignored.

Flowdalic:
> I know, in general catching Throwable is absolutely ugly, but when invoking "Plugins" or "Extensions" that are meant to be isolated from the core code, this could be helpful to increase that isolation.

Absolutely not. Catching Throwable is always bad. There are really no exceptions. The most generic a catch should get is catching all **checked** exceptions.

> Regarding the security labels: what is the behaviour of clients, not implementing this extension?

This was just meant as an example. If the extension contains semantic, and you remove the exception then you change the semantic of the stanza, which could lead to unforeseen consequences.

felixdoerre:
Am 05.06.2015 um 07:29 schrieb Florian Schmaus:
>
>     I know, in general catching Throwable is absolutely ugly, but when
>     invoking "Plugins" or "Extensions" that are meant to be isolated
>     from the core code, this could be helpful to increase that isolation.
>
> Absolutely not. Catching Throwable is always bad. There are really no 
> exceptions. The most generic a catch should get is catching all 
> *checked* exceptions.
>
Ok, maybe Throwable is too strong. But catching "Exception" (which also 
catches unckeched RuntimeExceptions), should be used when invoking 
plug-ins (in fact, two of the three references catch Throwable):

Eclipse:
http://git.eclipse.org/c/equinox/rt.equinox.bundles.git/tree/bundles/org.eclipse.equinox.common/src/org/eclipse/core/runtime/SafeRunner.java#n39

Openfire:
https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/container/PluginManager.java#L285
https://github.com/igniterealtime/Openfire/blob/master/src/java/org/jivesoftware/openfire/container/PluginManager.java#L535

Jenkins:
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/ClassicPluginStrategy.java#L437

So in all that applications, plug-ins cannot easily crash the main 
Application.

Also [RFC 6120 Section 
8.3](http://tools.ietf.org/html/rfc6120#section-8.3) suggests treating 
so-called "Stanza Errors" non-fatal.
The processing of extension elements is then covered in [Section 
8.4](http://tools.ietf.org/html/rfc6120#section-8.4):
"If a recipient receives a stanza that contains an element or attribute 
it does not understand, it MUST NOT attempt to process that XML data and 
instead MUST proceed as follows. " (Page 127 Bottom) I would call a 
parsing error as "can not understand".

Please consider following the "Robustness Principle" that is generally 
true for implementing protocols 
(http://tools.ietf.org/html/rfc761#section-2.10): "be conservative in 
what you do, be liberal in what you accept from others."

>     Regarding the security labels: what is the behaviour of clients,
>     not implementing this extension?
>
> This was just meant as an example. If the extension contains semantic, 
> and you remove the exception then you change the semantic of the 
> stanza, which could lead to unforeseen consequences.
>
What is the difference in doing that from clients, who don't know the 
extension at all? Aren't these the same consequences? Should't any 
extension look for exactly that consequences always (because it has to 
cope with client that don't know it) ?

Extensions that depend on such a feature should ensure that the parsing 
will not crash.

Flowdalic:
> I would call a parsing error as "can not understand".

I would not call them the same. The RFC clearly means with "not understand" that the entity "is not aware of the existence of the extension".

> Please consider following the "Robustness Principle"

The Postel's Law, also called the "Robustness Principle" has never been a good idea. See for example http://tools.ietf.org/html/rfc3117#section-4.5

> Aren't these the same consequences? 

No, they are not.

PR #46, opened 2015-06-20
XmppHostnameVerifier: check subject CommonName

huguesb:
Not all certs come with valid subjectAlternativeNames

### DIFF
--- a/smack-java7/src/main/java/org/jivesoftware/smack/java7/XmppHostnameVerifier.java
+++ b/smack-java7/src/main/java/org/jivesoftware/smack/java7/XmppHostnameVerifier.java
@@ -30,6 +30,9 @@ import java.util.Locale;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.naming.InvalidNameException;
+import javax.naming.ldap.LdapName;
+import javax.naming.ldap.Rdn;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLPeerUnverifiedException;
 import javax.net.ssl.SSLSession;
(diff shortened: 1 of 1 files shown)


smack-java7/src/main/java/org/jivesoftware/smack/java7/XmppHostnameVerifier.java
```diff
@@ -133,7 +136,20 @@ private static void matchDns(String name, X509Certificate cert) throws Certifica
...
+            LdapName dn = new LdapName(cert.getSubjectX500Principal().getName());
+            for (Rdn rdn : dn.getRdns()) {
+                if (rdn.getType().equalsIgnoreCase("CN")) {
+                    if (matchesPerRfc2818(name, rdn.getValue().toString())) {
+                        return;
+                    }
+                    break;
+                }
+            }
+        } catch (Exception e) {
```
Flowdalic: what does throw here?

huguesb: LdapName ctor may throw InvalidNameException. Should the catch be more specific or the exception propagated upwards?

Flowdalic: Ahh, I see. Please make the catch more specific and restrict the try block to only the places where the throw is expected to occur. I'm also not sure if the log level should be warning, all other logs are just 'info'. But maybe a warning is justified in this case.

Flowdalic:
Applied as https://github.com/igniterealtime/Smack/commit/f24ef7dda918c182924f536cf2859a9fa551b575

PR #47, opened 2015-06-21
HOXT: Check headers and data in IQChildElementXmlStringBuilder + Update documentation

Tibo-lg:
* `NullPointerException` are raised when no headers or no data are specified in a request. Also if no version is specified but I don't know if that is intended behavior.
* The documentation of HOXT specifies `Set` for creating headers but the `HeadersExtension` constructor takes a `List`.


Flowdalic:
Thanks for your contribution. Travis failed because of checkstyle validation failed. Please correct those errors. And please also squash b75a and 9a1c into one commit. f44 should use `optAppend()` instead.

PR #48, opened 2015-06-22
HOXT: Fix getIQChildElementBuilder in AbstractHttpOverXmpp.java + Update documentation

Tibo-lg:
* `NullPointerException` was raised when not setting any header or data in an  HttpOverXmppReq or Resp. Fixed by using `optAppend` instead of `append`. 
* The documentation used a `Set` for storing the head bu this is not accepted by `HeadersExtension` constructor.
* Error occurs if no version in set in a Req or Res. This is not fixed as I am not sure what should be the correct behavior. Two possibilities:
  * Mandatory to specify version, then an exception should probably be raised when it is not done.
  * Version number is set by default to "1.1" if not provided. In that case should the Host header be set by default as well?   

PS: Sorry I messed up the previous request trying to squash the commits.

### DIFF
--- a/documentation/extensions/hoxt.md
+++ b/documentation/extensions/hoxt.md
@@ -73,11 +73,11 @@ HttpOverXmppReq req = new HttpOverXmppReq(HttpMethod.POST, "/mailbox");
 req.setVersion("1.1");
 
 // prepare headers
-Set<Header> set = new HashSet<Header>();
-set.add(new Header("Host", "juliet.capulet.com"));
-set.add(new Header("Content-Type", "application/x-www-form- urlencoded"));
-set.add(new Header("Content-Length", Integer.toString(urlEncodedMessage.length())));
-req.setHeaders(new HeadersExtension(set));
+List<Header> list = new ArrayList<Header>();
+list.add(new Header("Host", "juliet.capulet.com"));
+list.add(new Header("Content-Type", "application/x-www-form- urlencoded"));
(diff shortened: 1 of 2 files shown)


Flowdalic:
> This is not fixed as I am not sure what should be the correct behavior.

That's a design flaw in Smack's HOXT API: Its classes where designed as Java Beans with getters and setters. Whereas it should use a builder and create immutable instances (Which is what new Smack code should do). The build() operation should fail if mandatory fields are not set to correct values (like 'version' in this case). Having a default value of "1.1" sounds ok in this case.

> PS: Sorry I messed up the previous request trying to squash the commits.

FYI, there was no need to create a new PR. You could have simply update the existing one by doing a force-push to its source branch.

Flowdalic:
Applied as https://github.com/igniterealtime/Smack/commit/e20c17050eea75864b32862cb301be2a8b9f2d3f and https://github.com/igniterealtime/Smack/commit/3ff1ab6e1f3f5ff1ec5e2693acc26b527c14803e

Tibo-lg:
>That's a design flaw in Smack's HOXT API: Its classes where designed as Java Beans with getters and setters. Whereas it should use a builder and create immutable instances (Which is what new Smack code should do). The build() operation should fail if mandatory fields are not set to correct values (like 'version' in this case). Having a default value of "1.1" sounds ok in this case.

I will try to update it but it may take some time.

>FYI, there was no need to create a new PR. You could have simply update the existing one by doing a force-push to its source branch.

Thanks for the tips!

PR #49, opened 2015-06-22
SASL Proxy Auth support

dwd:
This adds the ability to provide a distinct authorization identifier for use
by SASL mechanisms. Not all SASL mechanisms support this operation, in
particular CRAM-MD5.

Both the javax and provided SASL implementations are extended, and an authzid
parameter added to the authenticate method. Alternate methods are provided to
avoid API change.

The authorization identifier is passed as a BareJid in order to assure the
correct form.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
@@ -31,6 +31,7 @@ import org.jivesoftware.smack.util.CollectionUtil;
 import org.jivesoftware.smack.util.Objects;
 import org.jivesoftware.smack.util.StringUtils;
 import org.jxmpp.jid.DomainBareJid;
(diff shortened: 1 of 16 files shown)


Flowdalic:
Hi Dave, thanks for your contribution. :)

I had just a quick lock, but would you be so kind and:
- Fix the unit test
- Add the issue key 'SMACK-677' somewhere in the commit messages body
- Introduce a ConnectionConfiguration setting for authzid, (`SASLAuthentication.authenticate()` is not meant a part of Smack's public API)
- Remove authzidSupported() and throw a UnsupportedOperationException instead when used with CRAM

There's possible more to follow, I'll eventually contact you in IRC.

Flowdalic:
Closed in favor of https://github.com/Flowdalic/Smack/commit/9c772add9369b654f0bd65cd96b076dec7053cc5

Thanks for your contribution Dave.

PR #50, opened 2015-06-28
Update HOXT API to use the builder pattern

Tibo-lg:
I tried to implement the builder pattern for the HOXT API:
* The `AbstractHttpOverXmpp`, `HttpOverXmppReq` and `HttpOverXmppResp` implement the builder pattern in the same style as `ConnectionConfiguration` and `XMPPTCPConnectionConfiguration`.
* The providers had to be refactored to make use of the builders; I had to separate the parsing of headers and data.
* I realized my previous patch did not solve the `NullPointerException` issue. I updated so that `headers` is given directly to `optAppend` but `data` does not extend `Element` so it needs an extra `if`.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/hoxt/provider/AbstractHttpOverXmppProvider.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/hoxt/provider/AbstractHttpOverXmppProvider.java
@@ -49,73 +49,78 @@ public abstract class AbstractHttpOverXmppProvider<H extends AbstractHttpOverXmp
     static final String ATTRIBUTE_VERSION = "version";
 
     /**
-     * Parses Headers and Data elements.
+     * Parses HeadersExtension element if any.
      *
-     * @param parser      parser
-     * @param elementName name of concrete implementation of this element
-     * @param body        parent Body element
-     * @throws Exception 
+     * @param parser parser
+     * @return HeadersExtension or null if no headers
+     * @throws Exception
      */
-    protected void parseHeadersAndData(XmlPullParser parser, String elementName, AbstractHttpOverXmpp body) throws Exception {
-        boolean done = false;
-
-        while (!done) {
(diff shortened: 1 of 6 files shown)


Flowdalic:
Thanks for your contribution. I really appreciate your work and effort towards working together on a HOXT API. +1

I hadn't had a close look at the changes, But could you please
- remove the merge commit (i.e. your work should be (usually) a single commit)
- fix the build errors on java 8 (see travis)

smack-experimental/src/main/java/org/jivesoftware/smackx/hoxt/provider/AbstractHttpOverXmppProvider.java
```diff
@@ -49,73 +49,71 @@
...
-                    throw new UnsupportedOperationException("jingle is not supported yet");
-                } else {
-                    // other elements are not allowed
-                    throw new IllegalArgumentException("unsupported child tag: " + parser.getName());
-                }
-            } else if (eventType == XmlPullParser.END_TAG) {
-                if (parser.getName().equals(ELEMENT_DATA)) {
-                    done = true;
+                if (eventType == XmlPullParser.START_TAG) {
+                    if (parser.getName().equals(ELEMENT_TEXT)) {
```
Flowdalic: That's an ideal use case for a string switch-case. See (most) other providers for examples.

Tibo-lg:
>remove the merge commit (i.e. your work should be (usually) a single commit)
I fixed that!

>fix the build errors on java 8 (see travis)
Should also be fixed.

>That's an ideal use case for a string switch-case. See (most) other providers for examples.
Agreed, I had lazily reused the existing code. It is updated now.

Flowdalic:
Cherry picked as https://github.com/igniterealtime/Smack/commit/4d57848a0f71ca6641c7fd95976aa6c0e06ee94d

I'd did some further cleaning of the HOXT code based on your commit.

Thank you !

PR #52, opened 2015-11-11
Update AbstractXMPPConnection.java

guvd:
the method connect() need to throw SASLErrorException


Flowdalic:
Why? As far as I can see that's not the case.

guvd:
Maybe i´m wrong my friend, probably i dont have you experience, but, if I write wrong password to login the system give me back it:
org.jivesoftware.smack.sasl.SASLErrorException: SASLError using DIGEST-MD5: not-authorized


As you can see is SASLErrorErrorException, its not been throwed by the class and, i can´t catch it.

Am I wrong?

Flowdalic:
> the system give me back it:
> org.jivesoftware.smack.sasl.SASLErrorException: SASLError using DIGEST-MD5: not-authorized

Yes, but this is thrown own by login() in recent version of Smack.

PR #57, opened 2015-12-01
Uses the host and port from the ConnectionConfiguration object and no…

bgrozev:
…t from the

XMPPConnection object, because the latter were observed to not be
initialized when the log is printed, resulting in the following log:
"Timeout reached for the connection to null:0."

Note that this was seen in Smack 4.07, the current version was not
tested, as we don't have an easy way to do that.


Flowdalic:
Thanks for your contribution, but it appears the proper solution would be to set the host and port in `XMPPBOSHConnection.connectInternal()`. ConnectionConfiguration intentionally provides *no* `getHost()` and `getPort()`.

Flowdalic:
And the `errorMessage` String should take an eventually existing proxy configuration into account.

PR #58, opened 2015-12-06
Changes to send service-unavailable for not handled iq.

damencho:
Changes to send service-unavailable for not handled iq as said in RFC6120. https://tools.ietf.org/html/rfc6120#page-128


Flowdalic:
Thanks for your contribution. Cherry picked with minor modifications as 035bbea

PR #59, opened 2015-12-06
Fixes sending pubsub events for empty item.

damencho:
XEP-0060 allows no elements on second level of nodes.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/Node.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/Node.java
@@ -740,6 +740,10 @@ abstract public class Node
 				{
 					List<ExtensionElement> secondLevelList = ((EmbeddedPacketExtension)embedEvent).getExtensions();
 
+					// XEP-0060 allows no elements on second level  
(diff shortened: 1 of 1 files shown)


Flowdalic:
Could you provide a pointer to the related section of xep60? I assume it's not an issue that we skip the `getElementName().equals(secondElement)` check if `secondLevelList.isEmpty()`?

damencho:
No reason for the size(), you are right.
This is for example in case of: new EventContentFilter(EventElementType.items.toString(), "item")
Where items can have no item children, as said in XEP-0060:
<xs:element name='items'>
    <xs:complexType>
      <xs:choice>
        <xs:element ref='item' minOccurs='0' maxOccurs='unbounded'/>
        <xs:element ref='retract' minOccurs='0' maxOccurs='unbounded'/>
      </xs:choice>
      <xs:attribute name='node' type='xs:string' use='required'/>
    </xs:complexType>
  </xs:element>

Flowdalic:
Cherry picked, added more strictness and committed as https://github.com/igniterealtime/Smack/commit/46f01109499ceca81a28113c0384a893cdc632ca

PR #60, opened 2015-12-06
Uses voice as default number detail, if no type is available.

damencho:


Flowdalic:
Thanks for you contribution.

The added comment doesn't state what the behavior the change really introduces: It will set the label to `"VOICE"` if `name` is not a member of the `TEL` array. I'm going to modify it a bit.

But the motivation behind the change seems legit, since RFC 2426 § 3.3.1 clearly states 

> The default type is "voice"


Please note that Smack's vcard-temp code is terrible. I always wanted to fix it completely, but I guess it's better to do little steps into the right direction for now. :)

Flowdalic:
Cherry picked with minor modifications as https://github.com/igniterealtime/Smack/commit/db72917bb42efa862c797ef736c3ca7c89b89fdc

PR #61, opened 2015-12-06
Update to handle roster packet errors.

damencho:


Flowdalic:
Thanks for your contribution. Could you
- squash the commits?
- provide a better commit message? e.g. "Add a Roster callback for error Presences"

Flowdalic:
Uh, and I just noticed that it won't work this way. You want to hook into the `ExceptionCallback` which is installed together with `RosterResultListener` to listener for error IQ responses. I guess in the end we want a callback for this, and not only for error IQ responses.

PR #62, opened 2015-12-06
Add a Roster callback for error Presences.

damencho:


damencho:
I've screwed up my git and created new PR.
Does this looks good? 
One thing that bothers me is that in smack3.2 we were receiving a roster item that wasn't good and we were ignoring it. While here if an exception occurs roaster becomes uninitialized (rosterState = RosterState.uninitialized;). Anyway it is good to have roster errors notified to a listener with the packet holding the error.

Flowdalic:
> I've screwed up my git and created new PR.

Note that you don't need to create new PRs. You could have just updated the existing PR by force pushing to its source branch.


> Does this looks good? 
> One thing that bothers me is that in smack3.2 we were receiving a roster item that wasn't good and we were ignoring it. While here if an exception occurs roaster becomes uninitialized (rosterState = RosterState.uninitialized;). Anyway it is good to have roster errors notified to a listener with the packet holding the error.

I agree that a callback is missing if there is an error when initializing the Roster. What I don't like about 7b65040 is that it changes the Exception callback adding a Stanza parameter. But not always is a Stanza involved in the failure. Instead I would link XMPPError to the stanza, so that it's retrievable from the XMPErrorException. That's what I've done in https://github.com/Flowdalic/Smack/commit/45feaecdf7bef40e9a33072fc379c19187ed8bfa, then adding the callback simply becomes https://github.com/Flowdalic/Smack/commit/feff0815472c05ad4966ef24253b0717e4059c58

Feedback to this solution is very appreciated. If I don't hear anything within the next week, this will go into the official master as it is.

damencho:
Seems fine. Exception carry all the information that is needed and is up to implementation to extract it :) seems fine. Thanks.

Flowdalic:
Fixed with https://github.com/igniterealtime/Smack/commit/feff0815472c05ad4966ef24253b0717e4059c58

PR #63, opened 2016-01-13
add 'carriage return' char for escapeForXML

Bmooij:


Flowdalic:
I don't see a reason why CR has to be escaped. The commit itself has whitespace errors, existing unit tests are failing and new ones are missing. Closing until I hear a good explanation for doing so and the mentioned issues are fixed.

Bmooij:
The reason for escaping the CR char is that is now impossible to send the CR.

If this is ok, I will fix the whitespace errors and add/update the existing unit tests

Flowdalic:
> The reason for escaping the CR char is that is now impossible to send the CR.

Why is that? XML 1.0 § 2.2 allows for CR. Please provide a detailed description by you deem this necessary.

Bmooij:
OK, my approach was not good.

XML 1.0 does allow for CR.
But it is impossible to send sequence #xD #xA due XML 1.0 § 2.11 

> sequence #xD #xA and any #xD that is not followed by #xA is translated to a single #xA character

I would like to send #xD #xA

Flowdalic:
> I would like to send #xD #xA

You can't within an XML text. You could encode the String to binary and put it base64 encoded into XML. But the requirement to send `\r\n` seems to be caused by a bad design decision.

Bmooij:
Unfortunately I can not change the design.

XML 1.0 § 2.3
> the only way to get a #xD character to match this production is to use a character reference in an entity value literal.

But when I send "Hello World &#xD;&#xA;" it is converted to "Hello World &amp;#xD;&amp;#xA;"

Bmooij:
Clear, thank you for your support

PR #68, opened 2016-02-08
XMPPTCPConnection - setEnabledSSLProtocols/ ciphers can never work

tomxnosek:
Hi,
when i was trying to force usage of TLSv1.2, I found out that setting enabledSSLProtocols and enabledSSLCiphers in XMPPTCPConnectionConfiguration can never work.

It is because of opening input/output stream on the socket before setting the parameters - implicit handshake session is created (without the custom params) and used for the handshake.

More info here:
http://stackoverflow.com/questions/13943351/how-to-specify-the-ciphersuite-to-be-used-in-ssl-session
http://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLSocket.html

Thanks!

### DIFF
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
@@ -742,12 +742,12 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
         // Secure the plain connection
         socket = context.getSocketFactory().createSocket(plain,
                 host, plain.getPort(), true);
-        // Initialize the reader and writer with the new secured version
-        initReaderAndWriter();
 
         final SSLSocket sslSocket = (SSLSocket) socket;
         TLSUtils.setEnabledProtocolsAndCiphers(sslSocket, config.getEnabledSSLProtocols(), config.getEnabledSSLCiphers());
 
+        // Initialize the reader and writer with the new secured version
+        initReaderAndWriter();
         // Proceed to do the handshake
         sslSocket.startHandshake();
 


Flowdalic:
> It is because of opening input/output stream on the socket before setting the parameters

I can't find that as documented cause for an implicit handshake. [SSLSocket](http://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLSocket.html) only lists

- calling startHandshake which explicitly begins handshakes, or
- any attempt to read or write application data on this socket causes an implicit handshake, or
- a call to getSession tries to set up a session if there is no currently valid session, and an implicit handshake is done.

But I can't rule out that something tries to read or write from/to the socket. But of course it would be great if the exact reason was known.

I assume you have verified that fixes the issue. If that's the cause, then it satisfies an issue number. I've created [SMACK-712](https://github.com/igniterealtime/Smack/pull/68). The commit needs some polishing. Could you
- Mention the issue key somewhere in the commit message
- Write a short summary about what was wrong and how you fixed it in the commit message
- Set the commit title to something like "Fix setEnabledSSL(Protocols|Ciphers)" (or similar)

tomxnosek:
Hi,
I changed the commit message as you requested.

If you check [this](http://grepcode.com/file/repo1.maven.org/maven2/org.robovm/robovm-rt/0.0.5/org/apache/harmony/xnet/provider/jsse/OpenSSLSocketImpl.java#OpenSSLSocketImpl) implementation of SSLSocket, opening of I/O stream causes starting handshake.

Flowdalic:
I've uploaded Smack 4.1.7-SNAPSHOT, which includes a fix for this, to Maven Central's snapshot repositories. Could you try it out and report back if it solves the issue for you?

Flowdalic:
Closing as SMACK-712 has been fixed.

PR #69, opened 2016-02-12
Parsing of custom message attributes

Judas:
More explanations here : https://community.igniterealtime.org/message/255265#255265

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
@@ -29,6 +29,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.jivesoftware.smack.compress.packet.Compress;
+import org.jivesoftware.smack.packet.CustomAttributesExtension;
 import org.jivesoftware.smack.packet.EmptyResultIQ;
 import org.jivesoftware.smack.packet.ErrorIQ;
 import org.jivesoftware.smack.packet.IQ;
@@ -234,6 +235,12 @@ public class PacketParserUtils {
         }
         else {
             defaultLanguage = Stanza.getDefaultLanguage();
+        }
+
+        // Extract custom attributes and bundle them into an extension
+        CustomAttributesExtension customAttributesExtension = new CustomAttributesExtension.Builder().fromParser(parser).build();
+        if (customAttributesExtension.hasProperties()) {
+            message.addExtension(customAttributesExtension);
         }
 
         // Parse sub-elements. We include extra logic to make sure the values
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/CustomAttributesExtension.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/CustomAttributesExtension.java
@@ -1,0 +1,109 @@
+/**
+ *
+ * Copyright 2016 Jules Tréhorel
+ *
+ * All rights reserved. Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smack.packet;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jivesoftware.smack.util.Objects;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+import org.xmlpull.v1.XmlPullParser;
+
+/**
+ * An {@link ExtensionElement} storing the custom attributes listed in messages stanzas root <message>
+ * element. As this use-case is a bad practice, this {@link ExtensionElement} allows XMPP clients to
+ * retrieve those (wrongly set) values in a clean way.
+ *
+ * @author Jules Tréhorel
+ */
+public class CustomAttributesExtension implements ExtensionElement {
+    public final static String NAME = "customAttributes";
+    public final static String NAMESPACE = "http://smack.jivesoftware.com";
+
+    private StandardExtensionElement extension;
+
+	private CustomAttributesExtension(Map<String, String> properties) {
+		if (properties == null || properties.isEmpty())
+			return;
+
+		StandardExtensionElement.Builder extensionBuilder = StandardExtensionElement.builder(NAME, NAMESPACE);
+		for (Map.Entry<String, String> entry : properties.entrySet()) {
+			extensionBuilder.addElement(entry.getKey(), entry.getValue());
+		}
+        extension = extensionBuilder.build();
+	}
+
+    @Override
+    public String getElementName() {
+        return NAME;
+    }
+
+    @Override
+    public String getNamespace() {
+        return NAMESPACE;
+    }
+
+    public boolean hasProperties() {
+        return extension != null && extension.getElements().size() > 0;
+    }
+
+    public boolean hasProperty(String propertyKey) {
+        return hasProperties() && extension.getElements(propertyKey) != null;
+    }
+
+    public String getPropertyValue(String propertyKey) {
+        if (!hasProperty(propertyKey)) {
+            return null;
+        }
+        return extension.getElements(propertyKey).get(0).getText(); // Only one element per key
+    }
+
+    @Override
+    public XmlStringBuilder toXML() {
+        return extension.toXML();
+    }
+
+    public static final class Builder {
+        private final List<String> attributesWhiteList = Arrays.asList(new String[] {"xml:lang", "id", "to", "from", "type"});
+
+        private Map<String, String> properties;
+
+        public Builder() {
+        }
+
+        public Builder fromParser(XmlPullParser parser) {
+            Objects.requireNonNull(parser, "Parser must be set");
+
+            properties = new HashMap<String, String>();
+            for (int i = 0; i < parser.getAttributeCount(); i++) {
+                String attributeName = parser.getAttributeName(i);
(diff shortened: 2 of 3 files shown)


MarcFRICOU:
:+1:

Flowdalic:
I've replied at https://community.igniterealtime.org/message/255393#255393
Please keep the discussion there. Thank you.

smack-core/src/main/java/org/jivesoftware/smack/util/ParserUtils.java
```diff
@@ -194,4 +198,17 @@ public static URI getUriFromNextText(XmlPullParser parser) throws XmlPullParserE
         return uri;
     }
 
+    public static Map<String, String> getCustomAttributes(XmlPullParser parser) {
+        Map<String, String> customAttributes = new LinkedHashMap<>();
+        List<String> standardAttributes = Arrays.asList(new String[] {"xml:lang", "id", "to", "from", "type"});
```
Flowdalic: Why construct this every time the method is called? And why is it a `List`? This should be a static `Set`.

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -502,6 +650,19 @@ protected void addCommonAttributes(XmlStringBuilder xml) {
...
 
     /**
+     * Add custom attributes
+     *
+     * @param xml
+     */
+    protected void addCustomAttributes(XmlStringBuilder xml) {
+        if(customAttributes != null) {
+            for (Map.Entry<String, String> entry : customAttributes.entrySet()) {
+                xml.optAttribute(entry.getKey(), entry.getValue());
```
Flowdalic: Why `optAttribute` when we know that the attribute's value is not `null`?

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -502,6 +650,19 @@ protected void addCommonAttributes(XmlStringBuilder xml) {
     }
 
     /**
+     * Add custom attributes
+     *
+     * @param xml
+     */
+    protected void addCustomAttributes(XmlStringBuilder xml) {
+        if(customAttributes != null) {
```
Flowdalic: I would go for `if (customAttributes == null) return;` here, to avoid a level of indentation.

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -458,6 +468,144 @@ public ExtensionElement removeExtension(ExtensionElement extension)  {
...
+    /**
+     * Replaces all custom attributes of the stanza.
+     * @param attributes a map (name/value) of all the custom attributes of this stanza. SHould not be {@null}
+     * @throws IllegalStateException if custom attributes management is not enabled.
+     */
+    public void setCustomAttributes(Map<String, String> attributes) {
+        requireCustomAttributesEnabled("setCustomAttributes");
+        assert attributes != null;
+
+        synchronized (customAttributes) {
```
Flowdalic: I really don't think we need to design this API thread safe. I know legacy Smack code does it here and there, but we shouldn't do it in this case.

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -458,6 +468,144 @@ public ExtensionElement removeExtension(ExtensionElement extension)  {
...
+    /**
+     * Returns a map (name/value) of all the custom attributes of this stanza.
+     * @return a map (name/value) of all the custom attributes of this stanza.
+     * @throws IllegalStateException if custom attributes management is not enabled.
+     */
+    public Map<String, String> getCustomAttributes() {
+        requireCustomAttributesEnabled("getCustomAttributes");
+
+        synchronized (customAttributes) {
+            return customAttributes;
```
Flowdalic: Unnecessary `synchronized` block

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -458,6 +468,144 @@ public ExtensionElement removeExtension(ExtensionElement extension)  {
...
+    /**
+     * Enables the possibility to add custom attributes to the stanza.
+     * <br/><b>Enabling this feature breaks the XMPP standard and may induce unexpected behaviors!
+     * <br/>If your use-case allow it, prefer using ExtensionElement instead, and if not, be sure to triple-check the result.</b>
+     */
+    public static void enableCustomAttributes() {
+        customAttributesEnabled = true;
+    }
+
+    private static void requireCustomAttributesEnabled(String methodName) {
```
Flowdalic: If you really want to the callers method name here, then get it via runtime mechanism instead of hardcoding it at the call site. All in all, I don't think it's really required, so just remove the methodName argument.

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -92,6 +97,11 @@ protected Stanza(Stanza p) {
         for (ExtensionElement pe : p.getExtensions()) {
             addExtension(pe);
         }
+
+        if (isCustomAttributesEnabled()) {
```
Flowdalic: No need to check here.

smack-core/src/main/java/org/jivesoftware/smack/packet/IQ.java
```diff
@@ -117,6 +117,7 @@ public final XmlStringBuilder toXML() {
         XmlStringBuilder buf = new XmlStringBuilder();
         buf.halfOpenElement(IQ_ELEMENT);
         addCommonAttributes(buf);
+        addCustomAttributes(buf);
```
Flowdalic: I would merge `addCommonattributes` and `addCustomAttributes`, so just add the code in `addCustomAttributes` to the already existing method and rename it to `addAttributes`.

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -458,6 +468,144 @@ public ExtensionElement removeExtension(ExtensionElement extension)  {
         return removeExtension(extension.getElementName(), extension.getNamespace());
     }
 
+    /**
+     * Returns <tt>true</tt> if custom attributes management is enabled.
+     * @return <tt>true</tt> if custom attributes management is enabled.
+     */
+    public static boolean isCustomAttributesEnabled() {
```
Flowdalic: Not sure, but maybe `areCustomAttributesEnabled()`?

smack-core/src/main/java/org/jivesoftware/smack/packet/Stanza.java
```diff
@@ -458,6 +468,144 @@ public ExtensionElement removeExtension(ExtensionElement extension)  {
...
+    }
+
+    /**
+     * Adds a custom attribute to the stanza. If a custom attribute with the same name is already present, its value is updated.
+     * @param attributeName the custom attribute name.
+     * @param attributeValue the custom attribute value.
+     * @throws IllegalStateException if custom attributes management is not enabled.
+     */
+    public void setCustomAttribute(String attributeName, String attributeValue) {
+        requireCustomAttributesEnabled("setCustomAttribute");
```
Flowdalic: You need to check in the set* methods that the custom attribute name doesn't clash with a common attribute name to prevent users from producing invalid XML. Throw an IllegalArgumentException if so.

smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
```diff
@@ -611,6 +619,7 @@ public static IQ parseIQ(XmlPullParser parser) throws Exception {
         final Jid to = ParserUtils.getJidAttribute(parser, "to");
         final Jid from = ParserUtils.getJidAttribute(parser, "from");
         final IQ.Type type = IQ.Type.fromString(parser.getAttributeValue("", "type"));
+        final Map<String, String> customAttributes = ParserUtils.getCustomAttributes(parser);
```
Flowdalic: Why is this not conditional?

smack-core/src/main/java/org/jivesoftware/smack/util/ParserUtils.java
```diff
@@ -194,4 +198,17 @@ public static URI getUriFromNextText(XmlPullParser parser) throws XmlPullParserE
...
+            for (int i = 0; i < parser.getAttributeCount(); i++) {
+                String attributeName = parser.getAttributeName(i);
+                boolean isLangAttribute = "lang".equals(attributeName) && "xml".equals(parser.getAttributePrefix(i));
```
Flowdalic: Isn't this redundant with the `xml:lang` check because of `standardAttributes`?

Judas: I reused the logic I found in `PacketParserUtils.getLanguageAttribute()`. Do you think it's overkill in this case ?

Flowdalic:
I've just looked at this again, and I think that

```java
    public static Map<String, String> getCustomAttributes(XmlPullParser parser) {
        Map<String, String> customAttributes = new LinkedHashMap<>();
        for (int i = 0; i < parser.getAttributeCount(); i++) {
            String attributeName = parser.getAttributeName(i);
            boolean isLangAttribute = "lang".equals(attributeName) && "xml".equals(parser.getAttributePrefix(i));
            if (!Stanza.REGISTERED_ATTRIBUTES.contains(attributeName) && !isLangAttribute) {
                String attributeValue = parser.getAttributeValue(i);
                customAttributes.put(attributeName, attributeValue);
            }
        }
        return customAttributes;
    }
```

should take the attribute prefix into account and eventually prefix it. This is untested, but I believe that if we have

```xml
<element xmlns='foo' xmlns:myns='https://my.org' attr1='x' myns:attr2='y'>
```

then the current code would produce a map of

```text
attr1 → x
myns → https://my.org
attr2 → y
```

while it should produce a map like

```text
attr1 → x
xmlns:myns → https://my.org
myns:attr2 → y
```

If that's the case, then this should be fixed and a unit test for this should be created.

Flowdalic:
Are you still up to pursue this further? Let me know if not, if so I probably squash it myself and add the missing features.

Judas:
Hi Flow,
I didn't took the time to push it but I've added the unit test, and it seems the "good" result is returned by using your example. I'll try to update/push this week.

FGRibreau:
I definitely need this :+1:!

PR #72, opened 2016-04-07
Fix documentation formatting in PayloadItem.java and ItemsExtension.java

linux4life798:
These commits fix some of the very broken documentation formatting in the pubsub extension directory.

I made the formatting of ItemsExtension.java similar to that of Item.java.


Flowdalic:
Thanks for you contribution. Could you squash the three commits into one?

linux4life798:
Yep, sure. Didn't know how verbose to be.

Flowdalic:
You don't need to close a PR in order to do so. You could have simply
forcefully pushed to the source branch.

linux4life798:
My apologies. Thanks for the tip!

PR #73, opened 2016-04-07
Implementation of XEP-0308: Last Message Correction

ramabit:
## **Fixes SMACK-714**
Documentation: https://xmpp.org/extensions/xep-0308.html

**How to use it**

Send: 
`message.addExtension(new MessageCorrectExtension(idInitialMessage));`

Receive:
`ProviderManager.addExtensionProvider(MessageCorrectExtension.ELEMENT_NAME, MessageCorrectExtension.NAMESPACE, new MessageCorrectProvider());`

`MessageCorrectExtension messageCorrectExtension = MessageCorrectExtension.from(message)`

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
@@ -1,0 +1,105 @@
+/**
+ *
+ * Copyright 2016 Fernando Ramirez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.message_correct;
+
+import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+
+/**
+ * An Extension that implements XEP-0308: Last Message Correction
+ * 
+ * This extension is expected to be added to message stanzas. Please refer to
+ * the XEP for more implementation guidelines.
+ * 
+ * @author Fernando Ramirez, f.e.ramirez94@gmail.com
+ * @see <a href="http://xmpp.org/extensions/xep-0308.html">XEP-0308:&nbsp;Last&
+ *      nbsp;Message&nbsp;Correction</a>
+ */
+public class MessageCorrectExtension implements ExtensionElement {
+
+    /**
+     * The XML element name of a 'message correct' extension.
+     */
+    public static final String ELEMENT = "replace";
+
+    /**
+     * The namespace that qualifies the XML element of a 'message correct'
+     * extension.
+     */
+    public static final String NAMESPACE = "urn:xmpp:message-correct:0";
+
+    /**
+     * The id tag of a 'message correct' extension.
+     */
+    public static final String ID_TAG = "id";
+
+    /**
+     * The id of the message to correct.
+     */
+    private String idInitialMessage;
+
+    public MessageCorrectExtension(String idInitialMessage) {
+        this.setIdInitialMessage(idInitialMessage);
+    }
+
+    public String getIdInitialMessage() {
+        return idInitialMessage;
+    }
+
+    public void setIdInitialMessage(String idInitialMessage) {
+        this.idInitialMessage = idInitialMessage;
+    }
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see org.jivesoftware.smack.packet.PacketExtension#getElementName()
+     */
+    @Override
+    public String getElementName() {
+        return ELEMENT;
+    }
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see org.jivesoftware.smack.packet.PacketExtension#toXML()
+     */
+    @Override
+    public XmlStringBuilder toXML() {
+        XmlStringBuilder xml = new XmlStringBuilder(this);
+        xml.attribute(ID_TAG, getIdInitialMessage());
+        xml.closeEmptyElement();
+        return xml;
+    }
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see org.jivesoftware.smack.packet.PacketExtension#getNamespace()
+     */
+    @Override
+    public String getNamespace() {
+        return NAMESPACE;
+    }
+
+    public static MessageCorrectExtension from(Message message) {
+        return message.getExtension(ELEMENT, NAMESPACE);
+    }
+
+}
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/provider/MessageCorrectProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/provider/MessageCorrectProvider.java
@@ -1,0 +1,42 @@
+/**
+ *
+ * Copyright 2016 Fernando Ramirez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.message_correct.provider;
+
+import java.io.IOException;
+
+import org.jivesoftware.smack.SmackException;
+import org.jivesoftware.smack.provider.ExtensionElementProvider;
+import org.jivesoftware.smackx.message_correct.MessageCorrectExtension;
+import org.xmlpull.v1.XmlPullParser;
+import org.xmlpull.v1.XmlPullParserException;
+
+/**
+ * A ExtensionElementProvider for the MessageCorrectExtension. As
+ * MessageCorrection elements have only the ID of the message to replace.
+ * 
+ * @author Fernando Ramirez, f.e.ramirez94@gmail.com
+ */
+public class MessageCorrectProvider extends ExtensionElementProvider<MessageCorrectExtension> {
+
+    @Override
+    public MessageCorrectExtension parse(XmlPullParser parser, int initialDepth)
+            throws XmlPullParserException, IOException, SmackException {
+        String idMessageToReplace = parser.getAttributeValue("", MessageCorrectExtension.ID_TAG);
+        return new MessageCorrectExtension(idMessageToReplace);
+    }
+
+}
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/package-info.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/package-info.java
@@ -1,0 +1,21 @@
+/**
+ *
+ * Copyright 2016 Fernando Ramirez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Classes and Interfaces that implement Last Message Correction as defined in XEP-0308.
+ */
+package org.jivesoftware.smackx.message_correct;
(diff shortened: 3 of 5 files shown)


ramabit:
@Flowdalic TravisCI is killing the process of the 2nd check (java 1.7). I guess it is because of low memory.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
+/**
+ *
+ * Copyright the original author or authors
```
Flowdalic: That line should include a copyright year, i.e. 2016.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+ * the XEP for more implementation guidelines.
+ * 
+ * @author Fernando Ramirez, f.e.ramirez94@gmail.com
+ */
+public class MessageCorrectExtension implements ExtensionElement {
+
+    /**
+     * The XML element name of a 'message correct' extension.
+     */
+    public static final String ELEMENT_NAME = "replace";
```
Flowdalic: Should be just 'ELEMENT'

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see org.jivesoftware.smack.packet.PacketExtension#toXML()
+     */
+    @Override
+    public CharSequence toXML() {
+        final StringBuilder stringBuilder = new StringBuilder();
```
Flowdalic: Please use XmlStringBuilder here, and change the return type to this.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+        return NAMESPACE;
+    }
+
+    /**
+     * A ExtensionElementProvider for the MessageCorrectExtension. 
+     * As MessageCorrection elements have only the JID of the message to replace.
+     * 
+     * @author Fernando Ramirez, f.e.ramirez94@gmail.com
+     */
+    public static class Provider extends ExtensionElementProvider<MessageCorrectExtension> {
```
Flowdalic: Please factor the provider out in a extra class under `.provider`

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.provider.ExtensionElementProvider;
+import org.xmlpull.v1.XmlPullParser;
+import org.xmlpull.v1.XmlPullParserException;
+
+/**
+ * An Extension that implements XEP-0308: Last Message Correction
+ * 
+ * This extension is expected to be added to message stanzas. Please refer to
+ * the XEP for more implementation guidelines.
```
Flowdalic: You should at least mention *which* XEP you are talking about. An `<a/>` tag in `@see` couldn't hurt either.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * The id tag of a 'message correct' extension.
+     */
+    private static final String JID_TAG = "id";
+
+    /**
+     * The jid of the message to correct.
+     */
+    private String jidInitialMessage;
```
Flowdalic: That's no JID/XMPP address, this is just the message ID. Please rename this to `id`.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Classes and Interfaces that implement Last Message Correction as defined in XEP-0308.
+ */
+package org.jivesoftware.smackx.message.correct;
```
Flowdalic: Please change the package name to `message_correct` (if it is possible, but I think it should).

smack-extensions/src/test/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtensionTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final CharSequence messageCorrectionXml = "<replace id='bad1' xmlns='urn:xmpp:message-correct:0'/>";
+
+    private final CharSequence expectedXml = "<message to='juliet@capulet.net/balcony' id='good1'>"
+            + "<body>But soft, what light through yonder window breaks?</body>"
+            + "<replace id='bad1' xmlns='urn:xmpp:message-correct:0'/>" + "</message>";
+
+    @Before
+    public void setStanzas() throws Exception {
+        initialMessage = (Message) PacketParserUtils.parseStanza(initialMessageXml);
```
Flowdalic: That's a strange approach. Smack usually does the parsing in the `@Test` method. What's the advantage of having it done in a `@Before` method?

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * The namespace that qualifies the XML element of a 'message correct'
+     * extension.
+     */
+    public static final String NAMESPACE = "urn:xmpp:message-correct:0";
+
+    /**
+     * The id tag of a 'message correct' extension.
+     */
+    private static final String JID_TAG = "id";
```
Flowdalic: Same here, this is tag is not related to JIDs at all.

Flowdalic:
Please squash merge all commits into a single commit and make sure that the issue's key 'SMACK-714' is mentioned somewhere in the commit message's body. Please don't put it into the title. Something like "Fixes SMACK-714" is sufficient.

smack-extensions/src/main/java/org/jivesoftware/smackx/message/correct/MessageCorrectExtension.java
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * An Extension that implements XEP-0308: Last Message Correction
+ * 
+ * This extension is expected to be added to message stanzas. Please refer to
+ * the XEP for more implementation guidelines.
+ * 
+ * @author Fernando Ramirez, f.e.ramirez94@gmail.com
+ */
+public class MessageCorrectExtension implements ExtensionElement {
```
Flowdalic: It's Smackomatic (a Smack Idiom) that ExtensionElements have a static `from()` method. Since XEP-308 is only defined for messages, in this case we should have

`static MessageCorrectionExtension from(Message)`

Please add this method.

ramabit:
@Flowdalic Thanks for the feedback ! 
The corrections are done and the pull request description is updated.
TravisCI is failing because a test that I didn't touch. Also, I can't reproduce that failure locally.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
```diff
@@ -83,11 +81,13 @@ public String getElementName() {
      * @see org.jivesoftware.smack.packet.PacketExtension#toXML()
      */
     @Override
-    public CharSequence toXML() {
-        final StringBuilder stringBuilder = new StringBuilder();
-        stringBuilder.append('<').append(getElementName()).append(' ').append(JID_TAG).append('=').append('\'')
-                .append(getJidInitialMessage()).append('\'').append(" xmlns='").append(getNamespace()).append("'/>");
-        return stringBuilder.toString();
+    public XmlStringBuilder toXML() {
+        XmlStringBuilder xml = new XmlStringBuilder(this, getNamespace());
```
Flowdalic: Just `new XmlStringBuilder(this)` will do the right thing, i.e. adding the opening tag and the xmlns attribute.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
```diff
@@ -83,11 +81,13 @@ public String getElementName() {
...
+        xml.attribute(ID_TAG, getIdInitialMessage());
+        xml.attribute("xmlns", getNamespace());
```
Flowdalic: This would create a duplicate `xmlns` attribute and thus invalid XML.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
```diff
@@ -83,11 +81,13 @@ public String getElementName() {
...
+        xml.append('/');
+        xml.rightAngleBracket();
```
Flowdalic: Instead of these two, please use `closeEmptyElement()` instead.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
```diff
@@ -100,21 +100,8 @@ public String getNamespace() {
...
-
-        @Override
-        public MessageCorrectExtension parse(XmlPullParser parser, int initialDepth)
-                throws XmlPullParserException, IOException, SmackException {
-            String jidMessageToReplace;
-            jidMessageToReplace = parser.getAttributeValue("", JID_TAG);
-            return new MessageCorrectExtension(jidMessageToReplace);
-        }
+    public static MessageCorrectExtension from(Message message) {
+        return new MessageCorrectExtension(message.getStanzaId());
```
Flowdalic: It appears I was unclear with what `from(Message)` should do: Instead of creating a new MessageCorrect, it should extract the extension from the message if there is one. The body of the function should be
`return message.getExtension(ELEMENT, NAMESPACE);`

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/provider/MessageCorrectProvider.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * The id tag of a 'message correct' extension.
+     */
+    private static final String ID_TAG = "id";
+
+    @Override
+    public MessageCorrectExtension parse(XmlPullParser parser, int initialDepth)
+            throws XmlPullParserException, IOException, SmackException {
+        String jidMessageToReplace;
+        jidMessageToReplace = parser.getAttributeValue("", ID_TAG);
```
Flowdalic: Instead of defining ID_TAG again, you could simply reuse ID_TAG from MessageCorrectExtension by making it public.

Flowdalic:
> TravisCI is failing because a test that I didn't touch. Also, I can't reproduce that failure locally.

Some older Smack unit tests are not properly synchronized causing them to behave non deterministc especially on slow systems like the Travis CI infrastructure. I've restarted Travis (not sure if you are also able to hit the restart button, e.g. at https://travis-ci.org/igniterealtime/Smack/jobs/122258804 in the upper right corner).

Looks good so far :+1: I've added a few last remarks, please incorporate the changes, then squash merge all you commits into a single commit and then force push to the source branch of this PR.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
ramabit: But I want to show the id attribute before the xmlns attribute. I'll do this
`        XmlStringBuilder xml = new XmlStringBuilder();
        xml.halfOpenElement(ELEMENT);
        xml.attribute(ID_TAG, getIdInitialMessage());
        xml.xmlnsAttribute(NAMESPACE);
        xml.closeEmptyElement();`

Flowdalic:
By the way, this PR only implements the low-level API parts for XEP-308. While it's ok to start with them first, do you plan to submit an subsequent PR with the high-level API, i.e. the MessageCorrectionManager)?

ramabit:
> By the way, this PR only implements the low-level API parts for XEP-308. While it's ok to start with them first, do you plan to submit an subsequent PR with the high-level API, i.e. the MessageCorrectionManager)? 
> 

It's not planned up to now, but I think we can discuss it.

smack-extensions/src/main/java/org/jivesoftware/smackx/message_correct/MessageCorrectExtension.java
Flowdalic: > But I want to show the id attribute before the xmlns attribute.

Why? The order is completely irrelevant, but it's common practice to have the unprefixed `xmlns` attribute first (if one is required). This is why the XmlStringBuilder API is designed this way.

ramabit: Ok :)

ramabit:
@Flowdalic Changes done

Flowdalic:
Perfect, now you need to squash the commits into one.

ramabit:
@Flowdalic Github has a new option to do that when merge.

ramabit:
@Flowdalic I can close this PR an open another one squashing the commits

Flowdalic:
> I can close this PR an open another one squashing the commits

Simply update **this** PR by updating its source branch with a forced push.

ramabit:
Simply update this PR by updating its source branch with a forced push.
> 

@Flowdalic done

Flowdalic:
Thanks, but the message's body now full of meaningless half sentences, it also does not contain the issue key, which is required so that JIRA connects the issue with this commit (https://github.com/igniterealtime/Smack/pull/73#issuecomment-207594222). Please fix that.

ramabit:
@Flowdalic like that?

Flowdalic:
Picked as https://github.com/igniterealtime/Smack/commit/ca58c65a26034bd2b7c4b7e8c2bc70ab593d2ec7

Thanks again :)

PR #75, opened 2016-05-01
Working with tor and Hidden Services.

Fluturenet:
Modified version of smack-tcp able to connect to a TOR proxy and an hidden Onion server.

### DIFF
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
@@ -147,6 +147,7 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
 
     private static final int QUEUE_SIZE = 500;
     private static final Logger LOGGER = Logger.getLogger(XMPPTCPConnection.class.getName());
+    private XMPPTCPOnionCompanion xmpptcpOnionCompanion = null;
 
     /**
      * The socket which is used for this connection.
@@ -326,6 +327,26 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
      */
     public XMPPTCPConnection(CharSequence jid, String password) throws XmppStringprepException {
         this(XmppStringUtils.parseLocalpart(jid.toString()), password, XmppStringUtils.parseDomain(jid.toString()));
+    }
+
+    /**
+     * Adapted Constructor for the Onion.
+     *
+     */
+    public XMPPTCPConnection(XMPPTCPConnectionConfiguration config, XMPPTCPOnionCompanion xmpptcpOnionCompanion){
+        super(config);
+        this.config = config;
+        addConnectionListener(new AbstractConnectionListener() {
+            @Override
+            public void connectionClosedOnError(Exception e) {
+                if (e instanceof XMPPException.StreamErrorException) {
+                    dropSmState();
+                }
+            }
+        });
+
+        this.xmpptcpOnionCompanion = xmpptcpOnionCompanion;
+
     }
 
     /**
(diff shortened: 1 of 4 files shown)


smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
```diff
@@ -329,6 +330,26 @@ public XMPPTCPConnection(CharSequence jid, String password) throws XmppStringpre
...
+     *
+     */
+    public XMPPTCPConnection(XMPPTCPConnectionConfiguration config, XMPPTCPOnionCompanion xmpptcpOnionCompanion){
+        super(config);
+        this.config = config;
+        addConnectionListener(new AbstractConnectionListener() {
+            @Override
+            public void connectionClosedOnError(Exception e) {
+                if (e instanceof XMPPException.StreamErrorException) {
+                    dropSmState();
```
Flowdalic: Is this related to this TOR support?

Flowdalic:
I'm wondering if this change is necessary. It appears to me that you can use Smack with TOR without leaking DNS lookups by setting the host and port manually and using a socks proxy. Or am I mistaken?

Fluturenet:
the most important pitfall that makes the connection impossible is the extensive use of InetAddress class that autmatically resolves host names. 
The Guava verion doesn't: `http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/net/InetAddresses.html`
If you try to resolve a .onion host you will get an empty string.

There is also something that I think is a procedural error:
Socks5ProxySocketConnection  method accepts only an InetAddress then resolves it with the local DNS then sends the resolved host name to the sock server: Wouldn't it be better to delegate the name resolution to the socks server instead? 

I added getHost and getPort to the configurator because these values where protected.

> Edit: Ahh there is a getAllByName(). Wouldn't be the better solution to be simply not performing this lookup if 'host' is an IPv(4|6)? This would likely bring the required changes for TOR support down to a few LOCs (compared to this commit).

TOR refuses numerical IP, a FQDN is mandatory and tor hidden services can be reached only with their .onion name.

PS: this is the product of an hobbist coder not a professional, I think that, if yuo modify the proxyInfo, the configurator and the connection class everything can be made in a more clean way.

Flowdalic:
I've uploaded an experimental change to https://github.com/Flowdalic/Smack/tree/tor

Could you test and report back if it allows connections to `.onion` domains by setting a host, port and socks5 proxy?

Fluturenet:
What can I say?
[`It's Alive!!!](https://www.youtube.com/watch?v=bPmVhyHBRAM)

Works perfectly

doesn't it breaks:
xmppConnection.getConfiguration().getProxyInfo().getProxySocketConnection().connect(); ?

It doesn't matter for me 'cause I'm starting a new app.

Michele

Flowdalic:
> What can I say?
> `It's Alive!!!
> 
> Works perfectly

Hehe, good. :)

> doesn't it breaks:
> xmppConnection.getConfiguration().getProxyInfo().getProxySocketConnection().connect(); ?

I'm sorry I don't understand? Which LOC are you referring to? What problem are you envisioning?

I want to put a thank you note into the commit message.How should I write your name? Do you want to reveal your real name, do you prefer to stay anonymous or should I go with 'Michele'?

Fluturenet:
> I want to put a thank you note into the commit message.How should I write your name? Do you want to reveal your real name, do you prefer to stay anonymous or should I go with 'Michele'?

It makes me so proud. You can insert my name (Michele Pinto).

>  doesn't it breaks:
    xmppConnection.getConfiguration().getProxyInfo().getProxySocketConnection().connect(); ?
> I'm sorry I don't understand? Which LOC are you referring to? What problem are you envisioning?

If someone has already written a code (extension) that relies on the getProxySocketConnection().connect() expects to find a InetAddress and not a String. It can be solved with a overloading of the connect() method with an annotation @Deprecated

Flowdalic:
Fixed with https://igniterealtime.org/issues/browse/SMACK-720

PR #76, opened 2016-05-02
Implementation of XEP-0313: Message Archive Management (MAM)

ramabit:
Documentation: http://xmpp.org/extensions/xep-0313.html

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
@@ -1,0 +1,561 @@
+/**
+ *
+ * Copyright © 2016 Florian Schmaus and Fernando Ramirez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.mam;
+
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.ConnectionCreationListener;
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.PacketCollector;
+import org.jivesoftware.smack.SmackException.NoResponseException;
+import org.jivesoftware.smack.SmackException.NotConnectedException;
+import org.jivesoftware.smack.SmackException.NotLoggedInException;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPConnectionRegistry;
+import org.jivesoftware.smack.XMPPException.XMPPErrorException;
+import org.jivesoftware.smack.filter.IQReplyFilter;
+import org.jivesoftware.smack.packet.IQ;
+import org.jivesoftware.smack.packet.IQ.Type;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.forward.packet.Forwarded;
+import org.jivesoftware.smackx.mam.element.MamElements;
+import org.jivesoftware.smackx.mam.element.MamFinIQ;
+import org.jivesoftware.smackx.mam.element.MamPrefsIQ;
+import org.jivesoftware.smackx.mam.element.MamQueryIQ;
+import org.jivesoftware.smackx.mam.filter.MamResultFilter;
+import org.jivesoftware.smackx.rsm.packet.RSMSet;
+import org.jivesoftware.smackx.xdata.FormField;
+import org.jivesoftware.smackx.xdata.packet.DataForm;
+import org.jxmpp.jid.Jid;
+import org.jxmpp.util.XmppDateTime;
+
+/**
+ * Message Archive Management Manager class.
+ * 
+ * @see <a href="http://xmpp.org/extensions/xep-0313.html">XEP-0313: Message
+ *      Archive Management</a>
+ * @author Fernando Ramirez and Florian Schmaus
+ * 
+ */
+public final class MamManager extends Manager {
+
+    static {
+        XMPPConnectionRegistry.addConnectionCreationListener(new ConnectionCreationListener() {
+            @Override
+            public void connectionCreated(XMPPConnection connection) {
+                getInstanceFor(connection);
+            }
+        });
+    }
+
+    private static final Map<XMPPConnection, MamManager> INSTANCES = new WeakHashMap<>();
+
+    /**
+     * Get the singleton instance of MamManager.
+     * 
+     * @param connection
+     * @return the instance of MamManager
+     */
+    public static synchronized MamManager getInstanceFor(XMPPConnection connection) {
+        MamManager mamManager = INSTANCES.get(connection);
+
+        if (mamManager == null) {
+            mamManager = new MamManager(connection);
+            INSTANCES.put(connection, mamManager);
+        }
+
+        return mamManager;
+    }
+
+    private MamManager(XMPPConnection connection) {
+        super(connection);
+    }
+
+    /**
+     * Query archive with a maximum amount of results.
+     * 
+     * @param max
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchive(Integer max) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(max, null, null, null, null);
+    }
+
+    /**
+     * Query archive with a JID (only messages from/to the JID).
+     * 
+     * @param withJid
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchive(Jid withJid) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(null, null, null, withJid, null);
+    }
+
+    /**
+     * Query archive filtering by start and/or end date. If start == null, the
+     * value of 'start' will be equal to the date/time of the earliest message
+     * stored in the archive. If end == null, the value of 'end' will be equal
+     * to the date/time of the most recent message stored in the archive.
+     * 
+     * @param start
+     * @param end
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchive(Date start, Date end) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(null, start, end, null, null);
+    }
+
+    /**
+     * Query Archive adding filters with additional fields.
+     * 
+     * @param additionalFields
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchive(List<FormField> additionalFields) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(null, null, null, null, additionalFields);
+    }
+
+    /**
+     * Query archive filtering by start date. The value of 'end' will be equal
+     * to the date/time of the most recent message stored in the archive.
+     * 
+     * @param start
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchiveWithStartDate(Date start) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(null, start, null, null, null);
+    }
+
+    /**
+     * Query archive filtering by end date. The value of 'start' will be equal
+     * to the date/time of the earliest message stored in the archive.
+     * 
+     * @param end
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchiveWithEndDate(Date end) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        return queryArchive(null, null, end, null, null);
+    }
+
+    /**
+     * Query archive applying filters: max count, start date, end date, from/to
+     * JID and with additional fields.
+     * 
+     * @param max
+     * @param start
+     * @param end
+     * @param withJid
+     * @param additionalFields
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult queryArchive(Integer max, Date start, Date end, Jid withJid, List<FormField> additionalFields)
+            throws NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException,
+            NotLoggedInException {
+        DataForm dataForm = null;
+        String queryId = UUID.randomUUID().toString();
+
+        if (start != null || end != null || withJid != null || additionalFields != null) {
+            dataForm = getNewMamForm();
+            addStart(start, dataForm);
+            addEnd(end, dataForm);
+            addWithJid(withJid, dataForm);
+            addAdditionalFields(additionalFields, dataForm);
+        }
+
+        MamQueryIQ mamQueryIQ = prepareMamQueryIQSet(dataForm, queryId);
+        addResultsLimit(max, mamQueryIQ);
+        return queryArchive(mamQueryIQ);
+    }
+
+    private static void addAdditionalFields(List<FormField> additionalFields, DataForm dataForm) {
+        if (additionalFields == null) {
+            return;
+        }
+        for (FormField formField : additionalFields) {
+            dataForm.addField(formField);
+        }
+    }
+
+    private static void addResultsLimit(Integer max, MamQueryIQ mamQueryIQ) {
+        if (max == null) {
+            return;
+        }
+        RSMSet rsmSet = new RSMSet(max);
+        mamQueryIQ.addExtension(rsmSet);
+
+    }
+
+    private static void addWithJid(Jid withJid, DataForm dataForm) {
+        if (withJid == null) {
+            return;
+        }
+        FormField formField = new FormField("with");
+        formField.addValue(withJid.toString());
+        dataForm.addField(formField);
+    }
+
+    private static void addEnd(Date end, DataForm dataForm) {
+        if (end == null) {
+            return;
+        }
+        FormField formField = new FormField("end");
+        formField.addValue(XmppDateTime.formatXEP0082Date(end));
+        dataForm.addField(formField);
+    }
+
+    private static void addStart(Date start, DataForm dataForm) {
+        if (start == null) {
+            return;
+        }
+        FormField formField = new FormField("start");
+        formField.addValue(XmppDateTime.formatXEP0082Date(start));
+        dataForm.addField(formField);
+    }
+
+    private void preparePageQuery(MamQueryIQ mamQueryIQ, RSMSet rsmSet) {
+        mamQueryIQ.setType(IQ.Type.set);
+        mamQueryIQ.addExtension(rsmSet);
+    }
+
+    /**
+     * Returns a page of the archive.
+     * 
+     * @param dataForm
+     * @param rsmSet
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult page(DataForm dataForm, RSMSet rsmSet) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        MamQueryIQ mamQueryIQ = new MamQueryIQ(UUID.randomUUID().toString(), dataForm);
+        preparePageQuery(mamQueryIQ, rsmSet);
+        return queryArchive(mamQueryIQ);
+    }
+
+    /**
+     * Returns the next page of the archive.
+     * 
+     * @param mamQueryResult
+     *            is the previous query result
+     * @param count
+     *            is the amount of messages that a page contains
+     * @return the MAM query result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamQueryResult pageNext(MamQueryResult mamQueryResult, int count) throws NoResponseException,
+            XMPPErrorException, NotConnectedException, InterruptedException, NotLoggedInException {
+        RSMSet previousResultRsmSet = mamQueryResult.mamFin.getRSMSet();
+        RSMSet requestRsmSet = new RSMSet(count, previousResultRsmSet.getLast(), RSMSet.PageDirection.after);
+        return page(mamQueryResult.form, requestRsmSet);
+    }
+
+    /**
+     * Obtain page before the first message saved (specific chat).
+     *
+     * @param chatJid
+     * @param firstMessageId
+     * @param max
+     * @return the MAM query result
+     * @throws XMPPErrorException
+     * @throws NotLoggedInException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NoResponseException
+     */
+    public MamQueryResult pageBefore(Jid chatJid, String firstMessageId, int max) throws XMPPErrorException,
+            NotLoggedInException, NotConnectedException, InterruptedException, NoResponseException {
+        RSMSet rsmSet = new RSMSet(null, firstMessageId, -1, -1, null, max, null, -1);
+        DataForm dataForm = getNewMamForm();
+        addWithJid(chatJid, dataForm);
+        return page(dataForm, rsmSet);
+    }
+
+    /**
+     * Obtain page after the last message saved (specific chat).
+     *
+     * @param chatJid
+     * @param lastMessageId
+     * @param max
+     * @return the MAM query result
+     * @throws XMPPErrorException
+     * @throws NotLoggedInException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NoResponseException
+     */
+    public MamQueryResult pageAfter(Jid chatJid, String lastMessageId, int max) throws XMPPErrorException,
+            NotLoggedInException, NotConnectedException, InterruptedException, NoResponseException {
+        RSMSet rsmSet = new RSMSet(lastMessageId, null, -1, -1, null, max, null, -1);
+        DataForm dataForm = getNewMamForm();
+        addWithJid(chatJid, dataForm);
+        return page(dataForm, rsmSet);
+    }
+
+    /**
+     * Get the form fields supported by the server.
+     * 
+     * @return the list of form fields.
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public List<FormField> retrieveFormFields() throws NoResponseException, XMPPErrorException, NotConnectedException,
+            InterruptedException, NotLoggedInException {
+        String queryId = UUID.randomUUID().toString();
+        MamQueryIQ mamQueryIQ = prepareMamQueryIQGet(queryId);
+        return queryFormFields(mamQueryIQ);
+    }
+
+    private MamQueryIQ prepareMamQueryIQSet(DataForm dataForm, String queryId) {
+        MamQueryIQ mamQueryIQ = new MamQueryIQ(queryId, dataForm);
+        mamQueryIQ.setType(IQ.Type.set);
+        return mamQueryIQ;
+    }
+
+    private MamQueryIQ prepareMamQueryIQGet(String queryId) {
+        MamQueryIQ mamQueryIQ = new MamQueryIQ(queryId, null);
+        mamQueryIQ.setType(IQ.Type.get);
+        return mamQueryIQ;
+    }
+
+    private MamQueryResult queryArchive(MamQueryIQ mamQueryIq) throws NoResponseException, XMPPErrorException,
+            NotConnectedException, InterruptedException, NotLoggedInException {
+        final XMPPConnection connection = getAuthenticatedConnectionOrThrow();
+        MamFinIQ mamFinIQ = null;
+
+        PacketCollector mamFinIQCollector = connection.createPacketCollector(new IQReplyFilter(mamQueryIq, connection));
+
+        PacketCollector.Configuration resultCollectorConfiguration = PacketCollector.newConfiguration()
+                .setStanzaFilter(new MamResultFilter(mamQueryIq)).setCollectorToReset(mamFinIQCollector);
+        PacketCollector resultCollector = connection.createPacketCollector(resultCollectorConfiguration);
+
+        try {
+            connection.sendStanza(mamQueryIq);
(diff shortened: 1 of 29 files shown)


Flowdalic:
After a quick look: Appears to be missing the provider registration and the entry in https://github.com/igniterealtime/Smack/blob/master/documentation/extensions/index.md

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinProvider.java
```diff
@@ -0,0 +1,31 @@
...
+package org.jivesoftware.smackx.mam.provider;
+
+/**
+ * MAM Fin Provider class.
+ * 
+ * @see <a href="http://xmpp.org/extensions/xep-0313.html">XEP-0313: Message
+ *      Archive Management</a>
+ *
+ */
+public class MamFinProvider {
```
Flowdalic: This and the following provider look unfinished

ramabit:
@Flowdalic ah sorry. When I implemented xep-308 I didn't change that file (index.md). Should I do it?

Flowdalic:
> When I implemented xep-308 I didn't change that file (index.md). Should I do it?

Sure :)

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinProvider.java
ramabit: Didn't know if that provider was necessary

ramabit: @Flowdalic I am adding providers implementation. Also I must add new tests so this will take several hours.

Flowdalic: No problem, you are not in a race against the time. This will likely take some time to merge. In fact, I would welcome if could also write some integration tests for MAM.

ramabit: @Flowdalic I am having some trouble running the integration tests ("gradle integrationTest" command). Seems to display always errors, even when I didn't change anything

ramabit:
@Flowdalic I added providers implementation + providers tests + MAM description in index.md

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -0,0 +1,522 @@
...
+                    .nextResultOrThrow(connection.getPacketReplyTimeout() + extraTimeout);
+            mamFinExtension = MamFinExtension.from(mamFinMessage);
+        } finally {
+            resultCollector.cancel();
+            finMessageCollector.cancel();
+        }
+
+        List<Forwarded> messages = new ArrayList<>(resultCollector.getCollectedCount());
+
+        for (Message resultMessage = resultCollector.pollResult(); resultMessage != null;) {
```
steckbrief: This line leads to an endless loop, because only the first message is polled from the queue.
Quick fix:
for (Message resultMessage = resultCollector.pollResult(); resultMessage != null; resultMessage = resultCollector.pollResult()) {

ramabit: @steckbrief thanks! :)

Flowdalic:
The provider registration is still missing: You need to add them to https://github.com/igniterealtime/Smack/blob/master/smack-experimental/src/main/resources/org.jivesoftware.smack.experimental/experimental.providers so that they are registered when Smack initializes itself.

White at it, please rename MamPacket to MamElements. I want to fade out the term "Packet" from Smack, as it has no definition in the XMPP world.

When done, please squash the commits into one.

ramabit:
@Flowdalic should i also change the name of the package "org.jivesoftware.smackx.mam.packet" to "org.jivesoftware.smackx.mam.elements"?

annovanvliet:
Good initiative and highly appreciated!

Which server did you use as a reference to test against?

Sometime ago I also made an implementation but not pushed it, while there was no consensus about the version of the XEP and the way the server it had implemented.

Flowdalic:
> to "org.jivesoftware.smackx.mam.elements"?

'element' please, so that it is consistent with what is done in the message_correct package.

ramabit:
@Flowdalic done

ramabit:
@annovanvliet thanks! :) I am testing it with a MongooseIM server

ramabit:
@Flowdalic what do you think about those last fixes? ^

ramabit:
@Flowdalic I found some bugs here. Fixes coming shortly

ramabit:
@Flowdalic Fixes done. You can continue testing it :)

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -0,0 +1,533 @@
...
+        }
+
+        MamQueryIQ mamQueryIQ = prepareMamQueryIQSet(dataForm, queryId);
+
+        addResultsLimit(max, mamQueryIQ);
+
+        return queryArchive(mamQueryIQ, 0);
+    }
+
+    private void addAdditionalFields(List<AdditionalField> additionalFields, DataForm dataForm) {
```
Flowdalic: Is there any reason why this method and the following are not static? Also if you design them null safe, then please go with the `if (foo == null) return` approach, do keep the level indentation down.

Flowdalic:
- Smack does not model two classes for request and result types of the same IQ. Please remove MamPrefsResultIQ, only MamPrefsIQ should be used.
- MamFinExtension: That extension can be removed. I guess you took it from my MAM patch. But this patch was written in a time when there was actually a <fin/> extension, i.e. the <fin/> was send as extension element within an message stanza. The new MAM version (0.4 IIRC) does it right and signals the end of the MAM query results with the IQ result of the IQ initating the query. I guess that's where your confusion comes from. So there should only be a MamFinIQ

I did not finish the review, because I've run out of time for today. So there is likely more to come, but the overall impression is good.

For small PRs I'd usually ask people to squash their commits, but since this is a rather larger patch, I'd ask you to add from now on changes on top of 846a (Ideally using `--fixup 57ba` if you want to be super git leet). This keeps it easier for me to review the delta. Once your branch is ready for merge, we'll squash merge it.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
ramabit: @Flowdalic I think that is not necessary to make them static. Those methods were made to refactor the "queryArchive" method.
I didn't understand why the "`if (foo == null) return` approach" would be correct. If one of those values is null, I have to continue with the query but without applying that filter.

Flowdalic: You think in the wrong direction. ;) It is not necessary to make them non-static i.e. virtual methods.

Instead of
```java
static void eventuallyDoSomething(Foo foo, Bar bar) {
  if (foo != null) {
      foo.someting(bar);
  }
}
```
you should write
```java
static void eventuallyDoSomething(Foo foo, Bar bar) {
  if (foo == null) return;
  foo.something(bar);
}
```

ramabit: Ok I'll do that 👍

ramabit:
@Flowdalic Several fixes done. I'll wait the feedback :)

naeems:
Hi Flow,
Please make it quick to finalize this PR. I am now a days trying to implement XEP-0313. But this implementation seems more concrete. So please finalize this asap.

Thanks

ramabit:
@Flowdalic At the automatic checks what seems to fail is a test that I didn't modify. Please fix it

Flowdalic:
More remarks in order of severity:

- `queryMamPrefs()`'s try block is broken, the code cannot work. This shows that an integration test for this PR would be really nice. Or at least some real world testing.
  - also the last LOC of the try block can be done autoside the try block
  - the whole code of the function seems to carry a lot of cruft, likely because it is copy&pasted from the other function, e.g. there is no need for extraTimeout
-  `AdditionalField` should be removed. I see no reason why those are not simply `FormField`s
- `withJid`, `neverJids` and Co are String types, when they should be `Jid` type
- prepareRetrievePreferencesStanza should be simply the no-argument constructor of MamPrefsIQ
- Don't announce the MAM feature, this is only for entities providing MAM (although, this is something I was doing wrong in the initial implementation)

ramabit:
@Flowdalic  Thanks for the feedback. I am testing MAM implementation using it on an app connected to a XMPP server. I'll fix those things you found and all the bugs I find while testing, I tell you when that is ready :)

ramabit:
@Flowdalic Several fixes done and tested with an app connected to a MongooseIM server. It is working fine for me. I'll wait your feedback

ramabit:
@Flowdalic At the automatic check failed 1 test that I didn't touch. (only with java7)

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamElements.java
```diff
@@ -167,8 +168,8 @@ public CharSequence toXML() {
             XmlStringBuilder xml = new XmlStringBuilder();
             xml.openElement("always");
 
-            for (String jid : alwaysJids) {
-                xml.element("jid", jid);
+            for (Jid jid : alwaysJids) {
+                xml.element("jid", jid.toString());
```
Flowdalic: I don't think you need to call `toString()` here

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamElements.java
```diff
@@ -201,8 +202,8 @@ public CharSequence toXML() {
             XmlStringBuilder xml = new XmlStringBuilder();
             xml.openElement("never");
 
-            for (String jid : neverJids) {
-                xml.element("jid", jid);
+            for (Jid jid : neverJids) {
+                xml.element("jid", jid.toString());
```
Flowdalic: dito

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamElements.java
ramabit: Ok I'll remove the .toString() call

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamElements.java
ramabit: Ok I'll remove the .toString() call

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -348,39 +346,24 @@ private MamQueryIQ prepareMamQueryIQGet(String queryId) {
...
         PacketCollector resultCollector = connection.createPacketCollector(resultCollectorConfiguration);
 
-        try {
-            connection.createPacketCollectorAndSend(mamQueryIq).nextResultOrThrow();
-            IQ iq = finIQCollector.nextResultOrThrow(connection.getPacketReplyTimeout() + extraTimeout);
-            mamFinIQ = MamFinIQ.from(iq);
-        } finally {
-            resultCollector.cancel();
-            finIQCollector.cancel();
-        }
```
Flowdalic: I wonder why you changed this method? Now the packet collectors logic is broken. You need to ensure the *both* collectors are canceled in *any* case. My original code ensured that. Your new code does not. You also need to use `setCollectorToReset()` to reset the timeout counter.

ramabit: I changed it because I tested with an app and a MongooseIM server and the fin collector wasn't working fine. Now it works

ramabit: If you want I can add the:
try {
...
} finally {
resultCollector.cancel();
}

Flowdalic: Well it appears the code is missing now two critical non-functional properties:
1. Cancel both collectors in any case, in order to avoid leaking any collector
2. setCollectorToReset() is not used

Especially 2. is required in the retrieval pattern MAM (and similar XEPs) uses:
IQ-REQ → M1 → ... → MN → IQ-RES
The collector for the messages (M1-MN) needs to reset the timeout of the collector of the IQ response (IQ-RES), as a MAM (and similar) queries could potentially return a large data set, and thus you want to extend IQ collectors timeout over the normal IQ request-response timeout as long as data is arriving.

ramabit: Ok, I will add those 2 things and test if still works fine

ramabit:
@Flowdalic please check the new fixes :)

ramabit:
@Flowdalic To add integration tests I need some help with this: https://community.igniterealtime.org/message/258122

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -0,0 +1,558 @@
...
+            mamManager = new MamManager(connection);
+            INSTANCES.put(connection, mamManager);
+        }
+
+        return mamManager;
+    }
+
+    private MamManager(XMPPConnection connection) {
+        super(connection);
+        ServiceDiscoveryManager.getInstanceFor(connection).addFeature(MamElements.NAMESPACE);
```
ramabit: @Flowdalic  Should I delete this line with "addFeature"?

ramabit:
@Flowdalic Integration test added ^

ramabit:
@Flowdalic Integration test added ^

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
Flowdalic: Yes

ramabit: done

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public MamIntegrationTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+        mamManager = MamManager.getInstanceFor(conOne);
+    }
+
+    @SmackIntegrationTest
+    public void mamTest() throws Exception {
+
+        if (!mamManager.isSupportedByServer()) {
```
Flowdalic: Should throw a test not supported exception in the *constructor*. See https://github.com/igniterealtime/Smack/blob/master/documentation/developer/integrationtest.md#testnotpossibleexception

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+            return;
+        }
+
+        Jid userOne = JidCreate.from(conOne.getUser().toString().split("/")[0]);
```
Flowdalic: If you want the bare JID then use `EntityFullJid.asEntityBareJid()` instead, which is type safe, more efficient than string parsing and less error prone.

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+        Jid userTwo = JidCreate.from(conTwo.getUser().toString().split("/")[0]);
+
+        MamQueryResult mamQueryResult = mamManager.queryArchive(userTwo);
+
+        if (mamQueryResult.forwardedMessages.isEmpty()) {
```
Flowdalic: A minimal integration test for MAM looks like:
1. useOne sends *one* message two userTwo
2. userTwo queries its archive
3. ensure that the last message in the archive it is exactly the message send (compare at least the stanza id and the body content)

Of course it would be really great to have at least one more sophisticated integration test, e.g. something that checks the functionality with more than one message.

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
```diff
@@ -0,0 +1,62 @@
...
+            Message message = new Message();
+            message.setBody("test-message");
+            message.setFrom(userOne);
+            message.setTo(userTwo);
+
+            conOne.sendStanza(message);
+            mamQueryResult = mamManager.queryArchive(userTwo);
+        }
+
+        Assert.assertFalse(mamQueryResult.forwardedMessages.isEmpty());
```
Flowdalic: `!isEmpty()` is a little bit to less checking of the result.

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
ramabit: fixed

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
ramabit: fixed

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
ramabit: done

smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
ramabit: fixed

ramabit:
@Flowdalic please check the new integration test :)

undev0:
Hi, @ramabit thank you for the contrib, I've some question about getting the last message for a roster. As you test it : 

`
MamQueryResult mamQueryResult = mamManagerConTwo.queryArchive(pageSize, null, null, userOne, null);`

It starts getting the messages from the first posted until the last one, which is not really performant.

Is there a way to get the following stanza request with your implementation in order to get only the last element ?

```
<iq type='set' id='x'>
  <query xmlns='urn:xmpp:mam:0'>
    <x xmlns='jabber:x:data' type='submit'>
      <field var='FORM_TYPE' type='hidden'>
        <value>urn:xmpp:mam:1</value>
      </field>
      <field var='with'>
        <value>jid@domain</value>
      </field>
    </x>
    <set xmlns='http://jabber.org/protocol/rsm'>
     <max>1</max>
     <before/>
    </set>
  </query>
</iq>
```

Thanks.

naeems:
Hi Ramabit,

I am trying to use your module to fetch messages from archive. I am using ejabberd server which supports XEP-313. I am sending following packet

<iq id='6S1WS-54' type='set'>
     <query xmlns='urn:xmpp:mam:1' queryid='481fabc8-d6c6-4d88-9b3e-157468baf652'>
         <x xmlns='jabber:x:data' type='submit'>
            <field var='FORM_TYPE' type='hidden'>
                <value>urn:xmpp:mam:1</value>
             </field>
             <field var='start'>
                 <value>2016-06-22T08:13:58.729+00:00</value>
             </field>
         </x>
     </query>
 </iq>

Its returning me only type='Chat' or type='normal' messages. I have message in archive which has type='groupchat' but its not returning me. Please suggest me what to do to get all type of messages. 

Thanks
Nomi

Flowdalic:
Please refrain from discussing the usage of the MAM API in this PR. The comments here are for the discussion about the code. You are welcome to ask your questions in the officlal Smack Support forum.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamFinIQ.java
```diff
@@ -0,0 +1,134 @@
...
+        return queryId;
+    }
+
+    @Override
+    protected IQChildElementXmlStringBuilder getIQChildElementBuilder(IQChildElementXmlStringBuilder xml) {
+        xml.optAttribute("queryid", queryId);
+        xml.optBooleanAttribute("complete", complete);
+        xml.optBooleanAttribute("stable", stable);
+        if (rsmSet == null) {
+            xml.closeEmptyElement();
```
Flowdalic: This should be `setEmptyElement()`. The reason this ever caused any issues is that Smack never puts an MamFinIQ on the wire.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamPrefsIQ.java
```diff
@@ -0,0 +1,159 @@
...
+
+        if (neverJids != null) {
+            MamElements.NeverJidListElement neverElement = new NeverJidListElement(neverJids);
+            xml.element(neverElement);
+        }
+
+        return xml;
+    }
+
+    public static MamPrefsIQ from(IQ iq) {
```
Flowdalic: Please remove that method. We use the `from()` pattern in Smack only to retrieve extension elements. In order to downcast an IQ, we prefer explicit casts.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/filter/MamResultFilter.java
```diff
@@ -0,0 +1,49 @@
...
+
+        if (mamResultExtension == null) {
+            return false;
+        }
+
+        String resultQueryId = mamResultExtension.getQueryId();
+        return ((queryId == null && resultQueryId == null) || (queryId != null && queryId.equals(resultQueryId)));
+    }
+
+    protected MamResultExtension getMamResultExtension(Message message) {
```
Flowdalic: That method appears to be unnecessary, simply use `from()` directly.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinIQProvider.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public MamFinIQ parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String queryId = parser.getAttributeValue("", "queryid");
+        boolean complete = Boolean.parseBoolean(parser.getAttributeValue("", "complete"));
+        boolean stable = Boolean.parseBoolean(parser.getAttributeValue("", "stable"));
+
+        boolean done = false;
+        RSMSet rsmSet = null;
+        while (!done) {
```
Flowdalic: Using a boolean as abort condition for the parsers main `while` loop is also a deprecated pattern. Instead replace this line with `outerloop: while(true) {`.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinIQProvider.java
```diff
@@ -0,0 +1,58 @@
...
+            int eventType = parser.next();
+            if (eventType == XmlPullParser.START_TAG) {
+                if (parser.getName().equals(RSMSet.ELEMENT)) {
+                    rsmSet = new RSMSetProvider().parse(parser);
+                }
+            } else if (eventType == XmlPullParser.END_TAG) {
+                if (parser.getName().equals("fin")) {
+                    done = true;
```
Flowdalic: And replace this line with `break outerloop;`

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinIQProvider.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class MamFinIQProvider extends IQProvider<MamFinIQ> {
+
+    @Override
+    public MamFinIQ parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String queryId = parser.getAttributeValue("", "queryid");
+        boolean complete = Boolean.parseBoolean(parser.getAttributeValue("", "complete"));
+        boolean stable = Boolean.parseBoolean(parser.getAttributeValue("", "stable"));
+
+        boolean done = false;
```
Flowdalic: This line can be removed.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -0,0 +1,91 @@
...
+        String defaultField = parser.getAttributeValue("", "default");
+
+        if (iqType == null) {
+            iqType = "result";
+        }
+
+        List<Jid> alwaysJids = null;
+        List<Jid> neverJids = null;
+
+        boolean done = false;
```
Flowdalic: Please remove the `done` boolean as explained above.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return new MamPrefsIQ(Type.fromString(iqType), alwaysJids, neverJids, defaultField);
+    }
+
+    private List<Jid> iterateJids(XmlPullParser parser, String listType) throws Exception {
+        List<Jid> jids = new ArrayList<>();
+        boolean done = false;
+
+        while (!done) {
```
Flowdalic: Dito

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinIQProvider.java
```diff
@@ -0,0 +1,58 @@
...
+        RSMSet rsmSet = null;
+        while (!done) {
+            int eventType = parser.next();
+            if (eventType == XmlPullParser.START_TAG) {
+                if (parser.getName().equals(RSMSet.ELEMENT)) {
+                    rsmSet = new RSMSetProvider().parse(parser);
+                }
+            } else if (eventType == XmlPullParser.END_TAG) {
+                if (parser.getName().equals("fin")) {
```
Flowdalic: The condition here should be `parser.getDepth() == initialDepth` as an integer comparison is (slightly) faster than a String comparison.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -0,0 +1,91 @@
...
+            int eventType = parser.next();
+            if (eventType == XmlPullParser.START_TAG) {
+                if (parser.getName().equals("always")) {
+                    alwaysJids = iterateJids(parser, "always");
+                }
+                if (parser.getName().equals("never")) {
+                    neverJids = iterateJids(parser, "never");
+                }
+            } else if (eventType == XmlPullParser.END_TAG) {
+                if (parser.getName().equals(MamPrefsIQ.ELEMENT)) {
```
Flowdalic: Dito

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -0,0 +1,91 @@
...
+
+        while (!done) {
+            int eventType = parser.next();
+            if (eventType == XmlPullParser.START_TAG) {
+                if (parser.getName().equals("jid")) {
+                    parser.next();
+                    jids.add(JidCreate.from(parser.getText()));
+                }
+            } else if (eventType == XmlPullParser.END_TAG) {
+                if (parser.getName().equals(listType)) {
```
Flowdalic: Dito

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamQueryIQProvider.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class MamQueryIQProvider extends IQProvider<MamQueryIQ> {
+
+    @Override
+    public MamQueryIQ parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String queryId = null;
+        String node = null;
+        DataForm dataForm = null;
+
+        if (parser.getName().equals(MamQueryIQ.ELEMENT)) {
```
Flowdalic: That is unusual, as the parser should be on START_TAG of the element. Why do you feel it is necessary to check for this condition?

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamResultProvider.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class MamResultProvider extends ExtensionElementProvider<MamResultExtension> {
+
+    @Override
+    public MamResultExtension parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String queryId = null;
+        String id = null;
+        Forwarded forwarded = null;
+
+        if (parser.getName().equals(MamResultExtension.ELEMENT)) {
```
Flowdalic: Here is also that unusual check.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamResultProvider.java
```diff
@@ -0,0 +1,62 @@
...
+            queryId = parser.getAttributeValue("", "queryid");
+            id = parser.getAttributeValue("", "id");
+
+            boolean done = false;
+            while (!done) {
```
Flowdalic: Please remove `done` boolean and replace the String comparison with an integer one.

Flowdalic:
Thanks, please now squash all commits of this branch into one. Note that the branch currently conflicts with `master`, and you have to resolve those conflicts.

An example of a typical git workflow to do this would be:
1. merge the official `master` branch into your branch and resolve the commits
2. create and switch to a new branch `mam-squashed`, and make sure that this branch HEAD is identical with the latest HEAD of the official master branch
3. use `git merge --squash <old-mam-branch>` to squash the commits of the `old-mam-branch` as a single commit into `mam-squashed`

Now you have to update this PR with that single commit. You would usually set the HEAD of this PRs source branch to that particular commit.

Flowdalic:
Thanks, your current author string looks like this:
```text
Author: ramabit <f.e.ramirez94@gmail.com
```
Is this what you want, or do you want your full name, e.g.
```text
Author: John Doe <john@doe.com>
```
as this will show up in some places and I feel you may want to take the credits for being the major author.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamFinIQProvider.java
```diff
@@ -0,0 +1,58 @@
...
+ * @author Fernando Ramirez
+ *
+ */
+public class MamFinIQProvider extends IQProvider<MamFinIQ> {
+
+    @Override
+    public MamFinIQ parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String queryId = parser.getAttributeValue("", "queryid");
+        boolean complete = Boolean.parseBoolean(parser.getAttributeValue("", "complete"));
+        boolean stable = Boolean.parseBoolean(parser.getAttributeValue("", "stable"));
```
Flowdalic: Careful when parsing. stable defaults to `true` according to xep313. Should have used ParserUtils.getBooleanAttribute().

Flowdalic:
I take over from here but will add notes about the things I found while testing your code.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -0,0 +1,561 @@
...
+     *            can be "roster", "always", "never" (look at the XEP-0313
+     *            documentation)
+     * @return the MAM preferences result
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     * @throws NotLoggedInException
+     */
+    public MamPrefsResult updateArchivingPreferences(List<Jid> alwaysJids, List<Jid> neverJids, String defaultField)
```
Flowdalic: defaultField should have been an enum. Never give the user a chance to use false values.

smack-experimental/src/test/java/org/jivesoftware/smackx/mam/MamFinProviderTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void checkMamFinProvider() throws Exception {
+        XmlPullParser parser = PacketParserUtils.getParserFor(exmapleMamFinXml);
+        MamFinIQ mamFinIQ = new MamFinIQProvider().parse(parser);
+
+        Assert.assertFalse(mamFinIQ.isComplete());
+        Assert.assertTrue(mamFinIQ.isStable());
+        Assert.assertNull(mamFinIQ.getQueryId());
+
+        RSMSet rsmSet = mamFinIQ.getRSMSet();
+        Assert.assertEquals(rsmSet.getAfter(), "09af3-cc343-b409f");
```
Flowdalic: The first argument of assertEquals should always  be the expected value.

Flowdalic:
Applied with 189cac072b8cedbcecca8bf2b3c5ec4368ea8cfa. And I've put some modifications on top with aeb385a022528b5cda3c3fe4136f0d0cd4c21179. I hope those do not break anything, but even if so, then we can still fix it. Merged it without intensive integration testing as I wanted to have some progress here.

Thanks @ramabit for your help.

ojuniour:
ANy Reason why I'm getting this error (even though the archives were successfully received?

`java.lang.ClassCastException: org.jivesoftware.smack.packet.EmptyResultIQ cannot be cast to org.jivesoftware.smackx.mam.element.MamFinIQ`

ramabit:
@ojuniour I would need more information about the fin IQ stanza received.
Please open a question or report an issue following this https://github.com/igniterealtime/Smack/wiki/How-to-ask-for-help,-report-an-issue-and-possible-solve-the-problem-yourself and adding more descriptive context

ojuniour:
Thanks for the fast response.. I'm gonna do as you said.. But just a quick question; what version of XMPP-0313 protocol did you implement? I would guess it's between between v0.3, 0.4, 0.4.1 and  v0.5? Because I receive an <IQ/> packet first, then <message/> data comes after, and finally a <fin/> to conclude..(mam1 namespace btw). 
 I will start a ticket at the link, and post the link here so you can follow (especially since you contributed most to the code.. so perhaps you have better understanding of the issue). Thank you.

Flowdalic:
Please do not hijack pull requests when having issues and use the forums instead. Thank you.

PR #77, opened 2016-05-10
Invalid presence priority should not crash connection

guusdk:
Packet parsing causes the connection to be disconnected when a presence stanza is parsed that contains an invalid priority value. Instead, the invalid value should be handled in a non-fatal way (similar as to how other problems are handled).

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
@@ -185,15 +185,15 @@ public final class Presence extends Stanza implements TypedCloneable<Presence> {
     }
 
     /**
-     * Sets the priority of the presence. The valid range is -128 through 128.
+     * Sets the priority of the presence. The valid range is -128 through 127.
      *
      * @param priority the priority of the presence.
      * @throws IllegalArgumentException if the priority is outside the valid range.
      */
     public void setPriority(int priority) {
-        if (priority < -128 || priority > 128) {
+        if (priority < -128 || priority > 127) {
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 2893c25b59c5565e5ff9ae236cf12f15400b8752

---

guusdk:
I am unsure why Travis fails. It appears unrelated to my code change.

Flowdalic:
The `setPriority()` fix looks good.

Masking parsing problems as you suggested with the change in PacketParserUtils is not a good idea (and so is catching `Exception` btw). I know this is a controversial stance so I'll just say here that the robustness principle has been proven wrong.

Instead Smack allows you to set a [parsing exception callback](https://www.igniterealtime.org/builds/smack/docs/latest/javadoc/org/jivesoftware/smack/AbstractXMPPConnection.html#setParsingExceptionCallback-org.jivesoftware.smack.parsing.ParsingExceptionCallback-). Which will prevent the parsing exception from causing a connection termination. It will swallow the invalid stream element though.

If you want you can refactor your commit so that it only includes the `setPriority()` fix, which I will happily merge. Travis failed because some of Smack's unit tests are unfortunately in-deterministic and thus faulty because they are not properly synchronized.

guusdk:
I simply mimicked what was done a few lines down, in the default label of the same switch statement. Consistency and all. :) There's something to be said to prevent naive implementations from being disconnected by making them receive stanzas that cannot be parsed, but hey, whatever floats your boat.

Flowdalic:
Closed via  fa10a01831f4cceddc2bf947b4ce11924af57a6c

PR #80, opened 2016-05-25
Attribute Support On ExtensionElement Added

s4kibs4mi:
There was no support for adding attribute on extension which is added.
Two new class added SimpleExtensionElement and SimpleExtensionElementProvider
To escape confusion I have created different classes.Also it will be flexible to use. No need to write parser in ExtensionElementProvider.

Uses : 
```
SimpleExtensionElement see = new SimpleExtensionElement("elementName", "nameSpace");
see.setAttribute("key", "value"); // to add attribute
see.setElement("key", "value"); // to add element
```

and in ProviderManager just add
`ProviderManager.addExtensionProvider("elementName", "nameSpace", new SimpleExtensionElementProvider();`

thats all.

### DIFF
--- a/smack-android/src/main/java/org/jivesoftware/smack/android/AndroidSmackInitializer.java
+++ b/smack-android/src/main/java/org/jivesoftware/smack/android/AndroidSmackInitializer.java
@@ -16,8 +16,6 @@
  */
 package org.jivesoftware.smack.android;
 
-import java.util.List;
-
 import org.apache.http.conn.ssl.StrictHostnameVerifier;
 import org.jivesoftware.smack.SmackConfiguration;
 import org.jivesoftware.smack.initializer.SmackInitializer;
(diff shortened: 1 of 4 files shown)


Flowdalic:
Thanks for your contribution. This PR has multiple issues, the most important first: Please don't introduce a new generic ExtensionElement (`SimpleExtensionElement`) instead extend `DefaultExtensionElement` with the functionality you are missing.

Furthermore:
- `gradle check` fails
- Unrelated whitespace/NOP changes
- Usage of non symbolic constants in the parser
- `label29` in parser. Not sure how this was introduced. Should probably be `outerloop` or something

Flowdalic:
There is no need to create another PR (#82) dealing with the same topic. You should have simply updated the existing one (eventually by force pushing to its source branch).

But as there is already another PR, I'm going to close this one.

PR #82, opened 2016-05-31
Attribute Support On ExtensionElement Added

s4kibs4mi:
There was no support for adding attribute on extension which is added.
Two new class added SimpleExtensionElement and SimpleExtensionElementProvider
To escape confusion I have created different classes.Also it will be flexible to use. No need to write parser in ExtensionElementProvider.

Uses :

```
SimpleExtensionElement see = new SimpleExtensionElement("elementName", "nameSpace");
see.setAttribute("key", "value"); // to add attribute
see.setElement("key", "value"); // to add element
```
and in ProviderManager just add
`ProviderManager.addExtensionProvider("elementName", "nameSpace", new SimpleExtensionElementProvider();`

thats all.


Flowdalic:
So, I just noticed that in Smack 4.2 I've added [`StandardExtensionElement`](https://www.igniterealtime.org/builds/smack/dailybuilds/javadoc/org/jivesoftware/smack/packet/StandardExtensionElement.html) a while ago. Isn't that already exactly what you want?

s4kibs4mi:
Yes it is. After mentioning it by me you have recently added it.

PR #84, opened 2016-06-30
XEP-0191: Blocking Command implementation

ramabit:
Documentation: http://xmpp.org/extensions/xep-0191.html

### DIFF
--- a/documentation/extensions/blockingcommand.md
+++ b/documentation/extensions/blockingcommand.md
@@ -1,0 +1,63 @@
+Blocking Command
+================
+
+Allows to manage communications blocking.
+
+  * Check push notifications support
+  * Get blocking list
+  * Block contact
+  * Unblock contact
+  * Unblock all
+
+
+**XEP related:** [XEP-0191](http://xmpp.org/extensions/xep-0191.html)
+
+
+Get an instance of Blocking Command Manager
+-------------------------------------------
+
+```
+BlockingCommandManager blockingCommandManager = BlockingCommandManager.getInstanceFor(connection);
+```
+
+
+Check blocking command support
+------------------------------
+
+```
+boolean isSupported = blockingCommandManager.isSupportedByServer();
+```
+
+
+Get block list
+--------------
+
+```
+List<Jid> blockList = blockingCommandManager.getBlockList();
+```
+
+
+Block contact
+-------------
+
+```
+blockingCommandManager.blockContact(jid);
+```
+*jid* is a `Jid`
+
+
+Unblock contact
+---------------
+
+```
+blockingCommandManager.unblockContact(jid);
+```
+*jid* is a `Jid`
+
+
+Unblock all
+-----------
+
+```
+blockingCommandManager.unblockAll();
+```
--- a/documentation/extensions/index.md
+++ b/documentation/extensions/index.md
@@ -69,6 +69,7 @@ Smack Extensions and currently supported XEPs of smack-extensions
 | Last Message Correction                     | [XEP-0308](http://xmpp.org/extensions/xep-0308.html) | Provides a method for indicating that a message is a correction of the last sent message. |
 | [Group Chat Invitations](invitation.md)   | n/a                                                      | Send invitations to other users to join a group chat room. |
 | [Jive Properties](properties.md)          | n/a                                                      | TODO |
+| [Blocking Command](blockingcommand.md)      | [XEP-0191](http://xmpp.org/extensions/xep-0191.html) | Communications blocking that is intended to be simpler than privacy lists (XEP-0016). |
 
 
 Experimental Smack Extensions and currently supported XEPs of smack-experimental
@@ -85,7 +86,7 @@ Experimental Smack Extensions and currently supported XEPs of smack-experimental
 | [Internet of Things - Discovery](iot.md)    | [XEP-0347](http://xmpp.org/extensions/xep-0347.html) | Describes how Things can be installed and discovered by their owners. |
 | Google GCM JSON payload                     | n/a                                                  | Semantically the same as XEP-0335: JSON Containers |
 | Client State Indication                     | [XEP-0352](http://xmpp.org/extensions/xep-0352.html) | A way for the client to indicate its active/inactive state. |
-| [Message Archive Management](mam.md)                  | [XEP-0313](http://xmpp.org/extensions/xep-0313.html) | Query and control an archive of messages stored on a server. |
+| [Message Archive Management](mam.md)        | [XEP-0313](http://xmpp.org/extensions/xep-0313.html) | Query and control an archive of messages stored on a server. |
 
 
 Legacy Smack Extensions and currently supported XEPs of smack-legacy
--- a/smack-experimental/src/main/resources/org.jivesoftware.smack.experimental/experimental.providers
+++ b/smack-experimental/src/main/resources/org.jivesoftware.smack.experimental/experimental.providers
@@ -166,5 +166,5 @@
       <namespace>urn:xmpp:iot:control</namespace>
       <className>org.jivesoftware.smackx.iot.control.provider.IoTSetResponseProvider</className>
     </iqProvider>
-
+    
 </smackProviders>
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
@@ -1,0 +1,248 @@
+/**
+ *
+ * Copyright 2016 Fernando Ramirez
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.blocking;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.AbstractConnectionListener;
+import org.jivesoftware.smack.ConnectionCreationListener;
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.SmackException.NoResponseException;
+import org.jivesoftware.smack.SmackException.NotConnectedException;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPConnectionRegistry;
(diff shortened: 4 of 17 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,177 @@
...
+
+        if (blockingCommandManager == null) {
+            blockingCommandManager = new BlockingCommandManager(connection);
+            INSTANCES.put(connection, blockingCommandManager);
+        }
+
+        return blockingCommandManager;
+    }
+
+    private BlockingCommandManager(XMPPConnection connection) {
```
Flowdalic: I don't think we need the full Manager pattern for BlockingCommand. It's such a simple XEP and the fact that the constructor does nothing further indicates that this is not needed. Simply make all methods static and add another parameter `XMPPConnection` to them. Look for at ClientStateIndicationManager, IIRC it does the same.

ramabit: @Flowdalic done

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,124 @@
...
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     */
+    public static List<Jid> getBlockingList(XMPPConnection connection)
+            throws NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException {
+        GetBlockListIQ getBlockListIQ = new GetBlockListIQ();
+        IQ responseIq = connection.createPacketCollectorAndSend(getBlockListIQ).nextResultOrThrow();
+        BlockListIQ blockListIQ = (BlockListIQ) responseIq;
```
Flowdalic: FYI: You don't need this extra line. nextResultOrThrow() is generic and does the cast for you. But you don't need to change it if you don't want to. Just wanted to point this out.

ramabit: Ok thanks :)

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/provider/BlockContactIQProvider.java
```diff
@@ -0,0 +1,59 @@
...
+public class BlockContactIQProvider extends IQProvider<BlockContactIQ> {
+
+    @Override
+    public BlockContactIQ parse(XmlPullParser parser, int initialDepth) throws Exception {
+        Jid jid = null;
+
+        outerloop: while (true) {
+            int eventType = parser.next();
+
+            if (eventType == XmlPullParser.START_TAG) {
```
Flowdalic: This should at least use switch-case. Ideally the parser name comparison below too, even if it has only one case at the moment.

Flowdalic:
From a data model POV `BlockListIQ` and `GetBlockListIQ` are the exact same IQs. No need to have seem in different classes. Please merge them, into `BlocklistIQ`.

Flowdalic:
Uhh, I just read a bit of XEP-191 and noticed that there is a blocklist push mechanism, similar to roster pushes, which you did not implement. Please implement that too, but this also means that you have to use the full manager idiom. Of course this also means that `getBlocklist()` - please rename the method, a "blockinglist" is something else - can work with  cached values.

ramabit:
@Flowdalic How should I implement that push mechanism? Give some help

Flowdalic:
> @Flowdalic How should I implement that push mechanism? Give me some help please

Register an IQ handler for the push. Then save the pushed blocking list in the manager. If the user calls `getBlocklist()` then first look if we already have one cached, only if not request the list from the server. Drop the cache if the connection goes down. Have a look at the privacy list manager, which does have the exact same caching mechanism, just for privacy lists.

ramabit:
@Flowdalic ^ push IQ handler implemented :)
The continuous integration check throws an error in a file that I didn't change

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,242 @@
...
+    /**
+     * Returns the block list.
+     * 
+     * @return the blocking list
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     */
+    public List<Jid> getBlockList()
```
Flowdalic: The views handed out to the user of the blocklist should be unmodifiable. Please wrap the blocklist with `Collections.unmodifiableList()`.

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,242 @@
...
+            throws NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException {
+
+        if (blockListCached != null) {
+            return blockListCached;
+        }
+
+        BlockListIQ blockListIQ = new BlockListIQ(null);
+        blockListIQ.setType(Type.get);
```
Flowdalic: IQs are per default of type `get`. This LOC can be removed.

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,242 @@
...
+     * @throws InterruptedException
+     */
+    public List<Jid> getBlockList()
+            throws NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException {
+
+        if (blockListCached != null) {
+            return blockListCached;
+        }
+
+        BlockListIQ blockListIQ = new BlockListIQ(null);
```
Flowdalic: Please add a non-parameter constructor for `BlockListIQ` and use this constructor here.

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
ramabit: Done

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
ramabit: Done

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
ramabit: Done

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,242 @@
...
+     * @throws NoResponseException
+     * @throws XMPPErrorException
+     * @throws NotConnectedException
+     * @throws InterruptedException
+     */
+    public List<Jid> getBlockList()
+            throws NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException {
+
+        if (blockListCached != null) {
+            return blockListCached;
```
Flowdalic: This still returns a modifiable view to internal state of the manager.

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        BlockListIQ blockListIQ = new BlockListIQ();
+        BlockListIQ blockListIQResult = connection().createPacketCollectorAndSend(blockListIQ).nextResultOrThrow();
+        blockListCached = blockListIQResult.getJids();
+        return (blockListCached == null) ? null : Collections.unmodifiableList(blockListCached);
```
Flowdalic: Methods like these should never return `null` when they could return just an empty list. This removes the burden from the caller to handle the empty list case by checking for `null`.
But this is not only true for the high lever parts of the API, i.e. here, but also for the lower ones: `BlockListIQ.getJids()` should also never return `null`, but `Collections.emptyList()`.

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
ramabit: Done

smack-experimental/src/main/java/org/jivesoftware/smackx/blocking/BlockingCommandManager.java
ramabit: Done

Flowdalic:
Thanks, please squash and update the PR. I'll take over from here.

ramabit:
@Flowdalic done :)

Flowdalic:
Sorry I forget to assign an issue to this. Please update the commit so that it contains SMACK-731 somewhere within its commit message.

ramabit:
@Flowdalic done

ramabit:
@Flowdalic the automatic checks are not working ok since the last addition ("VersionEye")

Flowdalic:
Did you loose the provider configuration for the blocking command providers in the `.providers` file when squashing the commits?

Flowdalic:
Hmm it appears that your code does not allow for multiple items in <block/>. From XEP-0191: "The <block/> element MAY contain more than one <item/> child."

ramabit:
@Flowdalic
1) The providers configuration is there. Check the tab "Files changed"
2) Ok, I will add support for multiple items

ramabit:
@Flowdalic multiple items support added

Flowdalic:
Ok, then please move the code from smack-experimental into smack-extensions, because -experimental is for experimental XEPs which XEP-0191 is not. This means that you must move the entry in `extenions.md` under the correct section too, and please make sure to add the entry in the right position, they are ordered by their XEP number. When done, please squash and update this PR.

ramabit:
@Flowdalic done

Flowdalic:
Needs squashing

ramabit:
@Flowdalic squash done

ramabit:
@Flowdalic the Travis CI is not working fine

Flowdalic:
> @Flowdalic the Travis CI is not working fine

Yep, I know. I think I've already explained to you that this is because some unit tests are not properly synchronized. Which makes them fail sometimes, especially when executed within low processing power environments like Travis CI.

ramabit:
@Flowdalic ok good to know, so I don't worry about that, hopefully everything seems to be ready for a merge ;)

Flowdalic:
Closed with 9554c7be26344baf2412c78b6bdf5fc18f0c5e6a and 98bacb144d126bfbcbee41650a49e1f3b1077d90

guusdk:
@ramabit Feel free to submit a patch for https://issues.igniterealtime.org/browse/OF-827 too :)

PR #85, opened 2016-07-12
Token-based reconnection implementation

ramabit:
Docs: http://www.xmpp.org/extensions/inbox/token-reconnection.html

Login with token:
`xmppConnection.login(token, resourcepart);`

Get tokens:
`TBRManager tbrManager = TBRManager.getInstanceFor(xmppConnection);`
`TBRTokens tbrTokens = tbrManager.getTokens();`
`String accessToken = tbrTokens.getAccessToken();`
`String refreshToken = tbrTokens.getRefreshToken();`

Get last refresh token obtained from "success" nonza:
`String refreshToken = xmppConnection.getXOAUTHLastRefreshToken();`

To avoid reconnection using token:
`xmppConnection.avoidTokenReconnection();`

### DIFF
--- a/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
+++ b/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
@@ -25,26 +25,6 @@ import java.io.Writer;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import org.jivesoftware.smack.AbstractXMPPConnection;
-import org.jivesoftware.smack.SmackException;
-import org.jivesoftware.smack.SmackException.NotConnectedException;
-import org.jivesoftware.smack.SmackException.ConnectionException;
-import org.jivesoftware.smack.XMPPException.StreamErrorException;
-import org.jivesoftware.smack.XMPPConnection;
-import org.jivesoftware.smack.XMPPException;
-import org.jivesoftware.smack.packet.Element;
-import org.jivesoftware.smack.packet.IQ;
-import org.jivesoftware.smack.packet.Message;
-import org.jivesoftware.smack.packet.Stanza;
-import org.jivesoftware.smack.packet.Nonza;
-import org.jivesoftware.smack.packet.Presence;
-import org.jivesoftware.smack.sasl.packet.SaslStreamElements.SASLFailure;
-import org.jivesoftware.smack.sasl.packet.SaslStreamElements.Success;
-import org.jivesoftware.smack.util.PacketParserUtils;
-import org.jxmpp.jid.DomainBareJid;
-import org.jxmpp.jid.parts.Resourcepart;
-import org.xmlpull.v1.XmlPullParser;
-import org.xmlpull.v1.XmlPullParserFactory;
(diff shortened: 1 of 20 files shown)


ramabit:
@Flowdalic Take a look at this :)
I have a question: how could I retrieve a refresh token from a "success" Nonza after authenticating? Because the actual Smack's authentication workflow doesn't allow me to do that I believe.

Flowdalic:
Hi @ramabit. Thanks for your contribution, but I'm sorry I'm not going to review or accept a PR that is full of non semantic changes. Please fix this.

smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
```diff
@@ -58,8 +59,8 @@
 import org.igniterealtime.jbosh.ComposableBody;
 
 /**
- * Creates a connection to an XMPP server via HTTP binding.
- * This is specified in the XEP-0206: XMPP Over BOSH.
+ * Creates a connection to an XMPP server via HTTP binding. This is specified in
```
Flowdalic: Non semantic change (of many)

ramabit:
@Flowdalic I understand, but the the automatic checks won't pass :/
should I ignore it?

Flowdalic:
> @Flowdalic I understand, but the the automatic checks won't pass :/
> should I ignore it?

I don't understand. You mean if you don't introduce non-semantic changes, then `gradle check` will fail? Every commit in `master` passes `gradle check`. If you base your work on `master` and don't introduce new checkstyle violations, then `gradle check` will succeed.

ramabit:
@Flowdalic Yes, you are right. But I can't use Command+Shift+F shortcut, and that makes not to introduce new checkstyle violations difficult and painful. But well, I will try

Flowdalic:
Most formatters have a setting to only format your own code. Which is what you usually *always* want. People, especially in FOSS projects, don't like to review commits with tons of unrelated non-semantic whitespaces changes. They just distract from the real changes.

ramabit:
@Flowdalic now is ok? Should I squash the commits?

ramabit:
@Flowdalic I am testing a way to retrieve the refresh token :)

ramabit:
@Flowdalic Fixes added and the problem retrieving the refresh token was solved

smack-core/src/main/java/org/jivesoftware/smack/tbr/element/TBRGetTokensIQ.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    public TBRGetTokensIQ(Jid to) {
+        super(ELEMENT, NAMESPACE);
+        this.setType(Type.get);
+        this.setTo(to);
+    }
+
+    @Override
+    protected IQChildElementXmlStringBuilder getIQChildElementBuilder(IQChildElementXmlStringBuilder xml) {
+        xml.rightAngleBracket();
```
Flowdalic: Must/Should be `setEmptyElement()`.

ramabit: @Flowdalic Done

PR #92, opened 2016-10-07
Instant Stream Resumption implementation

ramabit:
Documentation: http://xmpp.org/extensions/inbox/isr.html

**WARNING**: This implementation wasn't tested against a server yet.
However, it has many unit tests checking nonzas parsers and validations.

I send this PR in order to receive feedback and refine it.
fyi @Flowdalic

### DIFF
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/isr/HMAC.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/isr/HMAC.java
@@ -51,9 +51,8 @@ public class HMAC {
                 hash.append(hex);
             }
             digest = hash.toString();
-        } catch (UnsupportedEncodingException e) {
-        } catch (InvalidKeyException e) {
-        } catch (NoSuchAlgorithmException e) {
+        } catch (UnsupportedEncodingException | InvalidKeyException | NoSuchAlgorithmException e) {
+            throw new IllegalStateException("HMAC digest couldn't be done correctly.");
         }
         return digest;
     }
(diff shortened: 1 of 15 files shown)


smack-tcp/src/main/java/org/jivesoftware/smack/isr/HMAC.java
```diff
@@ -0,0 +1,60 @@
...
+            StringBuffer hash = new StringBuffer();
+            for (int i = 0; i < bytes.length; i++) {
+                String hex = Integer.toHexString(0xFF & bytes[i]);
+                if (hex.length() == 1) {
+                    hash.append('0');
+                }
+                hash.append(hex);
+            }
+            digest = hash.toString();
+        } catch (UnsupportedEncodingException e) {
```
Flowdalic: **Never** swallow Exceptions! Use a try/multi-catch and throw a unchecked exception, e.g. InvalidStateException in this case. In cases where the exception can be ignored, log it at least with a FINE/FINER/FINEST level.

ramabit: @Flowdalic fixed

smack-tcp/src/main/java/org/jivesoftware/smack/isr/HMAC.java
```diff
@@ -51,9 +51,8 @@ public static String hmacDigest(String msg, String keyString, String algo) {
                 hash.append(hex);
             }
             digest = hash.toString();
-        } catch (UnsupportedEncodingException e) {
-        } catch (InvalidKeyException e) {
-        } catch (NoSuchAlgorithmException e) {
+        } catch (UnsupportedEncodingException | InvalidKeyException | NoSuchAlgorithmException e) {
+            throw new IllegalStateException("HMAC digest couldn't be done correctly.");
```
ramabit: @Flowdalic done

PR #97, opened 2016-11-24
SMACK-741 : Ad-hoc command 'note' element 'type' attribute

annovanvliet:
is an optional attribute.


Flowdalic:
Thanks, already fixed with d6ebb884e7dcd2e0daab6b0c8f732b3c9d02b588

PR #103, opened 2017-02-07
presence mode:Invisible

hepizoj22:
http://xmpp.org/extensions/xep-0018.html


Flowdalic:
Thanks for your contribution. Unfortunately I've to reject this PR. XEP-0018 introduces a new presence type and thus modifies XMPP-IM (RFC 6121) routing rules. While XMPP is extensible, these kind of modifications are frowned upon because they harm interoperability, and I guess that this is the reason that XEP-0018's status is 'Rejected'..

PR #106, opened 2017-02-08
Make Smack buildable under windows.

annovanvliet:
Remove symbolic links and specify UTF-8 as character set.


Flowdalic:
Thanks for your contribution. I'm happy including the build.gradle changes, but I'm not willing to remove the symlinks just for Windows support. (Don't modern file and operating systems support from Microsoft support them by now?).

annovanvliet:
I wish it was, but only under strict conditions (admin account etc.)
But the main reason was for me the inconsistency, only some package-info.java where symlinks.

Flowdalic:
> But the main reason was for me the inconsistency, only some package-info.java where symlinks.

I also do love consistency, so feel encouraged to add symlinks for consistency. :)

annovanvliet:
Instead of removing, I added all lacking symbolic links in the smack-integration-test package, the place it make most sense to me to have them.

PR #108, opened 2017-02-17
updated connect method for BOSH connection

vfite:
Now in case when user was previously connected successfully and connection was closed on error if we  call "connect" method from ReconnectionManager as usual for tcp, connection won't be reestablished.
saslFeatureReceived checking doesn't pass if we try login outside "connect" method but in BOSHConnectionListener.
Besides login flow in this way becomes not clear and complicated. 
So moving login to one place like in tcp resolves issue.

Another issue - parsing bosh errors. Some error received from server isn't actually stream error. For example error with code 404.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,7 +1,7 @@
 // The name of the root project.
 // If we would not set the name, then gradle would use the directory
 // name of the root directory
-rootProject.name = 'Smack'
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: f8056387c3a906858ac3f12de5222c466fc2e4b4

---

PR #109, opened 2017-02-17
updated login flow for BOSH connection

vfite:
calling method "connect" on BOSH connection from ReconnectionManager after connection was closed abruptly causes Exception because saslFeatureReceived ckecking doesn't pass.  It happens because actual login starts from another place then connect - in BOSHConnectionListener.
So moving all related logic in one place to connect like in tcp connection resolves issue and becomes flow clear.

Another issue - parsing some bosh error. For ex error with invalid SID won't be parsed:
`<body type="error" xmlns="http://jabber.org/protocol/httpbind"><error type="cancel" code="404"><item-not-found xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/><text xml:lang="en" xmlns="urn:ietf:params:xml:ns:xmpp-stanzas">Invalid SID</text></error></body>`

### DIFF
--- a/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
+++ b/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
@@ -132,7 +132,7 @@ public class XMPPBOSHConnection extends AbstractXMPPConnection {
     }
 
     @Override
-    protected void connectInternal() throws SmackException {
+    protected void connectInternal() throws SmackException, XMPPException, IOException {
         done = false;
         try {
             // Ensure a clean starting state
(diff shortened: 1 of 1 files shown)


smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
```diff
@@ -197,6 +197,18 @@ protected void connectInternal() throws SmackException {
...
         callConnectionConnectedListener();
+
+        // Automatically makes the login if the user was previously connected successfully
+        // to the server and the connection was terminated abruptly
+
+        // in cases for reconnect:
+        // if do this asynchronously in BOSHConnectionListener saslFeatureReceived.check
+        // method doesn't pass
+        // and login flow becomes not clear and complicated.
+        if (wasAuthenticated) {
```
Flowdalic: Newer Smack versions do not automatically login any more when `connect()` is invoked. See idiomatic way is chain the call, e.g. `connection.connect().login()`. So please remove this.

smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
```diff
@@ -537,7 +538,12 @@ public void responseReceived(BOSHMessageEvent event) {
                                 }
                                 break;
                             case "error":
-                                throw new StreamErrorException(PacketParserUtils.parseStreamError(parser));
+                                //Some bosh error isn't stream error.
```
Flowdalic: That appears to be a sensible change. Could you rebase it in a single commit (and ideally issue an extra PR for it)? Looks like something we want to have in the 4.2 branch, for the upcoming 4.2.0 release in 2-3 weeks.

PR #114, opened 2017-03-15
XMPP-Connections have to be weak, so the GC can collect them

wglanzer:
If you have more connections / Smack-Clients inside one JVM this fix will take effect in reduced RAM-alloc

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/FaultTolerantNegotiator.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/FaultTolerantNegotiator.java
@@ -18,6 +18,7 @@ package org.jivesoftware.smackx.filetransfer;
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.ref.WeakReference;
 
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.XMPPConnection;
@@ -40,13 +41,13 @@ public class FaultTolerantNegotiator extends StreamNegotiator {
 
     private final StreamNegotiator primaryNegotiator;
     private final StreamNegotiator secondaryNegotiator;
-    private final XMPPConnection connection;
+    private final WeakReference<XMPPConnection> connection;
 
     public FaultTolerantNegotiator(XMPPConnection connection, StreamNegotiator primary,
             StreamNegotiator secondary) {
         this.primaryNegotiator = primary;
         this.secondaryNegotiator = secondary;
-        this.connection = connection;
+        this.connection = new WeakReference<>(connection);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/FaultTolerantNegotiator.java
```diff
@@ -18,6 +18,7 @@
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.ref.WeakReference;
```
Flowdalic: Are weak references here really necessary? `StreamNegotiator`s appear to be short lived objects.

smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/IBBTransferNegotiator.java
```diff
@@ -18,6 +18,7 @@
 
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.ref.WeakReference;
```
Flowdalic: Same as for FaultTollerantNegoiator.

smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/Socks5TransferNegotiator.java
```diff
@@ -20,6 +20,7 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PushbackInputStream;
+import java.lang.ref.WeakReference;
```
Flowdalic: Same as for FaultTollerantNegoiator.

wglanzer:
I just did a quick screenshot of the GC-Root:
gc-root 4 ibbtransfernegotiator
I think it is because of the "org.jivesoftware.smackx.filetransfer.FileTransferNegotiator#INSTANCES"-Map.
The value, the FileTransferNegotiator-Objects, got a hard reference on the XMPPConnection which is used as INSTANCES-Key. 
But the WeakHashMap only has weak KEYS and STRONG values -> The XMPPConnections won't be removed because of this reference.

Flowdalic:
Ahh right, see https://github.com/Flowdalic/Smack/commit/0a6843f41f6bc2f8fc52019868179279a9b44fdc

wglanzer:
Thank you :-)
How long will it take until this fix will be available in an official smack-version?

Flowdalic:
> How long will it take until this fix will be available in an official smack-version?

Smack's 4.2.1 release is scheduled to be in 2-3 months. *But* you can use 4.2.1-SNAPSHOT from Maven Central or the 4.2.1 unique snapshots from https://igniterealtime.org/repo, which already contain the fix.

PR #118, opened 2017-03-17
fixed AbstractJidTypeFilter.accept()

fermh2010:
With this fix AbstractJidTypeFilter.accept() uses jid from getJidToInspect() instead of always using stanza sender jid (which prevented ToTypeFilter from working properly)


Flowdalic:
Thanks, this looks like something for the 4.2 branch. Could you re-issue this PR with '4.2' as the destination branch?

Flowdalic:
Cherry picked as https://github.com/igniterealtime/Smack/commit/4fb14490770329679f146096e529fb21ba3f7c35

PR #123, opened 2017-04-07
Fix getOrCreateLeafNode for Prosody

vanitasvitae:
I caught the service-unavailable error instead of the never thrown PubSubAssertionError.
This fixes getOrCreateLeafNode for Prosody servers, since they return an error (https://prosody.im/issues/issue/805)

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
@@ -289,15 +289,15 @@ public final class PubSubManager extends Manager {
                     throw e2;
                 }
             }
+            if (e1.getXMPPError().getCondition() == Condition.service_unavailable) {
+                // This could be caused by Prosody bug #805 (see https://prosody.im/issues/issue/805). Prosody does not
+                // answer to disco#info requests on the node ID, which makes it undecidable if a node is a leaf or
+                // collection node.
+                LOGGER.warning("The PubSub service " + pubSubService
+                        + " threw an DiscoInfoNodeAssertionError, trying workaround for Prosody bug #805 (https://prosody.im/issues/issue/805)");
+                return getOrCreateLeafNodeProsodyWorkaround(id);
+            }
             throw e1;
-        }
-        catch (PubSubAssertionError.DiscoInfoNodeAssertionError e) {
-            // This could be caused by Prosody bug #805 (see https://prosody.im/issues/issue/805). Prosody does not
-            // answer to disco#info requests on the node ID, which makes it undecidable if a node is a leaf or
-            // collection node.
-            LOGGER.warning("The PubSub service " + pubSubService
-                            + " threw an DiscoInfoNodeAssertionError, trying workaround for Prosody bug #805 (https://prosody.im/issues/issue/805)");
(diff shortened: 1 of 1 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
```diff
@@ -289,16 +289,16 @@ public LeafNode getOrCreateLeafNode(final String id)
                     throw e2;
                 }
             }
+            if(e1.getXMPPError().getCondition() == Condition.service_unavailable) {
```
Flowdalic: Nit: Missing space between `if` and `(`

smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
```diff
@@ -307,7 +307,8 @@ private LeafNode getOrCreateLeafNodeProsodyWorkaround(final String id)
             return createNode(id);
         }
         catch (XMPPErrorException e1) {
-            if (e1.getXMPPError().getCondition() == Condition.conflict) {
+            if (e1.getXMPPError().getCondition() == Condition.conflict ||
+                    e1.getXMPPError().getCondition() == Condition.service_unavailable) {
```
Flowdalic: Do we need the service-service unavailable check here? I'd expect prosody and most servers to either create the node or throw a conflict error.

smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
```diff
@@ -289,16 +289,16 @@ public LeafNode getOrCreateLeafNode(final String id)
...
+                // This could be caused by Prosody bug #805 (see https://prosody.im/issues/issue/805). Prosody does not
+                // answer to disco#info requests on the node ID, which makes it undecidable if a node is a leaf or
+                // collection node.
+                LOGGER.warning("The PubSub service " + pubSubService
+                        + " threw an DiscoInfoNodeAssertionError, trying workaround for Prosody bug #805 (https://prosody.im/issues/issue/805)");
+                return getOrCreateLeafNodeProsodyWorkaround(id);
+            }
             throw e1;
         }
-        catch (PubSubAssertionError.DiscoInfoNodeAssertionError e) {
```
Flowdalic: I think this means that you can also delete the DiscoInfoNodeAssertionError and PubSubAssertionError class, no?

smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
vanitasvitae: I tested it and it gives the service-unavailable error only :/
I don't understand, why we retry to create the node in the workaround method anyway. I mean, we tried before in the method above and got to the workaround method because it failed, right?

smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/PubSubManager.java
vanitasvitae: Yep, I forgot about that

vanitasvitae: Fixed

PR #124, opened 2017-04-12
if an iq is received and there is no iq handler registered to handle …

bbaldino:
…it, an error reply will be sent but the invokeStanzaCollectorsAndNotifyRecvListeners method will still fall through and pass the iq to any stanza listeners.  the docs (and the path where an iq request handler is successfully found) state that iq messages shouldn't be forwarded to the listeners.  this change moves the return statement to the bottom of the 'if (packet instanceof IQ)' block so that iq messages are never forwarded to the listeners.

i'll warn that this code is untested as i don't have the android sdk installed so haven't been able to build.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
@@ -1133,15 +1133,15 @@ public abstract class AbstractXMPPConnection implements XMPPConnection {
                             }
                         }
                     });
-                    // The following returns makes it impossible for packet listeners and collectors to
-                    // filter for IQ request stanzas, i.e. IQs of type 'set' or 'get'. This is the
-                    // desired behavior.
-                    return;
                 }
                 break;
             default:
                 break;
             }
+            // The following returns makes it impossible for packet listeners and collectors to
+            // filter for IQ request stanzas, i.e. IQs of type 'set' or 'get'. This is the
+            // desired behavior.
+            return;
         }
 
         // First handle the async recv listeners. Note that this code is very similar to what follows a few lines below,


GitCop:

 * Commit: c1151e22dba73765867d40eae05e7ffa0f16fa71

---

GitCop:

 * Commit: d921934d093b8b123429a377958b6cf279108682

---

vanitasvitae:
Hi! First of all thank you for your contribution :)
Your code fails 8 tests, so I think it breaks some intended behaviour? Please take a look at the [travis build log](https://travis-ci.org/igniterealtime/Smack/builds/222875550) for more details.

bbaldino:
@vanitasvitae thanks for the response.  first i just wanted to check if my interpretation of the docs/intended behavior is correct?  an unhandled iq shouldn't be passed through, correct?  also, i think i was having a hard time running the unit tests locally because the full build required the android sdk, is there a way to run them without that?

vanitasvitae:
@bbaldino I don't think there is a way to run the tests without Android SDK. What you can do is update your PR and see, if travis succeeds ;)

Unfortunatelly I'm not very familiar with the IQ-handler code. @Flowdalic is the one to ask here.

Flowdalic:
> an unhandled iq shouldn't be passed through, right?

No, it shouldn't. You are right. Sorry for the late response.

But this change also requires some other changes (Probably mostly in ThreadedDummyConnection). And should probably also target the 4.2 branch.

smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
```diff
@@ -1133,15 +1133,15 @@ public void run() {
...
-                    // The following returns makes it impossible for packet listeners and collectors to
-                    // filter for IQ request stanzas, i.e. IQs of type 'set' or 'get'. This is the
-                    // desired behavior.
-                    return;
                 }
                 break;
             default:
                 break;
             }
+            // The following returns makes it impossible for packet listeners and collectors to
```
Flowdalic: This makes some unit test fail. Most likely due interference with ThreadedDummyConnection.

ibauersachs:
@Flowdalic I'm not sure the failing tests are related to the `ThreadedDummyConnection`. Since apparently `IQRequestHandler`s should deal with all IQs, `StanzaCollector`s will never receive an IQ reply anymore. This would make the widely used `XMPPConnection.createStanzaCollectorAndSend(IQ packet)` obsolete.

Options that run all tests successfully:
1. Move the `return` just one level down (instead of two as in this PR). IQs not of type `get` or `set` would thus still be handled as they were before the introduction of `IQRequestHandler`s. I assume this was the original idea, since it seems to be the only reason why the outer `switch(type)` is there.
2. Changing `invokeStanzaCollectorsAndNotifyRecvListeners` so that the `StanzaCollector`s are invoked before the `IQRequestHandler`s.

Do you want me to send a new PR?

Flowdalic:
> Since apparently IQRequestHandlers should deal with all IQs,

No, `IQRequestHandler` is only for IQ of type `get` or `set` (IQ requests). :)

> StanzaCollectors will never receive an IQ reply anymore. This would make the widely used XMPPConnection.createStanzaCollectorAndSend(IQ packet) obsolete.

The previous explanation renders this conclusion false.

The `return` where it is right now makes it possible for StanzaListeners and Collectors to collect IQ requests for which no `IQRequestHandler` is registered. Moving it two layers down, as done in this PR, would make it impossible to for any listener/collector to match any IQ stanza, no matter the IQ type. Moving it one layer down, as suggested by Ingo, would make `IQRequestHandlers` the only way to listen for IQ requests. This is what I plan to end up with. The current implementation was meant only for the transitional period.

ibauersachs:
Did we all simply miss the replyToUnkownIq option and this PR/issue can be closed?

bbaldino:
> Did we all simply miss the replyToUnkownIq option and this PR/issue can be closed?

been a while since i was looking at this, but the docs state that iq messages shouldn't be forwarded to invokeStanzaCollectorsAndNotifyRecvListeners (no matter what), so the intention here was to fix a scenario where an iq message did fall through to the listeners.  i don't remember (and didn't intend) any alteration to whether or not an unknown iq was replied to as part of this.

> Moving it two layers down, as done in this PR, would make it impossible to for any listener/collector to match any IQ stanza, no matter the IQ type.

i don't recall this being the case, as i was still processing iq messages with this change, but maybe (probably) i'm misunderstanding what you're saying here.

Flowdalic:
Closed in favor of https://github.com/igniterealtime/Smack/commit/8020fc14161c3c37aec29c338b6c60e75b527d16

PR #126, opened 2017-04-20
Update smack-repl gradle script

vanitasvitae:
replace deprecated "<<" semantic with "doLast".

This time target the correct branch


vanitasvitae:
I'm not so sure, whether this has been done correctly. Most times it works fine, but sometimes the build hangs for ~1-2 mins in the smack-repl module. I'd like, if someone with more expertise can take another look at this before it gets merged.

Flowdalic:
No risk no fun :)

PR #130, opened 2017-05-04
Presence Parser does only accept a valid priority number.

annovanvliet:
When an server or other party is sending a Presence with an invalid priority (e.g. 5225), it will trigger a reconnect of the smack client.

My proposed change adds a check during parsing which sets the priority to the default when the value is not in the correct range.

Receiving a presence with an invalid number does not interrupt the
connection.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
@@ -552,6 +552,8 @@ public class PacketParserUtils {
                     break;
                 case "priority":
                     int priority = Integer.parseInt(parser.nextText());
+                    if ( priority < -128 || priority > 127 )
+                        priority = 0;
                     presence.setPriority(priority);
                     break;
                 case "show":


annovanvliet:
Some history research suggested it has been an issue before.

See https://community.igniterealtime.org/message/98384
It was fixed in issue https://issues.igniterealtime.org/browse/SMACK-60

Flowdalic:
Works as intended. If you don't want to trigger a reconnect, set an parsing exception handler. This of course, would mean that subsequent systems like Managers or the Roster will never see that particular presence stanza.

laukner:
I forgot to say that this happening upon joining a room and still believe that Smack has to be fixed. 
All Smack-based clients that do not parse the exception handler will be vulnerable to DoS attacks!

I haven't done any debugging yet but think it happens in  

...smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java

Line 329:
```
	// Wait for a presence packet back from the server.
	// Use a bare JID filter, since the room may rewrite the nickname.

...
	try {
		presence = connection.createStanzaCollectorAndSend(responseFilter, joinPresence).nextResultOrThrow(conf.getTimeout());
	}

	catch (NotConnectedException | InterruptedException | NoResponseException | XMPPErrorException e) {
		// Ensure that all callbacks are removed if there is an exception
		removeConnectionCallbacks();
		throw e;
	}
```

PR #131, opened 2017-05-19
Fix NPE in hashCode() in Occupant when jid is null

alextov:


Flowdalic:
Thanks for your contribution. I've created [SMACK-764](https://issues.igniterealtime.org/browse/SMACK-764). Could you update the commit so that the issue key 'SMACK-765' is stated somewhere in the commit message body? Just something like "Fixes SMACK-765". Thank you.

alextov:
Sure! Added to the commit message.

PR #132, opened 2017-05-29
Make sure, archiving is activated

vanitasvitae:
The test failed when archiving preferences were set to never.
This PR makes sure, the preferences are set to always.


smack-integration-test/src/main/java/org/jivesoftware/smackx/mam/MamIntegrationTest.java
```diff
@@ -31,8 +27,13 @@
 import org.jivesoftware.smack.packet.Message;
 import org.jivesoftware.smackx.forward.packet.Forwarded;
 import org.jivesoftware.smackx.mam.MamManager.MamQueryResult;
+import org.jivesoftware.smackx.mam.element.MamPrefsIQ;
 import org.jxmpp.jid.EntityBareJid;
 
+import java.util.List;
```
Flowdalic: That just moves the imports down. Could you remove that from the commit?

PR #136, opened 2017-06-03
Fix typos and xml issues in jingle package

vanitasvitae:

PR #137, opened 2017-06-03
Add tests for jingle classes

vanitasvitae:
Depends on #135, #136


Flowdalic:
Thanks! Minor nit: Please don't reference anything githubish in the commit messages. I try to keep the git repo github agnostic. Mentioning Smack issue keys in commit messages, like SMACK-XXX are of course OK and actually required or at least encouraged

PR #138, opened 2017-06-07
Create alternative-session JingleReason

vanitasvitae:
The JingleReason class did not allow proper usage of the Reason `<alternative-session/>` since that had a child element `<sid/>`. This commit fixes the issue by introducing subclasses.

### DIFF
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/jingle/JingleReasonTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/jingle/JingleReasonTest.java
@@ -31,38 +31,54 @@ public class JingleReasonTest extends SmackTestSuite {
     @Test
     public void parserTest() {
         assertEquals("<reason><success/></reason>",
-                new JingleReason(JingleReason.Reason.success).toXML().toString());
+                JingleReason.Success.toXML().toString());
         assertEquals("<reason><busy/></reason>",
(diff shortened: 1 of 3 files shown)


smack-extensions/src/test/java/org/jivesoftware/smackx/jingle/JingleReasonTest.java
```diff
@@ -31,38 +31,54 @@
...
-                new JingleReason(JingleReason.Reason.failed_application).toXML().toString());
+                JingleReason.FailedApplication.toXML().toString());
         assertEquals("<reason><incompatible-parameters/></reason>",
-                new JingleReason(JingleReason.Reason.incompatible_parameters).toXML().toString());
-        assertEquals(JingleReason.Reason.alternative_session, JingleReason.Reason.fromString("alternative-session"));
+                JingleReason.IncompatibleParameters.toXML().toString());
+        assertEquals("<reason><alternative-session><sid>1234</sid></alternative-session></reason>",
+                JingleReason.AlternativeSession("1234").toXML().toString());
+        // Alternative sessionID must not be empty
+        try {
```
Flowdalic: Those two should go into an extra test method annotated with @Exepected("NullPointerException")

vanitasvitae: Ah, thats how you handle those :)

PR #144, opened 2017-06-16
Add Jingle File Transfer elements and JingleUtil class

vanitasvitae:
This PR adds a first dummy JingleFileTransferManager, as well as elements and their providers used in Jingle File Transfer (XEP-0234).

It also adds the class JingleUtil, which will later be used to generate Jingle requests/responses.


vanitasvitae:
Squashed.

PR #147, opened 2017-06-25
Fix OMEMO MUC receiving

vanitasvitae:
The method OmemoManager.getSender() was faulty and returned null for MUC
occupants, which lead to NPEs when receiving MUC messages with OMEMO.


Flowdalic:
Please change commit title to "Fix receiving OMEMO MUC messages"

vanitasvitae:
Done.

PR #151, opened 2017-07-05
fix SlotRequest to use attributes instead of elements

n8fr8:
XEP-363 Spec indicates the values needed for requesting an upload slot should be passed as attributes and not child elements

See: https://xmpp.org/extensions/xep-0363.html#request

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/element/SlotRequest.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/element/SlotRequest.java
@@ -82,10 +82,10 @@ public class SlotRequest extends IQ {
 
     @Override
     protected IQChildElementXmlStringBuilder getIQChildElementBuilder(IQChildElementXmlStringBuilder xml) {
+        xml.attribute("filename", filename);
+        xml.attribute("size", String.valueOf(size));
+        xml.optAttribute("content-type", contentType);
         xml.rightAngleBracket();
-        xml.element("filename", filename);
-        xml.element("size", String.valueOf(size));
-        xml.optElement("content-type", contentType);
         return xml;
     }
 }


n8fr8:
This has been tested with Prosody 0.9.12

vanitasvitae:
Looks good to me :+1:

vanitasvitae:
The syntax that was implemented before is used at least until version [0.2.5](https://xmpp.org/extensions/attic/xep-0363-0.2.5.html) and changed probably in version 0.3.0.
I'm not sure, whether it makes sense to implement a fallback mechanism.
On the other hand there was no namespace bump with the update to 0.3.0.

Flowdalic:
Thanks. As @vanitasvitae this was changed with xep363 0.3.0 in https://github.com/xsf/xeps/commit/3849f03db419a990c8321e6902b270021ec7a3dd

- We need to make this part of the compatibility layer Smack already has for xep363
- The commit should go in the 4.2 branch, so that it hits 4.2.1
- I've created SMACK-774, the commit message should mention the issue key

vanitasvitae:
Whoops, there was a namespace bump from `'urn:xmpp:http:upload'` to `'urn:xmpp:http:upload:0'`. Are there places where this is important?

n8fr8:
@Flowdalic okay, I can work on this over the next few days. 

Regarding the compatibility layer, do you mean related to support for the v2 namespace? Tie it into that logic that exists in the class now?

Flowdalic:
Thanks for your offer. Input from downstream projects is vital to Smack, and we welcome any contribution.

Nevertheless, I hope you don't mind that I took the liberty to prepare a patch myself: https://github.com/Flowdalic/Smack/commit/858ba3f82a2068bc7c1dcfeceaba1838ed315d82

n8fr8:
Wonderful! The compatibility support is great, and the improved XML implementation outstanding. Thanks for the quick fix.

PR #158, opened 2017-07-20
Show parsed IQ packets in EnhancedDebugger

ibauersachs:
Fixes SMACK-728


Flowdalic:
Thanks for pointing this out. I've taken this as motivation for a proper fix by reworking Smack's Debugger facilities: https://github.com/igniterealtime/Smack/commit/b8ee8d808f7d6ac13604cc210dbf875ed6fdc4f9

ibauersachs:
Unless I missed something, this rework is only in master. Could you please consider merging this PR for 4.2 anyway? Or is there a specific problem with it you want to avoid?

PR #159, opened 2017-08-01
Make add/remove for presence interceptors consistent

ibauersachs:
addPresenceInterceptor uses PresenceListener, make removePresenceInterceptor use this type too.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -1011,7 +1011,7 @@ public class MultiUserChat {
      *
      * @param presenceInterceptor the stanza(/packet) interceptor to remove.
      */
-    public void removePresenceInterceptor(StanzaListener presenceInterceptor) {
+    public void removePresenceInterceptor(PresenceListener presenceInterceptor) {
         presenceInterceptors.remove(presenceInterceptor);
     }
 


Flowdalic:
Thanks Ingo :)

Flowdalic:
I just noticed that this was issued against the 4.2 branch while I believe this is an API breaking change (you narrow the type). On the other hand, clirr did not complain and most users will use the method with the right type anyway. So I guess it's ok to have it in 4.2.

Flowdalic:
Argl, looking at the clirr reports it looks like they all to just compare smack-core. Maybe it's time to look for an replacement to clirr.

ibauersachs:
This is indeed API breaking, however the existing method doesn't work anyway. It's impossible to add a PresenceListener and the remove it as StanzaListener (i.e. unless the object in question implements both interfaces).

ibauersachs:
Re clirr: Maybe there's something similar to `com.github.siom79.japicmp:japicmp-maven-plugin:0.10.0` for Gradle?

Flowdalic:
> .e. unless the object in question implements both interfaces

Ahh right, I was somehow under the impression that this is the case, while a quick look at the code proved me wrong.

Flowdalic:
There is https://github.com/melix/japicmp-gradle-plugin, but it appears taht it has not the same convenience level as the clirr-gradle-plugin: The clirr-gradle-plugin automatically determines the base version to compare against (Which in Smack's case is not the lowest version with the same major version, but the lowest version with the same major and minor version, as Smack only guarantees API compatibility across patch levels.

ibauersachs:
I can't look into the details for japicmp right now, but the Maven version allows you to specify the base version (which would 4.2.0 for the 4.2 branch). This AFAIK doesn't automatically imply semver.

Flowdalic:
> but the Maven version allows you to specify the base version (which would 4.2.0 for the 4.2 branch).

Do I need to specify the base version or does the plugin automatically determine the base version (the clirr-gradle-plugin does the latter, see https://github.com/aalmiray/clirr-gradle-plugin/issues/3)?

ibauersachs:
The Maven version can do some automagical detection, but please don't ask how it works exactly :-)

PR #160, opened 2017-08-01
Get descriptive text from error without lang

ibauersachs:
A stanza with `<error><text>Some text</text></error>`, i.e. without a
xml:lang attribute, did not return the value 'Some text' in
getDescriptiveText().

Insert the empty string as key when the attribute is not present while
parsing the xml. When writing, omit the attribute if the key is the
empty string.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractError.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractError.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import org.jivesoftware.smack.util.Objects;
 import org.jivesoftware.smack.util.PacketUtil;
 import org.jivesoftware.smack.util.XmlStringBuilder;
 
@@ -85,6 +86,7 @@ public class AbstractError {
      * @return the descriptive text or null.
      */
     public String getDescriptiveText(String xmllang) {
+        Objects.requireNonNull(xmllang, "xmllang must not be null");
         return descriptiveTexts.get(xmllang);
     }
 
@@ -105,7 +107,8 @@ public class AbstractError {
             String xmllang = entry.getKey();
             String text = entry.getValue();
             xml.halfOpenElement("text").xmlnsAttribute(textNamespace)
-                    .xmllangAttribute(xmllang).rightAngleBracket();
+                    .optXmlLangAttribute(xmllang)
+                    .rightAngleBracket();
             xml.escape(text);
             xml.closeElement("text");
         }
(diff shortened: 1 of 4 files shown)


smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractError.java
```diff
@@ -85,7 +88,12 @@ public String getDescriptiveText() {
      * @return the descriptive text or null.
      */
     public String getDescriptiveText(String xmllang) {
-        return descriptiveTexts.get(xmllang);
+        try {
```
Flowdalic: If we say we never use `null` as key, we simply but a  `Objects.requireNotNull(xmllang)` guard before the `get()`. This would also avoid the try/catch block.

smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractError.java
```diff
@@ -70,6 +70,9 @@ public String getDescriptiveText() {
             descriptiveText = getDescriptiveText("en");
             if (descriptiveText == null) {
                 descriptiveText = getDescriptiveText("");
+                if (descriptiveText == null) {
+                    descriptiveText = getDescriptiveText(null);
```
Flowdalic: I wonder if it wouldn't be better if we *never* used `null` as key, and instead use the empty string. Then this 3 LOCs could be removed.

smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractError.java
```diff
@@ -105,7 +113,8 @@ protected void addDescriptiveTextsAndExtensions(XmlStringBuilder xml) {
             String xmllang = entry.getKey();
             String text = entry.getValue();
             xml.halfOpenElement("text").xmlnsAttribute(textNamespace)
-                    .xmllangAttribute(xmllang).rightAngleBracket();
+                    .xmllangAttribute("".equals(xmllang) ? null : xmllang)
```
Flowdalic: That one is a bit trickier. While there is `XmlStringBuilder.optXmlLangAttribute(String)` the method only checks if the argument is `null`. But I think it's safe to extend the semantic of `optXmlLangAttribute()` to use `StringUtils.isNotNullOrEmpty(String)`, which would make it suitable here. And would avoid the conditional operator construct.

ibauersachs:
Enforcing non-null keys is fine, but we also need to handle the case where the builder accepts an entire map object which might already contain null (like in one of the unit tests, setDescriptiveTexts). Do you want to create a deep copy of the map there and refuse it when a null-key is present?

ibauersachs:
@Flowdalic any feedback on the null-keys in the builder?

ibauersachs:
@Flowdalic updated to address your comments.

PR #161, opened 2017-08-05
Make XMPPError Condition to Type mapping consistent

ibauersachs:
The Javadoc for several `Condition`s did not match what was actually mapped. Some `Condition`s also had `Type`s that did not match RFC6120. When RFC6120 suggets multiple types, pick the first in the list. In case no type is suggested (`undefined-condition`), use the type in the example stanza.

A new unit test checks that the number of `Condition`s matches the number of entries in the mapping table. For 4.3 it might be better to remove the mapping table and specify the `Type` as a property (constructor parameter+field+getter) in the `Condition` enum. This would make the test obsolete and would gurantee that each `Condition` has a `Type`.

### DIFF
--- a/smack-core/src/test/java/org/jivesoftware/smack/packet/XMPPErrorTest.java
+++ b/smack-core/src/test/java/org/jivesoftware/smack/packet/XMPPErrorTest.java
@@ -1,0 +1,35 @@
+/**
+ *
+ * Copyright © 2017 Ingo Bauersachs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


smack-core/src/test/java/org/jivesoftware/smack/packet/XMPPErrorTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.lang.reflect.Field;
+import java.util.Map;
+
+import org.junit.Test;
+
+public class XMPPErrorTest {
+    @Test
+    public void testConditionHasDefaultTypeMapping() throws NoSuchFieldException, IllegalAccessException {
+        Field conditionToTypeField = XMPPError.class.getDeclaredField("CONDITION_TO_TYPE");
```
Flowdalic: How about making the CONDITION_TO_TYPE field package protected to avoid having to use reflection?

ibauersachs: Sure, I just thought it is better to keep things that shouldn't be accessible from actual code (i.e. not test code) really private.

PR #162, opened 2017-08-05
Replace unicode characters with encoded notation

ibauersachs:
Checkstyle cannot deal with UTF-8 [1], which makes running it on
smack-im impossible (at least on Windows).

[1] http://checkstyle.sourceforge.net/writingchecks.html#Limitations


Flowdalic:
I usually would look at your PRs starting from the oldest to the newest. I know that you have a few outstanding ones. Rest assured that they are not forgotten. :) But since I'm very busy right now and I'll  be on vacation next week just a short question: Would wrapping the lines around
```
// CHECKSTYLE:OFF
…
// CHECKSTYLE:ON
```
also solve the issue for you?

ibauersachs:
No, unfortunately that doesn't help. Checkstyle's parser just cannot deal with unicode, no matter whether the checks for a certain area are turned on or off.

Flowdalic:
Closing this as I decided the it is not worth the trade-off. Thanks for your contribution anyways.

PR #163, opened 2017-08-06
Make UnknownIqRequestReplyMode public

ibauersachs:
If it is not made public, then SmackConfiguration.(g|s)etUnknownIqRequestReplyMode cannot
be used.


smack-core/src/main/java/org/jivesoftware/smack/SmackConfiguration.java
```diff
@@ -179,7 +179,7 @@ public static void addSaslMech(String mech) {
         }
     }
 
-   /**
+    /**
```
Flowdalic: Please remove that whitespace-only change. Thank you.

PR #168, opened 2017-08-17
Add support for XEP-0166, XEP-0260, XEP-0261

vanitasvitae:
Reworked Jingle, Jingle S5B transport and Jingle IBB transport.


vanitasvitae:
closed in favor of #169

PR #169, opened 2017-08-20
Add support for XEP-0234 (Jingle File Transfer)

vanitasvitae:
This PR is based on #168 .

TODO:
* [x] Rebase on master
* [x] Squash


smack-extensions/src/main/java/org/jivesoftware/smackx/jingle/transport/jingle_s5b/JingleS5BTransport.java
```diff
@@ -432,8 +432,9 @@ private void handleCandidateUsed(JingleS5BTransportInfoElement info, JingleEleme
         }
 
         if (theirSelectedCandidate == null) {
-            LOGGER.log(Level.SEVERE, "ILLEGAL CANDIDATE ID!!!");
+            LOGGER.log(Level.SEVERE, "Unknown candidateID.");
```
Flowdalic: It appears that this error message (and others in this file) could need some more context. A simple start would be possibly to simply add `wrapping`.

smack-extensions/src/main/java/org/jivesoftware/smackx/jingle/JingleManager.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- * Copyright 2017 Florian Schmaus
+ * Copyright 2017 Paul Schaub
```
vanitasvitae: Completely rewrote the class, in case you are wondering ;)

vanitasvitae:
Apparently this has already been merged at some point.

PR #176, opened 2017-10-06
Fix getInstanceFor for offline usage.

vanitasvitae:
This should fix the issue of multiple OmemoStore account identities being generated due to the BareJid not being known before user login.

This PR changes the behaviour of `getInstanceFor(connection)` and `getInstanceFor(connection, deviceId)` by throwing an exception when the jid cannot be retrieved. To allow offline usage, `getInstanceFor(connection, bareJid, deviceId)` is introduced which the client can use to pass the bareJid they expect to smack-omemo.


vanitasvitae:
closed in favor of #177

PR #178, opened 2017-10-27
Makes xmpperror descriptive text optional as said in the rfc.

damencho:


Flowdalic:
Thanks for your contribution. I think this may be a candidate for Smack's '4.2' branch. If so, then please close this PR and create a new one targeting the 4.2 branch. If you don't care that this will only hit Smack 4.3, then please say so and I will merge it right way.

damencho:
Ok, I will do the PR in 4.2 in few minutes. Thanks.

PR #180, opened 2017-10-28
Updated help links

thakkaryash94:
replaced community links to discourse links

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,10 @@ Only a few users have access for for filling bugs in the tracker. New
 users should:
 
 1. Read ["How to ask for help or report an issue"](https://github.com/igniterealtime/Smack/wiki/How-to-ask-for-help,-report-an-issue-and-possible-solve-the-problem-yourself)
-2. [Create a forums account](https://community.igniterealtime.org/login!input.jspa?registerOnly=true) (only e-mail is a requirement, you can skip all the other fields).
+2. [Create a discourse account](https://discourse.igniterealtime.org/signup) (you can also sign up with your google account).
 3. Login to the forum account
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -27,10 +27,10 @@ Only a few users have access for for filling bugs in the tracker. New
 users should:
 
 1. Read ["How to ask for help or report an issue"](https://github.com/igniterealtime/Smack/wiki/How-to-ask-for-help,-report-an-issue-and-possible-solve-the-problem-yourself)
-2. [Create a forums account](https://community.igniterealtime.org/login!input.jspa?registerOnly=true) (only e-mail is a requirement, you can skip all the other fields).
+2. [Create a discourse account](https://discourse.igniterealtime.org/signup) (you can also sign up with your google account).
 3. Login to the forum account
```
vanitasvitae: This should probably say "Login to the account"

README.md
```diff
@@ -27,10 +27,10 @@ Only a few users have access for for filling bugs in the tracker. New
...
 4. Press New in your toolbar and choose Discussion
```
vanitasvitae: This step also changed

README.md
thakkaryash94: bullet starts with "Create" so I suppose it means sign up not sign in!

vanitasvitae: I meant step 3 :)

thakkaryash94: I checked on discourse website, it is written as "discourse is a modern forum", So I keep it as it is. but to use our forum they need discourse account, that's why I wrote "Create a discourse account" and kept "Login to the forum account" as it is.

vanitasvitae: Ah okay 👍

PR #182, opened 2017-11-16
Drop stream management state on StreamManagementException

mityada:
Stream management state should be dropped on StreamManagementException. Otherwise client may keep trying to reconnect with a bad SM state.


Flowdalic:
Thanks for your contribution. This looks like a candidate for the 4.2 branch, so that it goes into Smack 4.2.2. Please create a new PR targeting the 4.2 branch with this commit and close this one.

PR #185, opened 2017-11-16
Prevent race condition after stream resumption

mityada:
New stanzas sent directly after stream resumption might have been added to `unacknowledgedStanzas` _before_ the old unacknowledged stanzas are resent. This caused new stanzas to be sent twice and later led to a `StreamManagementCounterError`.

### DIFF
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
@@ -1155,8 +1155,7 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
                             if (!smSessionId.equals(resumed.getPrevId())) {
                                 throw new StreamIdDoesNotMatchException(smSessionId, resumed.getPrevId());
                             }
-                            // Mark SM as enabled and resumption as successful.
-                            smResumedSyncPoint.reportSuccess();
+                            // Mark SM as enabled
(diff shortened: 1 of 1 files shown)


smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
```diff
@@ -1172,6 +1169,9 @@ private void parsePackets() {
                             if (!stanzasToResend.isEmpty()) {
                                 requestSmAcknowledgementInternal();
                             }
+                            // Mark SM as enabled and resumption as successful.
```
Flowdalic: I think this is too late. It has probably go to right before the `sendStanzaInternal()` for-loop a few lines above, because `sendStanzaInternal()` calls `isSmEnabled()` which now would return `false. Thus, as far as I can tell know, the right place to mark the SM sync points as successful is *after* we drained the `unacknowledgedStanzas` but *before* we send them using `sendStanzaInternal()`.

mityada: You are right, I didn't notice this. But in this case new stanzas may be sent before unacknowledged stanzas. May be we should mark `smEnabledSyncPoint` as successful _before_ resending stanzas so `isSmEnabled()` will return `true` and mark `smResumedSyncPoint` _after_ so stanzas will always be sent in the correct order.

PR #188, opened 2017-12-13
Fix minor codestyle issues

vanitasvitae:
An early Christmas present :)
I skimmed through *all* of Smacks source code and made the following changes to improve code quality and satisfy myself/my IDE:

* Removed unnecessary generics
* Removed unnecessary visibilities in interface methods
* Removed unnecessary modifiers for enums and enum constructors
* Removed redundant variable initializers
* Make members private/static/final if sensible
* Fix typos in documentation and comment
* Simplified a few expressions

PR #190, opened 2017-12-19
Retain old API

vanitasvitae:
PR #188 changed the signatures of some methods in order to fix typos.
This PR restores the old API, by adding back the old signatures annotated as deprecated.


vanitasvitae:
I splitted the changes into two commits, because I'm not sure, if you think its necessary to merge the changes done to smack-jingle-old.

PR #191, opened 2017-12-23
Fix javadoc issues in some packages

vanitasvitae:
This PR fixes javadoc issues in all packages except smack-integration-test/.../omemo and smack-jingle-old.

Now there are no more warnings popping up when running `gradle javadoc`, except the following:

```
:smack-extensions:javadoc/path/to/Smack/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/MUCInitialPresence.java:183: warning - Tag @link: can't find History(int, int, int, Date) in org.jivesoftware.smackx.muc.packet.MUCInitialPresence.History
```

I'm unable to fix those, as my the syntax is like documented elsewhere. Maybe this is a bug?

```
/**
* Deprecated constructor.
* @deprecated use {@link #History(int, int, int, Date)} instead.
*/
@Deprecated
public History() {
...
}

public History(int maxChars, int maxStanzas, int seconds, Date since) {
...
}
```


Flowdalic:
Thanks a ton and merry christmas. :-)

PR #194, opened 2018-01-14
Add support for XEP-0394: Message Markup

vanitasvitae:
This PR adds support for [XEP-0394: Message Markup](https://xmpp.org/extensions/xep-0394.html).


Flowdalic:
Thanks! :) Created SMACK-794, please update the PR so that the commit message mentions the issue key somewhere.

smack-experimental/src/main/resources/org.jivesoftware.smack.experimental/experimental.providers
```diff
@@ -284,4 +284,11 @@
       <className>org.jivesoftware.smackx.eme.provider.ExplicitMessageEncryptionProvider</className>
     </extensionProvider>
 
+    <!-- XEP-0394: Message Markup -->
+    <extensionProvider>
+          <elementName>markup</elementName>
+          <namespace>urn:xmpp:markup:0</namespace>
+          <className>org.jivesoftware.smackx.message_markup.provider.MarkupElementProvider</className>
+        </extensionProvider>
```
Flowdalic: Nit: Indentation

vanitasvitae:
Done :)

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/element/BlockQuoteElement.java
```diff
@@ -0,0 +1,64 @@
...
+        return end;
+    }
+
+    @Override
+    public String getElementName() {
+        return ELEMENT;
+    }
+
+    @Override
+    public CharSequence toXML() {
```
Flowdalic: The `toXML()` method should always declare that it returns `XmlStringBuilder` if possible, allowing the runtime to take certain shortcuts. This applies to all classes of this PR.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/provider/MarkupElementProvider.java
```diff
@@ -0,0 +1,140 @@
...
+
+        while (true) {
+            int tag = parser.next();
+            String name = parser.getName();
+            int start, end;
+            switch (tag) {
+                case START_TAG:
+                    switch (name) {
+                        case BlockQuoteElement.ELEMENT:
+                            start = ParserUtils.getIntegerAttributeOrThrow(parser, BlockQuoteElement.ATTR_START,
```
Flowdalic: This pattern repeats. I wonder if it would be a good idea to create a static helper method `StartEnd paraseStartEndAttributes(XmlPullParser)`, where `StartEnd` is simply a class with the two start and end integers. Ideally we also have then a setter in the element class which takes this type.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/element/MarkupElement.java
```diff
@@ -0,0 +1,98 @@
...
+
+        xml.closeElement(this);
+        return xml;
+    }
+
+    /**
+     * Interface for child elements.
+     */
+    public interface MarkupChildElement extends NamedElement {
+
```
Flowdalic: If this interface would contain the static `ATTR_(START|END)`, then you can remove their multiple definitions in the implementing classes.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/MessageMarkup.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Mark a section of a message as deleted.
+     *
+     * @param start start index
+     * @param end end index
+     * @return builder
+     */
+    public MessageMarkup setDeleted(int start, int end) {
```
Flowdalic: It says "return builder", but it's not part of the build API while I also think it should be. This is true for the following methods too.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/MessageMarkup.java
```diff
@@ -0,0 +1,220 @@
...
+        children.addAll(spans);
+        children.addAll(quotes);
+        children.addAll(codes);
+        children.addAll(lists);
+        return new MarkupElement(children);
+    }
+
+    private static void verifyStartEnd(int start, int end) {
+        if (start >= end || start < 0) {
+            throw new IllegalArgumentException("Start value MUST be greater equal than 0 and MUST be smaller than end value.");
```
Flowdalic: It is always a good idea to also return the actual values which where verified in the message that you are returning, be it via an exception or the log subsystem.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/MessageMarkup.java
vanitasvitae: What do you mean? The `MessageMarkup` class *is* the builder, maybe you missed that?

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/element/MarkupElement.java
vanitasvitae: Added ATTR values to the interface. I think I'm not going to make it a class though, because that would make things more complicated.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/provider/MarkupElementProvider.java
vanitasvitae: I'm not sure whether it is worth to create those methods just now. That snippet wouldn't be reused too often and I think It would overall make the code more complicated than it is. Right now you can clearly see, what happens.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_markup/MessageMarkup.java
Flowdalic: Ahh now I see. Well I had expect it to be called MarkupElementBuilder then, this follows the typical pattern that
1. Builder API type names end with "Builder"
2. If you strip the "Builder" from the type name, you get the type that is build

Flowdalic:
Squashed and cherry picked as a729a7c43b9d3a8bd09744916d718675b6a54a80

PR #195, opened 2018-01-16
Add support for XEP-0382: Spoiler Messages

vanitasvitae:


smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/SpoilerManager.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.Map;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.spoiler.element.SpoilerElement;
+
+public final class SpoilerManager extends Manager {
```
Flowdalic: I think the manager should probably get auto initialized (see other managers for the corresponding Smack pattern). I am also wondering if it should provide methods to start and stop announcing support for spoiler messages (e.g. `startAnnouncingSupportForSpoilerMessages() and `stopAnnouncingSupportForMessages()`). But then again, I can't come up with a good use case for those methods, given that the manager would, once instantiated, start announcing support.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/SpoilerManager.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+        return manager;
+    }
+
+    /**
+     * Create an empty SpoilerElement.
+     *
+     * @return empty SpoilerElement
+     */
+    public SpoilerElement createSpoiler() {
```
Flowdalic: There is no need for those methods to be non-static. It is also common in Smack that those type of methods are part of the SpoilerElement class. I also don't see a reason why the API would need the with-argument variant of the `createSpoiler(…)` method, the user could simply use the constructor. And for the non-argument variant, I would probably go with a static final instance field, .e.g `SpoilerElement.EMPTY_INSTANCE`.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/provider/SpoilerProvider.java
```diff
@@ -0,0 +1,53 @@
...
+import org.xmlpull.v1.XmlPullParser;
+
+public class SpoilerProvider extends ExtensionElementProvider<SpoilerElement> {
+
+    @Override
+    public SpoilerElement parse(XmlPullParser parser, int initialDepth) throws Exception {
+        String lang = null;
+        String hint = null;
+
+        // TODO: Find a better way to get 'xml:lang' attribute value
```
Flowdalic: See `ParserUtils.getXmlLang()` :)

smack-experimental/src/test/java/org/jivesoftware/smackx/spoiler/SpoilerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        SpoilerManager m = SpoilerManager.getInstanceFor(connection);
+        SpoilerElement empty = m.createSpoiler();
+
+        assertNull(empty.getHint());
+        assertNull(empty.getLanguage());
+
+        assertXMLEqual(xml, empty.toXML().toString());
+
+        XmlPullParser parser = TestUtils.getParser(xml);
+        SpoilerElement parsed = new SpoilerProvider().parse(parser);
```
Flowdalic: The common Smack idiom is to have a public field with an instance of a provider, usually for testing purposes, e.g. `SpoilerProvider.INSTANCE`.

smack-experimental/src/test/java/org/jivesoftware/smackx/spoiler/SpoilerTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jivesoftware.smack.test.util.TestUtils;
+import org.jivesoftware.smackx.spoiler.element.SpoilerElement;
+import org.jivesoftware.smackx.spoiler.provider.SpoilerProvider;
+
+import org.junit.Test;
+import org.xmlpull.v1.XmlPullParser;
+
+public class SpoilerTest extends SmackTestSuite {
+
+    private static final XMPPConnection connection = new DummyConnection();
```
Flowdalic: XEPs like this usually don't need a connection instance for unit testing. And I think with the changes I suggested, this can go away.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/element/SpoilerElement.java
```diff
@@ -0,0 +1,101 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.spoiler.element;
+
+import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+import org.jivesoftware.smackx.spoiler.SpoilerManager;
+
+public class SpoilerElement implements ExtensionElement {
```
Flowdalic: The class's API is missing convenient methods to check whether or not an Message contains a spoiler extension element and to retrieve the metadata of an possibly existing spoiler extension element. For example something like
`static boolean containsSpoiler(Message)`
`static Map<String,String> getSpoilers(Message)`

I wonder if the API should make it possible to determine by a single `getSpoilers()` invocation if there was no spoiler, or if there was just a single empty spoiler. So `getSpoilers()` returns 'null' in case there are no spoilers, and the empty Map if there are just empty spoilers.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/provider/SpoilerProvider.java
vanitasvitae: Ah, I searched in XmlPullParser for it, since the corresponding setter is in XmlStringBuilder.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/element/SpoilerElement.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Create a new SpoilerElement with a hint about a content and a language attribute.
+     *
+     * @param language language of the hint.
+     * @param hint hint about the content.
+     */
+    public SpoilerElement(String language, String hint) {
+        this.language = language;
+        this.hint = hint;
```
Flowdalic: The constructor should check that hint is not null or empty in case language is not null or empty.

vanitasvitae:
I updated the PR to reflect your requested changes.

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/element/SpoilerElement.java
vanitasvitae: Done

smack-experimental/src/main/java/org/jivesoftware/smackx/spoiler/SpoilerManager.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Create a new SpoilerManager and add Spoiler to disco features.
+     *
+     * @param connection xmpp connection
+     */
+    private SpoilerManager(XMPPConnection connection) {
+        super(connection);
+        ServiceDiscoveryManager.getInstanceFor(connection).addFeature(NAMESPACE_0);
```
Flowdalic: I don't think we want to automatically announce support for spoiler messages. The XEP is pretty clear that if a spoiler message is received the content must only be displayed upon user's request. But this can not, or only with to much effort, be guranteed on the library level.

I think we should remove the automatic announcement of spoiler messages and instead provide a method the user explicitly has to invoke in order to announce support for spoiler messages.

As beneficial result, the whole automatic manager initialization boilerplate code can go away

Flowdalic:
Squashed and commited as https://github.com/igniterealtime/Smack/commit/ce19ea41140fa191a52e3b94c17b54ebee9d2603

PR #199, opened 2018-02-04
Add support for XEP-0372: References

vanitasvitae:
This PR adds support for XEP-0372: References to Smack.
The XEP seems to have some discrepancies, so there are some TODOs in the code, which will fixed once the author gives some feedback on how to deal with them.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/reference/ReferenceManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/reference/ReferenceManager.java
@@ -16,8 +16,6 @@
  */
 package org.jivesoftware.smackx.reference;
 
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
 
@@ -25,12 +23,7 @@ import org.jivesoftware.smack.ConnectionCreationListener;
 import org.jivesoftware.smack.Manager;
 import org.jivesoftware.smack.XMPPConnection;
 import org.jivesoftware.smack.XMPPConnectionRegistry;
-import org.jivesoftware.smack.packet.ExtensionElement;
-import org.jivesoftware.smack.packet.Stanza;
 import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
-import org.jivesoftware.smackx.reference.element.ReferenceElement;
-
-import org.jxmpp.jid.BareJid;
 
 public final class ReferenceManager extends Manager {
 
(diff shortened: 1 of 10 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/reference/ReferenceManager.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Add a reference to another users bare jid to a stanza.
+     *
+     * @param stanza stanza.
+     * @param begin start index of the mention in the messages body.
+     * @param end end index of the mention in the messages body.
+     * @param jid referenced jid.
+     */
+    public static void addMention(Stanza stanza, int begin, int end, BareJid jid) {
```
Flowdalic: Those static method should all be moved to `ReferenceElement`

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/element/ReferenceElement.java
```diff
@@ -0,0 +1,141 @@
...
+        if (begin != null && begin < 0) {
+            throw new IllegalArgumentException("Attribute 'begin' MUST NOT be smaller than 0.");
+        }
+        if (end != null && end < 0) {
+            throw new IllegalArgumentException("Attribute 'end' MUST NOT be smaller than 0.");
+        }
+        if (begin != null && end != null && begin >= end) {
+            throw new IllegalArgumentException("Attribute 'begin' MUST be smaller than attribute 'end'.");
+        }
+        if (type == null) {
```
Flowdalic: Nit: Use `Objects.requireNonNull()` in those cases

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/element/ReferenceElement.java
```diff
@@ -0,0 +1,141 @@
...
+                .optIntAttribute(ATTR_END, end != null ? end : -1)
+                .attribute(ATTR_TYPE, type.toString())
+                .optAttribute(ATTR_ANCHOR, anchor)
+                .optAttribute(ATTR_URI, uri);
+
+        if (child == null) {
+            return xml.closeEmptyElement();
+        } else {
+            return xml.rightAngleBracket()
+                    .append(child.toXML())
```
Flowdalic: I think just `.append(child)` should be sufficient.

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/provider/ReferenceProvider.java
```diff
@@ -0,0 +1,55 @@
...
+            if (eventType == XmlPullParser.START_TAG) {
+                String elementName = parser.getName();
+                String namespace = parser.getNamespace();
+                ExtensionElementProvider<?> provider = ProviderManager.getExtensionProvider(elementName, namespace);
+                if (provider != null) {
+                    child = provider.parse(parser);
+                }
+            }
+            if (eventType == XmlPullParser.END_TAG) {
+                return new ReferenceElement(begin, end, type, anchor, uri, child);
```
Flowdalic: I'd prefer if this parser would follow the same idiom all other parsers use, i.e. `break outerloop` here.

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/element/ReferenceElement.java
```diff
@@ -0,0 +1,141 @@
...
+    public enum Type {
+        mention,
+        data
+    }
+
+    private final Integer begin;
+    private final Integer end;
+    private final Type type;
+    private final String anchor;
+    private final String uri;
```
Flowdalic: If it is an URI, then why not use `URI` as type here?

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/ReferenceManager.java
vanitasvitae: I'm just thinking, whether this would lead to a simple API. Personally my first entry point to a XEP is always the manager class, that's why I put the method there. Shall I move the code to the element, but also keep a forwarding method in the manager?

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/element/ReferenceElement.java
vanitasvitae: I was not sure, whether the java URI class can be used as the xmpp uri. I'll fix that.

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/ReferenceManager.java
Flowdalic: > Shall I move the code to the element, but also keep a forwarding method in the manager?

No, as it would create unnecessary redundancy.

smack-experimental/src/main/java/org/jivesoftware/smackx/reference/element/ReferenceElement.java
vanitasvitae: Nope, the argument of `append()` cannot be an `ExtensionElement`.

Flowdalic:
Created [SMACK-799](https://issues.igniterealtime.org/browse/SMACK-799) for this.

- [ ] We need to make sure to mention the issue key in the final commit message.

vanitasvitae:
Rebased

PR #202, opened 2018-02-07
Add support for XEP-0392: Consistent Color Generation

vanitasvitae:
This PR adds a utility class `ConsistentColor`, which provides methods to generate consistent colors for a given String while applying different color vision deficiency corrections.


Flowdalic:
Applied as e0af3db1890677be92fec2c8f0660af572e28e37

PR #205, opened 2018-02-18
Add IntegrationTests for chat2 API

vanitasvitae:
* Test if IncomingMessageListeners are getting fired for incoming chat messages.
* Test if OutgoingMessageListeners are getting fired for outgoing chat messages.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/chat/IncomingMessageListenerIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/chat/IncomingMessageListenerIntegrationTest.java
@@ -1,0 +1,63 @@
+/**
+ *
+ * Copyright 2018 Paul Schaub
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.chat;
+
+import org.jivesoftware.smack.chat2.Chat;
+import org.jivesoftware.smack.chat2.ChatManager;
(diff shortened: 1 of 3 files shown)


smack-integration-test/src/main/java/org/jivesoftware/smackx/chat/IncomingMessageListenerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.After;
+import org.jxmpp.jid.EntityBareJid;
+
+public class IncomingMessageListenerIntegrationTest extends AbstractSmackIntegrationTest {
+
+    public IncomingMessageListenerIntegrationTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+    }
+
+    private final String body = StringUtils.randomString(16);
```
Flowdalic: Those three should all be local variables in the test method.

smack-integration-test/src/main/java/org/jivesoftware/smackx/chat/IncomingMessageListenerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+        ChatManager m2 = ChatManager.getInstanceFor(conTwo);
+
+        m2.addIncomingListener(listener);
+        Chat chat = m1.chatWith(conTwo.getUser().asEntityBareJid());
+        chat.send(body);
+        syncPoint.waitForResult(10 * 1000);
+    }
+
+    @After
+    public void cleanup() {
```
Flowdalic: Don't use an @After method for this, instead use a `finally` clause within the test method.

smack-integration-test/src/main/java/org/jivesoftware/smackx/chat/IncomingMessageListenerIntegrationTest.java
```diff
@@ -0,0 +1,64 @@
...
+        public void newIncomingMessage(EntityBareJid from, Message message, Chat chat) {
+            if (body.equals(message.getBody())) {
+                syncPoint.signal();
+            }
+        }
+    };
+
+    @SmackIntegrationTest
+    public void test() throws Exception {
+        ChatManager m1 = ChatManager.getInstanceFor(conOne);
```
Flowdalic: The ChatManagers should be fields of the test class. And they should be named after the patter the other tests use, i.e. `chatManagerOne` and `chatManagerTwo`.

vanitasvitae:
Squashed.

PR #206, opened 2018-02-18
Prevent race condition in ecaps test

vanitasvitae:
The EntityCapsTest.testPreventDiscoInfo() integration test was sometimes failing because of a race condition where the presence sent by ServiceDiscoveryManager.addFeature() was received *after* the caps cache had been dropped. This PR adds a presence listener to wait for presence before dropping the cache.


Flowdalic:
I've cleaned up your code a little bit with https://github.com/igniterealtime/Smack/pull/206/commits/c2d15d32caf35e4b5f30321784aeb5fd61cfe293, otherwhise this is good to go if you are ok with it.

Flowdalic:
Lesson learned: I'm never going to use github's "Squash and Merge" feature again. The resulting commit message is terrible

PR #209, opened 2018-02-28
Rename and deprecate XMPPConnection methods

fuentesj11:
Rename and deprecate XMPPConnections methods as described in SMACK-802

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
@@ -319,7 +319,7 @@ public interface XMPPConnection {
      *
      * @param packetListener the stanza(/packet) listener to notify of new received packets.
      * @param packetFilter the stanza(/packet) filter to use.
-     * @see #addPacketInterceptor(StanzaListener, StanzaFilter)
+     * @see #addStanzaInterceptor(StanzaListener, StanzaFilter)
      * @since 4.1
      */
     public void addSyncStanzaListener(StanzaListener packetListener, StanzaFilter packetFilter);
@@ -345,7 +345,7 @@ public interface XMPPConnection {
      * 
      * @param packetListener the stanza(/packet) listener to notify of new received packets.
      * @param packetFilter the stanza(/packet) filter to use.
-     * @see #addPacketInterceptor(StanzaListener, StanzaFilter)
+     * @see #addStanzaInterceptor(StanzaListener, StanzaFilter)
      * @since 4.1
     */
     public void addAsyncStanzaListener(StanzaListener packetListener, StanzaFilter packetFilter);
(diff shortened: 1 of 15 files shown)


fuentesj11:
I thought the methods were supposed to be deprecated and thus the Werror flag is failing the build.  Am I wrong or do I need to submit this a different way?

smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
```diff
@@ -391,15 +413,38 @@ public StanzaCollector createStanzaCollectorAndSend(StanzaFilter packetFilter, S
...
+    @Deprecated
     public void addPacketInterceptor(StanzaListener packetInterceptor, StanzaFilter packetFilter);
 
+    /**
+     * Registers a stanza(/packet) interceptor with this connection. The interceptor will be
+     * invoked every time a stanza(/packet) is about to be sent by this connection. Interceptors
+     * may modify the stanza(/packet) to be sent. A stanza(/packet) filter determines which packets
+     * will be delivered to the interceptor.
+     *
+     * <p>
```
vanitasvitae: I think you missed a </p> here.

fuentesj11: Thank you I completely missed that!

vanitasvitae:
>I thought the methods were supposed to be deprecated and thus the Werror flag is failing the build. Am I wrong or do I need to submit this a different way?

The build is probably failing, because various pieces of code still point to the now deprecated method. The proper way around this would *probably* be to change all calls to the old method to instead call the renamed method.

@Flowdalic knows best what to do though :)

smack-core/src/test/java/org/jivesoftware/smack/test/util/CharSequenceEquals.java
```diff
@@ -1,47 +0,0 @@
-/**
```
vanitasvitae: Why did you delete this class?

fuentesj11:
Sorry about all those errors.  I should have been much better at that.

vanitasvitae:
You can check for errors using `gradle check`.

fuentesj11:
Thank you!  I literally had everything that could wrong with gradle in Intellij happen.

smack-core/src/test/java/org/jivesoftware/smack/test/util/CharSequenceEquals.java
Flowdalic: Yep, why did this class go? If it's not longer used then we should consider removing it, but with an extra commit.

fuentesj11:
Is anyone aware there is a CharsequenceEquals and CharSequenceEquals class?

vanitasvitae:
>Is anyone aware there is a CharsequenceEquals and CharSequenceEquals class?

Yeah, looks like the `CharsequenceEquals` class can be removed.

fuentesj11:
It that an issue/ticket/task in itself?  I just looked through all branches and it appears to be in all branches.

Flowdalic:
Squash please

PR #210, opened 2018-03-03
Don't crash when a user sends an invalid presence mode

ntrrgc:


smack-core/src/main/java/org/jivesoftware/smack/util/PacketParserUtils.java
```diff
@@ -558,7 +558,11 @@ public static Presence parsePresence(XmlPullParser parser)
                 case "show":
                     String modeText = parser.nextText();
                     if (StringUtils.isNotEmpty(modeText)) {
-                        presence.setMode(Presence.Mode.fromString(modeText));
+                        try {
+                            presence.setMode(Presence.Mode.fromString(modeText));
+                        } catch (IllegalArgumentException ex) {
+                            /* Invalid mode string received. */
```
vanitasvitae: Please don't just silently swallow the exception. I would at least log it using `LOGGER` :)

Flowdalic:
Thanks for your contribution.

Smack's design philosophy is to fail hard per default if the core specification is not followed, as it's the case here, in order to raise awareness of the violation. Another option Smack provides is to set a parsing exception callback using [AbstractXMPPConnection.setParsingExceptionCallback()](https://download.igniterealtime.org/smack/docs/latest/javadoc/org/jivesoftware/smack/AbstractXMPPConnection.html#setParsingExceptionCallback-org.jivesoftware.smack.parsing.ParsingExceptionCallback-). I think that it would hurt interoperability of a open-standard federated system if the simply ignore faulty parts of stanzas, hence I'm not going to merge this.

ntrrgc:
> Smack's design philosophy is to fail hard per default if the core specification is not followed, as it's the case here, in order to raise awareness of the violation.

So you are saying that, because of that philosophy, Smack will crash whenever it receives invalid data from the network?

That may work if all your XMPP clients are within the control of a single organization and only talk to each other, but that's not a very robust design whenever you need to expose your client to the open Internet (e.g. by joining a public or federated server) because any entity &mdash; malicious or not, could trivially do a DoS on your service.

Yes, I'm aware that standards are to be followed if interoperability is the goal, but for security purposes it's simply not acceptable to crash upon untrusted input from the Internet.

And even if DoS were not a concern, fixing the clients is not always a viable solution, especially when you are not the person running them.

In my case, I'm running a bot in a small, open XMPP chatroom, mostly gajim users. Everything was fine until one day a Psi+ user joined and my bot kept crashing and restarting because of this issue.

Luckily enough Psi+ has since fixed the bug, but even then, what could I do? Tell them "please, go away and don't set foot here until you have built an updated Psi+ for your operating system that does not have this bug which is crashing my bot"? What if this was in a bigger room where affected users joined more frequently? Just give up on having a bot altogether? What if there was no readily available patch for the affected client? What if the user could not or did not want to update their client? It's not as if they were crashing my bot on purpose. They would have it very easy if that was indeed their intention though.

Flowdalic:
> So you are saying that, because of that philosophy, Smack will crash whenever it receives invalid data from the network?

Not crash, it will cleanly disconnect the connection. Setting a non-throwing parsing exception callback will prevent that.

ntrrgc:
Using `setParsingExceptionCallback()` I was indeed able to handle and log this problem, and hopefully other similar ones. Thanks for the hint.

Neustradamus:
@ntrrgc, what was the Psi/Psi+ build version?
Because team said there is not problem...

ntrrgc:
@Neustradamus I don't remember, but by the time I reported this the issue was already fixed in Psi+ upstream IIRC; but it was nevertheless a problem with older versions such as those packaged by some distros.

PR #211, opened 2018-03-04
Add support for XEP-0319: Last User Interaction

vanitasvitae:
This PR adds Elements and Providers for [XEP-0319: Last User Interaction in Presence](https://xmpp.org/extensions/xep-0319.html).

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/last_interaction/provider/IdleProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/last_interaction/provider/IdleProvider.java
@@ -1,0 +1,40 @@
+/**
+ *
+ * Copyright © 2018 Paul Schaub
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


vanitasvitae:
@Flowdalic this PR only adds the elements and providers for the XEP. What do you think of also adding a manager class which has a timer. That timer can be reset by the user by calling `.activity()` or something. Once that method did not get called for more than 5 mins or so (configurable), the timer fires and a presence stanza with an idle element will be sent automatically.

vanitasvitae:
Another idea would be to create methods to automatically get the last user interaction of contacts by caching last interactions in a map.

vanitasvitae:
Resolved conflicts.

Flowdalic:
The commit message is missing the issue key (SMACK-636). Please add it.

smack-extensions/src/main/java/org/jivesoftware/smackx/last_interaction/provider/IdleProvider.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.jivesoftware.smack.provider.ExtensionElementProvider;
+import org.jivesoftware.smackx.last_interaction.element.IdleElement;
+
+import org.jxmpp.util.XmppDateTime;
+import org.xmlpull.v1.XmlPullParser;
+
+public class IdleProvider extends ExtensionElementProvider<IdleElement> {
+
+    public static IdleProvider TEST_INSTANCE = new IdleProvider();
```
Flowdalic: Nit: Missing `final` keyword

vanitasvitae:
Done.

PR #215, opened 2018-03-11
Add and register BoBExtensionProvider

fuentesj11:
SMACK-770 Issue

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/bob/provider/BoBExtensionProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/bob/provider/BoBExtensionProvider.java
@@ -16,17 +16,17 @@
  */
 package org.jivesoftware.smackx.bob.provider;
 
+import org.jivesoftware.smack.provider.ExtensionElementProvider;
 import org.jivesoftware.smackx.bob.BoBHash;
 import org.jivesoftware.smackx.bob.element.BoBExtension;
 import org.jivesoftware.smackx.xhtmlim.XHTMLText;
 import org.jivesoftware.smackx.xhtmlim.packet.XHTMLExtension;
-import org.jivesoftware.smackx.xhtmlim.provider.XHTMLExtensionProvider;
 import org.xmlpull.v1.XmlPullParser;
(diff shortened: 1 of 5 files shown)


smack-extensions/src/test/java/org/jivesoftware/smackx/bob/BoBExtensionProviderTest.java
```diff
@@ -0,0 +1,52 @@
...
+            + "<body xmlns='http://www.w3.org/1999/xhtml'>"
+            + "<p>Yet here's a spot."
+            + "<img alt='A spot'"
+            + "src='cid:sha1+8f35fef110ffc5df08d579a50083ff9308fb6242@bob.xmpp.org'/>"
+            + "</p></body></html></message>";
+
+    @Test
+    public void parseTest() throws Exception {
+        XmlPullParser parser = TestUtils.getParser(sampleBoBExtensionIM);
+        BoBExtension boBExtension = new BoBExtensionProvider().parse(parser);
```
vanitasvitae: Please create a public static BobExtensionProvider object in the BobExtensionProvider class for test purposes and use that instead of creating a new object here. (See for example [here](https://github.com/igniterealtime/Smack/commit/ce19ea41140fa191a52e3b94c17b54ebee9d2603#diff-556a2c9b3fe8fdd648cfd39e8a7ddf26R30) and [here](https://github.com/igniterealtime/Smack/commit/ce19ea41140fa191a52e3b94c17b54ebee9d2603#diff-98c8a53ff77a7c9a9459526f254e370fR52)).

smack-extensions/src/main/java/org/jivesoftware/smackx/bob/provider/BoBExtensionProvider.java
```diff
@@ -0,0 +1,63 @@
+/**
+ *
+ * Copyright 2018
```
Flowdalic: A copyright statement without an indication who owns the copyright is worthless. If you don't want to have the copyright, assign it to the "Ignite Realtime Foundation".

fuentesj11:
I need a better understanding of the actual XML in this case and I think I am a bit away from obtaining that knowledge.

PR #216, opened 2018-03-15
Fix jingle action

matteocampana:


smack-jingle-old/src/main/java/org/jivesoftware/smackx/jingleold/packet/Jingle.java
```diff
@@ -354,7 +354,7 @@ protected IQChildElementXmlStringBuilder getIQChildElementBuilder(IQChildElement
             buf.append(" responder=\"").append(getResponder()).append('"');
         }
         if (getAction() != null) {
-            buf.append(" action=\"").append(getAction().name()).append('"');
+            buf.append(" action=\"").append(getAction().toString()).append('"');
```
Flowdalic: I'm missing an explanation regarding what is wrong about using `name()` here?

vanitasvitae: I'd assume that since `JingleActionEnum` takes a String value in its constructor, `action.name()` would return for example `CONTENT_ACCEPT` which is the name of the enum, while `action.toString()` will return the value of actionCode (in this case `content-accept`), which is what you actually want to put into the XML.

matteocampana: Yes, exactly. 
And because of that, the parser sets `action=null` due to `JingleActionEnum` `getActiongetAction(String inActionCode)` method.
See also this discussion: https://discourse.igniterealtime.org/t/null-pointer-exception-in-jinglemanager/73226/5

smack-core/src/test/java/org/jivesoftware/smack/test/util/CharsequenceEquals.java
```diff
@@ -1,47 +0,0 @@
-/**
```
vanitasvitae: Can you revert this change, as it is already taken care for in #209?

matteocampana: Yes sure, I will do it.

vanitasvitae:
Can you please hard reset your branch to master, only apply the changes from 7cb9175 and force push to get rid of the other commits?

Flowdalic:
Picked onto 4.2 as https://github.com/igniterealtime/Smack/commit/713f0438df49c83da400975a054d384701da6e16

PR #218, opened 2018-04-01
Update/Edit connections.md

fuentesj11:


vanitasvitae:
Thank you very much! One nit: Can you change the title to something more descriptive like "Update connections documentation"? Same for #219, #220, #221, #222, #223 and #224. Right now its a little hard to keep track of which PR takes care for what :)

PR #230, opened 2018-04-08
first commit, add support of manual ACKment of received MESSAGE stanza

okyfirmansyah:
Hi, I write this PR to add support of manual ACKing of received <message> stanza(requires stream management). This to ensure handshaking/transactional nature in application level so no message will be lost if application is still not completed its message processing(for e.g storing to persistent database/queue, updating UI, etc). This especially true if application defer the handle of message to other module/thread and error/crash/kill is encountered like happened in our application, so maybe this can help other implementation too.

The way handshaking will be done in application is using Smack's **XMPPTcpConnection.markHandledStanzaId()** to explicitly state that the stanza/message is already processed completely. This will trigger the usual Stream Management's receive height counter increment, so the server can now safely remove the message from its offline/retry buffer. 

This **markHandledStanzaId()** can be called regardless order of actual receiption, because under the hood, the modification maintain list of all previously unacknowledged stanzas. Only when first and the consecutives stanzas are already marked as handled, the height counter will be updated and the stanzas will be removed from list. In this modification, <IQ> and <PRESENCE> will always be ACKed automatically regardless configuration.

This behaviour is configurable by mean of XMPPTcpConnectionConfiguration. Unless **Builder.setManualRcvAck(true)** is called, the receive ACK will be done automatically as default/current behaviour.

### DIFF
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
@@ -45,8 +45,10 @@ import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.AbstractMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ConcurrentHashMap;
@@ -261,7 +263,7 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
      * </p>
      */
     private long clientHandledStanzasCount = 0;
-
+    private Queue< Map.Entry<String, Integer> > unacknowledgedRcvdStanzas;
     private BlockingQueue<Stanza> unacknowledgedStanzas;
 
     /**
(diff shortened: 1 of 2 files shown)


vanitasvitae:
There are some checkstyle issues with your PR. Please make sure your branch passes the lint checker by running `gradle check` successfully.

Flowdalic:
I don't see a reason to have such an API nor do I want it. The user shouldn't be able to mess at such a low level functionality, not does he need to. If you feel like you would need such methods then you possbily are either not using IQs or using them wrong.

okyfirmansyah:
I am not sure about this IQ you are talking about. This manual ACK mechanism is for ensuring reliability of **MESSAGE** receiving part, not **IQ** part.

Without some sort of capability to let application mark received messages manually(and out of order), we are forced to make sure persistent storing of received messages as soon as it arriving(in the listener) in order to avoid message loss. This is too restrictive, for e.g if we need to propagate the received message across different modules/threads before deeming message as successfully handled(hence ack), or if we want to update UI ASAP as message is received but defer writing them to persistent storage later for performance reason(like if messages come in burst), there will always risk of client 'losing' the messages otherwise.

We are developing xmpp-based chat app on both ios(**xmppframework**) and android. On the former library, there is capability to manually ACK the received messages and we are extensively use it(the library writer itself said it is very critical function for any reliable messaging system). Under stress-testing the receiving part, we see that message sometimes 'loss' in android under combination of poor network connection+abusive user behaviour of app(killing app randomly), but this has never happen in ios app. We are now very dependent on that manual ACK capability, because it works great.

I can see in SMACK, there is **StanzaAcknowledgedListeners** for sending part to ensure reliability, but why providing some kind of symmetric solution in receiving side, in form of **markHandledStanzaId()**, considered to be low-level solution?

okyfirmansyah:
Yes we use MAM for archiving/reloading older messages, but I don't think it is what we are looking for to ensure reliability of received messages. How can we detect if a message is lost in first place? do we keep polling the MAM? do we build sequence missing detection or any other solution on top of message protocol? I could not find something practical, except using Stream Management/ACK control which I believe should serve this purpose.

PR #233, opened 2018-04-21
Fix extensions table

vanitasvitae:


Flowdalic:
What is broken here? Please mention it also in the commit message.

PR #235, opened 2018-04-22
Add BodyElement + Provider

vanitasvitae:
This PR adds a BodyProvider + BodyElement to `smack-core`.
The `BodyElement` extends `Message.Body`, whose constructors visibility had to be changed to `package` instead of `private`. `Message.Body` is also no longer declared as `final`.
The difference between `BodyElement` and `Message.Body` is, that the `.toXML()` method of the latter will never add a namespace of "jabber:client" to the generated XML, while `BodyElement` will. `BodyElement` can therefore be added as a child element to elements other than `Message`.


Flowdalic:
This PR should now be, at least partly, obsolete by https://github.com/igniterealtime/Smack/commit/5ab2903c3289a302efab62e3225188f6e8c26c99

PR #238, opened 2018-05-16
fix: Cleans the multiUserChats map.

bgrozev:
We use Smack in a server side application, which creates and joins a new `MultiUserChat` as part of its health check. Health checks run often (every few seconds) and as a result the `multiUserChats` map in `MultiUserChatManager.java` fills up with `WeakReferece`s which are eventually cleared, but never removed from the map. After a long run we've observed >2 million entries, which leads to problems.

This patch solves the problem by cleaning the map by removing all weak references with `null` referent. The cleaning is run once every 50 calls to `put`.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
@@ -491,8 +491,7 @@ public final class MultiUserChatManager extends Manager {
      * @param <V> The value type.
      */
     private static class CleaningWeakReferenceMap<K, V>
-        extends HashMap<K, WeakReference<V>>
-    {
+        extends HashMap<K, WeakReference<V>> {
         private static final long serialVersionUID = 0L;
 
         /**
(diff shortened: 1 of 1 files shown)


Flowdalic:
Thanks for your contribution. This appears to be a good candidate for the 4.3 branch so that it hits Smack 4.3.0. Could you put `CleaningWeakReferenceMap` into its on file under `smack.util`?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -479,4 +480,58 @@ void addJoinedRoom(EntityBareJid room) {
...
+
+        /**
+         * Removes all cleared entries from this map (i.e. entries whose value
+         * is a cleared {@link WeakReference}).
+         */
+        private void clean() {
+            Iterator<Entry<K, WeakReference<V>>> iter = entrySet().iterator();
+            while (iter.hasNext()) {
+                Entry<K, WeakReference<V>> e = iter.next();
+                if (e != null && e.getValue() != null
```
Flowdalic: I wonder if that `e != null` check is really necessary.

vanitasvitae:
@damencho afaik the next release will be 4.3.

damencho:
And when we can expect it?

vanitasvitae:
There are already two beta releases out.

bgrozev:
I will make the changes when I am back in the office on Monday.

Regards,
Boris

PR #239, opened 2018-05-21
Add method to delete directory

vanitasvitae:
This functionality is needed by the OMEMO tests of the file based store. It will also be needed for testing the OX store, so I figured it would be cool to have it in smack-core.


Flowdalic:
If it is (currently) only required by test code, then it should be put somewhere under `src/test/java`. I'm going to merge that anyways as is.

PR #242, opened 2018-06-12
fix: Prevent attempt to construct invalid address

guusdk:
When no join was properly registered, a nickname will not be defined. In that case, attempting
to construct the from address for the 'leave' presence stanza will result in:

   java.lang.IllegalArgumentException: The Resourcepart must not be null

This commit prevents that, by verifying that the nickname is non-null, before sending that
stanza.


GitCop:

 * Commit: 4b170309be2e6fca662160dc4bf0205ce20cbea5

---

PR #243, opened 2018-06-20
Add utility method to get temp directory

vanitasvitae:
This method might come in handy in use-cases like tests which write files to disk.


Flowdalic:
Should this go into `src/test`?

vanitasvitae:
I moved the method into the newly created class `FileTestUtil`.

PR #249, opened 2018-07-03
Avoid adding multiple storage hints to a message

vanitasvitae:
Since ExtensionElements are stored in a list in the message, duplicate entries are possible. This commit avoids adding a storage hint to the message in case it already contains one.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoPermanentStoreHint.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoPermanentStoreHint.java
@@ -60,7 +60,7 @@ public final class NoPermanentStoreHint extends MessageProcessingHint {
             // No need to set the no-permanent-store hint when a no-store hint is already set.
             return;
         }
-        setExplicitly(message);
+        message.overrideExtension(INSTANCE);
     }
 
     public static void setExplicitly(Message message) {
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoCopyHint.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoCopyHint.java
@@ -56,6 +56,6 @@ public final class NoCopyHint extends MessageProcessingHint {
     }
 
     public static void set(Message message) {
-        message.addExtension(INSTANCE);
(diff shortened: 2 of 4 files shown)


Flowdalic:
I think this should simply use `Message.overrideExtension(ExtensionElement` instead of `addExtension()`.

smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoPermanentStoreHint.java
```diff
@@ -56,11 +56,7 @@ public static boolean hasHint(Message message) {
...
     public static void set(Message message) {
-        if (StoreHint.hasHint(message)) {
-            // No need to set the no-permanent-store hint when a no-store hint is already set.
-            return;
-        }
-        setExplicitly(message);
+        message.overrideExtension(INSTANCE);
     }
 
     public static void setExplicitly(Message message) {
```
vanitasvitae: I'm not sure if this method is still needed now, but I kept it in for now in order to not break the API.

vanitasvitae:
I reworked the PR, using `overrideExtension` in the StoreHint, as well as other hints.

smack-experimental/src/main/java/org/jivesoftware/smackx/hints/element/NoPermanentStoreHint.java
Flowdalic: Help me here: Why do you think that it is no longer needed?

vanitasvitae: Does it make sense to have multiple NoPermanentStoreHints in a message?
Users can still use `set` in order to set a hint on the message.

Flowdalic: I just noticed, after reading the comment within the function, that it checks for the store hint but this is the NoPermanentStore class. Hence I believe you should re-add the `if (StoreHint.hasHint(message)) return`.

vanitasvitae: Readded the check

PR #252, opened 2018-07-17
Add Objects.requireNonNullOrEmpty

vanitasvitae:
Utility method to check, whether a collection is null or empty.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/Objects.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/Objects.java
@@ -16,6 +16,8 @@
  */
 package org.jivesoftware.smack.util;
 
+import java.util.Collection;
+
 public class Objects {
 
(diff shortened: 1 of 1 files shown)


akrherz:
This will come off as pedantic here, but I think this should be `requireNonNullNorEmpty`  as nor is the logical 'and is not' .

vanitasvitae:
You're probably right. I followed the pattern in [StringUtils](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/util/StringUtils.java#L447). Logically it is supposed to word `¬(null || empty)`.

Flowdalic:
Thanks for your feedback @akrherz (which is always appreciated). I also think that you are right. We should at one point deprecate the `…OrEmpty()` variants in `Objects` and `StringUtils` in favor of `…NorEmpty()`.

vanitasvitae:
I'll quickly do that :)

PR #260, opened 2018-07-31
Fix duplicate NS in stanza error child element

vanitasvitae:
This PR fixes an inconvenience, where an IQ with the implicit namespace `jabber:client` would append the namespace to an error child element like this:

```
<iq (xmlns='jabber:client) <!-- in parenthesis since the NS is implicit --> ... >
    <error xmlns='jabber:client' <!-- this NS is too much --> ... />
</iq>
```

The fix I found is to hand the IQs namespace down to the error element to prevent appending it twice.


Flowdalic:
Closing in favor of https://github.com/igniterealtime/Smack/commit/a3fcbdbf5af531c6ead3b310b6b9d69e321269eb

PR #263, opened 2018-08-14
Reconnection bug fixed

dhavalpatelasticsinc:


dhavalpatelasticsinc:
Reconnection manager bug fixed.

While reconnection, in between connection and login function, net loss then "connection.isConneced()" function retuen true. so authentication not done and reconnection become currupted.

ingjuanocampo:
This seems to be a very fair PR, currently we are facing same issue and perhaps this could help us. 
@dhavalpatelasticsinc Thank you for the fix, please update the format of the comments so Travis can build.

miguelhincapie:
Yeah I think this can help and be used as a work around to [OF sends unavailable presence](https://discourse.igniterealtime.org/t/android-smack-client-suddenly-sends-unavailable-presence-for-no-reason-then-server-thinks-im-offline/81366)

Flowdalic:
I do not see what this is supposed to fix. If `isReconnectionIsPossible()` returns false between a connect() and login() call how does " reconnection become currupted."? Closing this. Feel free to comment if you believe this is wrong, but please provide a detailed problem descriptions and elaborate how this changes fixes it.

miguelhincapie:
Hi @Flowdalic, I have been using smack in a real app that only can be accessed through wi-fi, but signal and stability of wi-fi are so poor. In this environment the app needs to reconnect too often. What we found out is sometimes smack said he is connected when that wasn't true at all, in other cases smack tried to re-use an old connection object but it seems to has old TLS values that can't be used anymore in the new tries.

So what I did at the end was avoid re-using connection at all.

Note: this can't be replicated in a stable environment.

Flowdalic:
> smack said he is connected when that wasn't true at all

Yep, that is to be expected. Such a method will never be able to *not* produce false positives. Use `pingMyServer()` if you want to know if the connection is still established (at the time the method call was started).

> in other cases smack tried to re-use an old connection object but it seems to has old TLS values that can't be used anymore in the new tries.

I would be love to see a more detailed report of this.

I am not sure if your post is a general comment or a response to my "Feel free to comment if you believe this is wrong" (your post doesn't seem to relate to this PR?), but nevertheless this should have been posted in the forums ideally.

PR #264, opened 2018-08-18
Do not allow OMEMO bundles without prekeys

vanitasvitae:
> elements and managing of the corresponding Signal sessions is
handled by the Signal library. The biggest problem with this
approach was that the Signal library accepted messages without a
one-time prekey, which OMEMO should never do (since the server
will never “run out” of one-time prekeys). Combined with the fact
that the signed prekeys never get removed/updated, this means that
there was no forward secrecy for PreKeySignalMessages.

### DIFF
--- a/smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/element/OmemoBundleElement.java
+++ b/smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/element/OmemoBundleElement.java
@@ -20,6 +20,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.util.StringUtils;
 import org.jivesoftware.smack.util.XmlStringBuilder;
 import org.jivesoftware.smack.util.stringencoder.Base64;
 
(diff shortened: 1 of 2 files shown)


vanitasvitae:
Maybe we should check if there are identity key, signed prekey etc present too? Might be good to fail early.

smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/provider/OmemoBundleVAxolotlProvider.java
```diff
@@ -96,6 +97,11 @@ else if (inPreKeys && name.equals(PRE_KEY_PUB)) {
                     break;
             }
         }
+
+        if (preKeys.isEmpty()) {
+            throw new XmlPullParserException("OMEMO Bundle MUST contain at least one prekey.");
```
Flowdalic: This should be a SmackException as it is not thrown from within the XML pull parser.

vanitasvitae: I noticed, that this change might be somewhat obsolete, as [this line](https://github.com/igniterealtime/Smack/blob/master/smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/util/OmemoKeyUtil.java#L126) already makes sure that we do not use empty bundles. Still, it might be good to fail early on, right?

Flowdalic:
Now, I have looked closer at this, I believe the right approach is probably to check this in the constructor, throwing an illegal argument exception if prekeys are empty.

smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/element/OmemoBundleElement.java
```diff
@@ -76,18 +96,33 @@ public OmemoBundleElement(int signedPreKeyId, String signedPreKeyB64, String sig
      * @param preKeys HashMap of preKeys
      */
     public OmemoBundleElement(int signedPreKeyId, byte[] signedPreKey, byte[] signedPreKeySig, byte[] identityKey, HashMap<Integer, byte[]> preKeys) {
+        if (signedPreKeyId <= 0) {
```
Flowdalic: That is exactly the kind of code duplication that you want to avoid. The usual solution is to change this constructor to a static constructor method: `public static OmemoBundleElement newOmemoBundleElement(int signedPreKeyId, byte[] signedPreKey, byte[] signedPreKeySig, byte[] identityKey, HashMap<Integer, byte[]> preKeys)` which internally used the constructor above.

Flowdalic:
Please dedup your code

vanitasvitae:
> The usual solution is to change this constructor to a static constructor method

This is not too easy. The class this code is in is abstract, so instantiating it is not possible directly. I'll have to think about this for some more time to find the best solution.

vanitasvitae:
Updated the PR.

smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/element/OmemoBundleElement.java
```diff
@@ -59,10 +60,29 @@
      * @param preKeysB64 HashMap of base64 encoded preKeys
      */
     public OmemoBundleElement(int signedPreKeyId, String signedPreKeyB64, String signedPreKeySigB64, String identityKeyB64, HashMap<Integer, String> preKeysB64) {
+        if (signedPreKeyId <= 0) {
+            throw new IllegalArgumentException("signedPreKeyId MUST be greater than 0.");
+        }
         this.signedPreKeyId = signedPreKeyId;
+
+        if (StringUtils.isNullOrEmpty(signedPreKeyB64)) {
```
Flowdalic: There is `StringUtils.requireNotNullOrEmpty()` for this very use case. Please use that.

vanitasvitae: Done :)

PR #271, opened 2018-10-02
Fix example for Establishing a Connection

beedle-:


Flowdalic:
Thanks for your contribution. Wouldn't it be better to rename `conn1` to `connection` instead of vice versa?

beedle-:
There is a second example with conn2 so I choose conn1 rather than Connection for the first example.

PR #272, opened 2018-10-04
ChatMarkersManager improved - part 2

miguelhincapie:
* Filters modified to allow xep-0333 be used in MUC.
* Validation when trying to obtain Chat instance.
* Created a method to notify the recipients about new state.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
@@ -37,6 +37,7 @@ import org.jivesoftware.smack.filter.AndFilter;
 import org.jivesoftware.smack.filter.MessageTypeFilter;
 import org.jivesoftware.smack.filter.MessageWithBodiesFilter;
 import org.jivesoftware.smack.filter.NotFilter;
+import org.jivesoftware.smack.filter.OrFilter;
 import org.jivesoftware.smack.filter.PossibleFromTypeFilter;
 import org.jivesoftware.smack.filter.StanzaExtensionFilter;
 import org.jivesoftware.smack.filter.StanzaFilter;
@@ -73,14 +74,14 @@ public final class ChatMarkersManager extends Manager {
 
     // @FORMATTER:OFF
     private static final StanzaFilter INCOMING_MESSAGE_FILTER = new AndFilter(
-            MessageTypeFilter.NORMAL_OR_CHAT,
+            new OrFilter(MessageTypeFilter.NORMAL_OR_CHAT, MessageTypeFilter.GROUPCHAT),
             new StanzaExtensionFilter(ChatMarkersElements.NAMESPACE),
             PossibleFromTypeFilter.ENTITY_BARE_JID,
             EligibleForChatMarkerFilter.INSTANCE
     );
 
     private static final StanzaFilter OUTGOING_MESSAGE_FILTER = new AndFilter(
-            MessageTypeFilter.NORMAL_OR_CHAT,
+            new OrFilter(MessageTypeFilter.NORMAL_OR_CHAT, MessageTypeFilter.GROUPCHAT),
             MessageWithBodiesFilter.INSTANCE,
             new NotFilter(ChatMarkersFilter.INSTANCE),
             EligibleForChatMarkerFilter.INSTANCE
@@ -144,8 +145,13 @@ public final class ChatMarkersManager extends Manager {
                 // Note that this listener is used together with a PossibleFromTypeFilter.ENTITY_BARE_JID filter, hence
                 // every message is guaranteed to have a from address which is representable as bare JID.
                 EntityBareJid bareFrom = message.getFrom().asEntityBareJidOrThrow();
+                Chat tempChat = null;
+                try {
+                    tempChat = chatManager.chatWith(bareFrom);
+                } catch (Exception e) {
 
-                final Chat chat = chatManager.chatWith(bareFrom);
+                }
+                final Chat chat = tempChat;
 
                 asyncButOrdered.performAsyncButOrdered(chat, new Runnable() {
                     @Override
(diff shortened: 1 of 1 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
```diff
@@ -144,8 +145,13 @@ public void processStanza(Stanza packet)
                 // Note that this listener is used together with a PossibleFromTypeFilter.ENTITY_BARE_JID filter, hence
                 // every message is guaranteed to have a from address which is representable as bare JID.
                 EntityBareJid bareFrom = message.getFrom().asEntityBareJidOrThrow();
+                Chat tempChat = null;
+                try {
+                    tempChat = chatManager.chatWith(bareFrom);
+                } catch (Exception e) {
```
vanitasvitae: Please do not swallow exceptions silently.

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
```diff
@@ -218,4 +224,41 @@ public synchronized boolean removeIncomingChatMarkerMessageListener(ChatMarkersL
...
+     * @throws NotConnectedException if the connection is not connected.
+     * @throws InterruptedException  if the connection is interrupted.
+     * @throws IllegalStateException if one of the params don't match the rules.
+     */
+    public void markMessage(Message message, ChatMarkersState chatMarkersState, String messageId)
+            throws
+            NotConnectedException,
+            InterruptedException,
+            IllegalStateException {
+        if (message == null) {
```
vanitasvitae: Please use `Objects.requireNotNull()` for those nullity checks.

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
```diff
@@ -218,4 +224,41 @@ public synchronized boolean removeIncomingChatMarkerMessageListener(ChatMarkersL
...
+
+        if (message.getTo() == null) {
+            throw new IllegalStateException("To attribute must not be null");
+        }
+
+        if (message.getStanzaId() == null) {
+            message.setStanzaId();
+        }
+
+        if (chatMarkersState == ChatMarkersState.received) {
```
vanitasvitae: This might be a good place for a switch-case environment?

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
Flowdalic: `catch (Exception e)` is also bad practice and consider code smell unless it is one of the rare valid cases where you really want to catch all `Exception`s.

miguelhincapie: True is: I didn't went deeper in `chatWith()` method to search for all exceptions that can be thrown 😅, but the idea around it is because current code doesn't allow to be used in MUC (MUC use old `Chat` v1) so if I try to get a Chat (v2) giving a Bare from a room (group chat) it's not gonna work... in fact MUC doesn't needs the Chat.

So I can dig into that method to look for specific exceptions but I think (I'm not sure, you are the guys that know it better than me) the`empty try catch` is needed.

Tell me what to do and I will do it.

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
miguelhincapie: done

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
miguelhincapie: switched

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
Flowdalic: Simply do not catch any exception, there is no point in what the code actually does. It is actually harmful, because the `run()` method below will possibly be invoked with `chat == null`.

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
```diff
@@ -218,4 +225,45 @@ public synchronized boolean removeIncomingChatMarkerMessageListener(ChatMarkersL
...
+            NotConnectedException,
+            InterruptedException,
+            IllegalStateException {
+        Objects.requireNonNull(message, "Message must not be null");
+
+        if (message.getTo() == null) {
+            throw new IllegalStateException("To attribute must not be null");
+        }
+
+        if (message.getStanzaId() == null) {
```
Flowdalic: Please use `!message.hasStanzaIdSet()` instead

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
```diff
@@ -218,4 +225,45 @@ public synchronized boolean removeIncomingChatMarkerMessageListener(ChatMarkersL
...
+     * Send a message stanza to the recipient defined in the <tt>To</tt> getter from the <tt>Message</tt>.
+     *
+     * @param message          instance of {@link Message} with a To previous defined.
+     * @param chatMarkersState one of the values given in {@link ChatMarkersState}.
+     * @param messageId        id of the message to be updated.
+     * @throws NotConnectedException if the connection is not connected.
+     * @throws InterruptedException  if the connection is interrupted.
+     * @throws IllegalStateException if one of the params don't match the rules.
+     */
+    public void markMessage(Message message, ChatMarkersState chatMarkersState, String messageId)
```
Flowdalic: I would expect that the user gives as a message with a markable extension to such a method. Then you could remove the `ChatMarkersState` and `String` (`messageId`) parameters, which would make the method much more user friendly.

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
miguelhincapie: done

smack-experimental/src/main/java/org/jivesoftware/smackx/chat_markers/ChatMarkersManager.java
miguelhincapie: but the user would need to deal with `message.addExtension(new ChatMarkersElements.ReceivedExtension(messageId));` by himself, in other words he would doing some of the job that this class has to do.

What I want is class that can be used almost out of the box but of course knowing how the XEP works.

PR #274, opened 2018-10-10
Fix Roster.getPresence ConcurrentModificationException

spslinger:
Since Smack 4.2.4, the getPresencesInternal method in the Roster class can return a LruCache object, which is a LinkedHashMap with access order. This means that any access using get or getOrDefault will be a modification of the Map. If you loop over the keySet of the Map and there are more than one, the second call to get will throw a ConcurrentModificationException!

Since the keys are only used here to obtain the corresponding values, the simplest solution is to just loop over the values instead.

### DIFF
--- a/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
+++ b/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
@@ -951,8 +951,7 @@ public final class Roster extends Manager {
             // This is used in case no available presence is found
             Presence unavailable = null;
 
-            for (Resourcepart resource : userPresences.keySet()) {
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: d10924cd1bd343cbcbeef5c54df38def7bbee331

---

Flowdalic:
Thanks for your contribution. Much appreciated. I have created SMACK-841 to track this.

This looks like an ideal candidate for the 4.3 branch. Would you mind re-issuing this PR against the 4.3 branch. Also please mention the issue key SMACK-841 in the commit message.

spslinger:
Closing this PR, will re-issue for the 4.3 branch with changed commit message.

PR #280, opened 2018-11-07
Enables the HTTP compression in JBOSH

MarcelHeckel:
Add an extra parameter "httpCompressionEnabled" to BOSHConfiguration
that is used to set the setCompressionEnabled() of BOSHClientConfig

Related to issue SMACK-642

### DIFF
--- a/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
+++ b/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
@@ -157,6 +157,8 @@ public class XMPPBOSHConnection extends AbstractXMPPConnection {
             if (config.isProxyEnabled()) {
                 cfgBuilder.setProxy(config.getProxyAddress(), config.getProxyPort());
             }
+            cfgBuilder.setCompressionEnabled(config.isCompressionEnabled());
+
             client = BOSHClient.create(cfgBuilder.build());
 
             client.addBOSHClientConnListener(new BOSHConnectionListener());
--- a/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
@@ -125,6 +125,8 @@ public abstract class ConnectionConfiguration {
 
     private final Set<String> enabledSaslMechanisms;
 
+    private final boolean compressionEnabled;
+
     protected ConnectionConfiguration(Builder<?,?> builder) {
         authzid = builder.authzid;
         username = builder.username;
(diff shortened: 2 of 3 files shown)


Flowdalic:
Thanks for you contribution. :)

 I am not sure if we should introduce a new boolean field in `BOSHConfiguration`, instead simply move `compressionEnabled` in `XMPPTCPConnectionConfiguration` one layer up the `ConnectionConfiguration` hirarchy into `ConnectionConfiguration`.

Also there where multiple style issues with your commit. Next time, please simply run `gradle check` and see if it passes, before issuing the PR.

MarcelHeckel:
Hi,
there is this line 1477 in org.jivesoftware.smack.AbstractXMPPConnection.parseFeatures(XmlPullParser)

        // If the server reported the bind feature then we are that that we did SASL and maybe
        // STARTTLS. We can then report that the last 'stream:features' have been parsed
        if (hasFeature(Bind.ELEMENT, Bind.NAMESPACE)) {
            if (!hasFeature(Compress.Feature.ELEMENT, Compress.NAMESPACE)
                            || !config.isCompressionEnabled()) {
                // This was was last features from the server is either it did not contain
                // compression or if we disabled it
                lastFeaturesReceived.reportSuccess();
            }
        }

And I was not sure if this would damage anything if isCompressionEnabled() is true, although it was done on the HTTP level

Flowdalic:
> there is this line 1477

Good observation

> And I was not sure if this would damage anything

I don't think so, as BOSH connections won't announce the 'Compress` feature.

I think you can make the changes as requested, please squash your commits then and update your PR by force pushing (and don't forget to run `gradle check` first ;-) )

MarcelHeckel:
I have now overwritten setCompressionEnabled() (to be consistent with the TCP config) and squash commits.

Flowdalic:
> I have now overwritten setCompressionEnabled() (to be consistent with the TCP config) and squash commits.

Any reason why you did not move `compressionEnabled` in XMPPTCPConnectionConfiguration one layer up the ConnectionConfiguration hierarchy into ConnectionConfiguration and use this (instead of httpCompressionEnabled)? I'd like to avoid duplicate code.

MarcelHeckel:
As I said: to be consistent with the TCP config!

But if you really want this...

Flowdalic:
> As I said: to be consistent with the TCP config!

Just to be sure I am not missing something: What kind of consistency are we talking about and why is it desirable?

MarcelHeckel:
"consistent" is maybe the wrong word.
I just meant that "XMPPTCPConnectionConfiguration" was defining its own "compressionEnabled" that's why I did the same for "BOSHConfiguration". 

Now I moved the  "compressionEnabled" from the "XMPPTCPConnectionConfiguration" to the "ConnectionConfiguration" as you suggested.

PR #284, opened 2018-11-15
Fix previous archive page requested incorrectly.

tairrzayev-chi:
Unless I am mistaken, the previous archive page should be before the first message in the previous result set, not the last.


Flowdalic:
Thank you for your contribution. This indeed does look like a bug. I've created [SMACK-843](https://issues.igniterealtime.org/browse/SMACK-843) to track this. This also looks like a good candidate for the 4.3 branch.
Ideally this method would be part of the integration tests (or even better: the unit tests). In case you ever wanted to write a test for Smack. ;)

Flowdalic:
Cherry picked onto 4.3 branch as https://github.com/igniterealtime/Smack/commit/0332fa54d10ff357e40ebd979364ca4da6181fb8

bapihandzap:
After a fixed previous page issue now all messages in not receiving from mam query.

PR #289, opened 2018-12-04
WIP: Fix GSO-006

vanitasvitae:
This PR adds mitigation for GSO-006 by providing a method for postponing preKey deletion.

TODO:

* [ ] Add JUnit tests
* [ ] Document GSO-006 somewhere in the javadoc
* [ ] Optional: Make mitigation default behavior?


vanitasvitae:
It turns out, this is more complicated as I thought. I'll have to think about a proper fix.

PR #291, opened 2019-01-15
Fixing Javadoc duplicate with inheritDoc

oumaziz:
I'm a PhD Student from University of Bordeaux working on documentation copy-pastes (duplicates).
I've noticed that your project does have a few documentation duplicates.
For my research work, i've tried fix this duplicate and i'd love to have your feedback about it:

- Do you think that this duplicate documentation should co-evolve forever?
- Are you aware of copy-pastes in your documentation?
- If yes, how do you handle them?
- Were you aware of JavaDoc reuse features such as {@inheritdoc}?
- Would you like to have a tool that helps you maintain duplicate documentation?

### DIFF
--- a/smack-jingle-old/src/main/java/org/jivesoftware/smackx/jingleold/mediaimpl/jmf/AudioMediaSession.java
+++ b/smack-jingle-old/src/main/java/org/jivesoftware/smackx/jingleold/mediaimpl/jmf/AudioMediaSession.java
@@ -119,10 +119,7 @@ public class AudioMediaSession extends JingleMediaSession {
     }
 
     /**
-     * Set transmit activity. If the active is true, the instance should transmit.
-     * If it is set to false, the instance should pause transmit.
-     *
-     * @param active active state
+     * {@inheritDoc}
      */
     @Override
     public void setTransmit(boolean active) {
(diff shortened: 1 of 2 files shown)


vanitasvitae:
Hi!

Thanks for the PR!

I can only speak for myself here.

> Do you think that this duplicate documentation should co-evolve forever?

The code snippets you fixed are currently not maintained, so they do have a lowish priority.

> Were you aware of JavaDoc reuse features such as {@inheritdoc}?

Yes, its used in some other places throughout the project.

> Are you aware of copy-pastes in your documentation?

There are situations where using duplicate javadoc may be a fine solution. A person that looks at the actual code might want to see the actual description if a method instead of just {@inheritdoc}.

Flowdalic:
Using just {@inheritDoc} is usually always wrong. In this case it would have been better to just delete the duplicated javadoc. You only want to use {@inerhitDoc} if the javadoc block add extra additional descriptive text.

oumaziz:
@vanitasvitae @Flowdalic Thank you both for your feedback! 
@Flowdalic regarding {@inheritdoc}, indeed, just not adding anything would perfectly work. I used {@inheritdoc} just to showcase the situation.

PR #279, opened 2018-11-06
PingManager improved to avoid unnecessary pings

MarcelHeckel:
Moved the "getLastStanzaReceived()" check inside the retry loop of the PingManager.pingServerIfNecessary() method. So unnecessary pings are avoided.

PR #296, opened 2019-03-13
PingManager: con. assumed ok for any packed within the ping interval

MarcelHeckel:
We assume that the connection is good even if there was no ping result but a stanzas within the ping interval.

New merge request because squashing was not possible due to conflicts. See original
See #279


Flowdalic:
Thanks. There are two minor issues with your commit title.

- Do not abbreviate ("con."
- Packet is not correctly spelled and the XMPP term "stanza" should be used instead.

I suggest
```
PingManager: Assume server ping successful if stanzas where received
```
as commit title.

Simply modify the commit in your `PingManager_ConnectionOkOnAnyPacket` branch:

```bash
git checkout PingManager_ConnectionOkOnAnyPacket
git commit --amend
git push --force <remote>
```

MarcelHeckel:
OK, done.

PR #297, opened 2019-03-14
Adding public final ServiceDiscoveryManager instance to MultiUserChatManager

adiaholic:
The expensive method ServiceDiscoveryManager.getInstanceFor() is replaced by a final private field

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
@@ -82,7 +82,7 @@ public final class MultiUserChatManager extends Manager {
     private static final String DISCO_NODE = MUCInitialPresence.NAMESPACE + "#rooms";
 
     private static final Logger LOGGER = Logger.getLogger(MultiUserChatManager.class.getName());
-    private final ServiceDiscoveryManager serviceDiscoveryManager;
+
     static {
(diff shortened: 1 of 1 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -82,7 +82,7 @@
     private static final String DISCO_NODE = MUCInitialPresence.NAMESPACE + "#rooms";
 
     private static final Logger LOGGER = Logger.getLogger(MultiUserChatManager.class.getName());
-
+    private final ServiceDiscoveryManager serviceDiscoveryManager;
```
Flowdalic: The declaration should be not with the block auf static field declarations, but further down where the non-static fields are declared (e.g. just above the first constructor). Also there should be a newline before the next non-empty line comes.

vanitasvitae:
Plus the Jive issue number ;)

Flowdalic:
> Plus the Jive issue number ;)

Ahh good point. Assuming you mean Jira, not Jive. But anyway, please mentioned the Smack issue number that is related to the commit somewhere in the commit body so that Jira automatically links the commit.

Flowdalic:
Squash into a single commit please. I am also not happy about the commit title. It really shouldn't continue in the commit message body if possible.

PR #298, opened 2019-03-15
Synchronize handling of received BOSH events

cl90:
Added synchronized modifier to responseReceived function in BOSHPacketReader to keep order of incoming events as required especially by SyncStanzaListeners.

### DIFF
--- a/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
+++ b/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java
@@ -458,7 +458,7 @@ public class XMPPBOSHConnection extends AbstractXMPPConnection {
          * @param event the BOSH client response which includes the received packet.
          */
         @Override
-        public void responseReceived(BOSHMessageEvent event) {
+        public synchronized void responseReceived(BOSHMessageEvent event) {
             AbstractBody body = event.getBody();
             if (body != null) {
                 try {


GitCop:

 * Commit: c4c99d2f4e87eecf93c24fa33f885bd14c2cad73

---

Flowdalic:
Thanks for your contribution.

While it would not surprise me if jBOSH/smack-bosh had issues keeping the order of top level stream elements, I fail to see how this fixes anything: Adding the `synchronized` keyword to the method would not impose an order if the method is called in parallel. Maybe I am mistaken and you can enlighten me how this helps.

cl90:
Let us assume that this function is called for two responses (event1 with a large body consisting of many stanzas and event2 consisting of a few) by parallel RequestProcessors in the correct order, but within a short period of time. Parsing and processing of event1 (including passing all its stanzas to the listeners) in a non-synchronised function might take too long to be finished before processing of event2 starts. This implies that the first stanza(s) of event2 will be processed before the last stanza(s) of event1 although event1 has clearly arrived earlier.

Using the synchronised keyword fixes exactly this: Parsing and processing of event2 is not started before the last stanza of event1 has been processed.

This still requires that the calls of this function by the RequestProcessors of BOSHClient occur in the correct order of events.

For instance, the erroneous behaviour might occur when processing large responses of a message archive query (as described in XEP-0313) performing at least two parallel BOSH requests: The client receives multiple bodies (possibly varying in size and stanza count) within a short period of time. However, the result IQ must be processed after last message in the result set, regardless which body these stanzas arrive in.

Flowdalic:
What if the thread processing event1 is delayed by scheduling before it reaches the synchronized method and meanwhile the thread processing event2 reaches the synchronized method?

cl90:
This change assumes that the synchronized method is called in the correct order of events by BOSHClient. Since this is apperently not guarenteed, the stanza order might still be changed (although it is less likely).

Flowdalic:
I would rather see a real fix instead of poor mans band-aid. Hence I am closing this.

Smack's BOSH code and jbosh are in a bad shape and in need for some care. Unfortunately to this date no one showed interest in funding the required overhaul.

PR #299, opened 2019-03-15
Autojoin callback on reconnect

vincentwlau:
Upon a successful rejoining the MUC's after reconnection, it will be useful to have a callback that the client can perform some post operations.  For example, a non-persistent MUC room may be recreated after the rejoin, and the room needs to be unlocked for other participants to join later.

This pull request is to add a new callback interface and a mechanism to register the callback.  It is similar to the failure callback.


vincentwlau:
@Flowdalic I squashed the two commits as you requested, then rebased to the latest master and resolved a conflict.  This is my first time to do a Pull Request with squash.  If I am not doing it correctly, please let me know.

Flowdalic:
> If you want to me to submit another PR with a cleaner branch, I'll do it.

No need to create another PR. You can simply update this PR by force pushing to its source branch.

Again, please squash so that there is only a single commit with the desired changes in the PR.

vincentwlau:
I believe that I have squashed and rebased up to 03/24.  If I still didn't do it correctly, please let me know.

PR #307, opened 2019-04-01
Fixed NPE in Roster.getPresencesInternal(BareJid)

MarcelHeckel:
Key used in getPresencesInternal can be null.
(see final BareJid key = from != null ? from.asBareJid() : null;)
But ConcurrentHashMap does not allow null keys.
So handle null key separately.

### DIFF
--- a/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
+++ b/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
@@ -371,7 +371,7 @@ public final class Roster extends Manager {
      * @return the user presences
      */
     private Map<Resourcepart, Presence> getPresencesInternal(BareJid entity) {
-        Map<Resourcepart, Presence> entityPresences = presenceMap.get(entity);
+        Map<Resourcepart, Presence> entityPresences = entity == null ? null : presenceMap.get(entity);
         if (entityPresences == null) {
             entityPresences = nonRosterPresenceMap.lookup(entity);
         }


Flowdalic:
Thanks for your contribution. After a quick look this appears to be a good candidate for the 4.3 branch. I am curious: Which stanza did you encounter that triggered the NPE?

MarcelHeckel:
I did not investigated the Stanza. But I got an several app crashes with the following stack trace


```
 FATAL EXCEPTION: Smack Cached Executor
    Process: de.fhg.ivi.senetz.mobile.android.mbk.debug, PID: 13365
    java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.Object.hashCode()' on a null object reference
        at java.util.concurrent.ConcurrentHashMap.get(ConcurrentHashMap.java:944)
        at org.jivesoftware.smack.roster.Roster.getPresencesInternal(Roster.java:374)
        at org.jivesoftware.smack.roster.Roster.getOrCreatePresencesInternal(Roster.java:388)
        at org.jivesoftware.smack.roster.Roster.access$1100(Roster.java:94)
        at org.jivesoftware.smack.roster.Roster$PresencePacketListener$1.run(Roster.java:1519)
        at org.jivesoftware.smack.AsyncButOrdered$Handler.run(AsyncButOrdered.java:121)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
        at java.lang.Thread.run(Thread.java:764)
```

Just wanted to avoid app crashes, when a such a strange presence (with no from) is received.

Flowdalic:
Created [SMACK-861](https://issues.igniterealtime.org/browse/SMACK-861).

Flowdalic:
Here is my shot to fix this: https://github.com/Flowdalic/Smack/commit/96a5ad5775f0213858c17c756d5c51a7ebb61e43

MarcelHeckel:
Yes it is better to avoid the null at all.

But regarding the code comment: I can't see a "8.1.2.1 4." in RFC 6121
https://xmpp.org/rfcs/rfc6121.html
Wrong Number?

Flowdalic:
Good catch: https://github.com/Flowdalic/Smack/commit/e01debdf585200a604e50aee251fdb50250b361f

MarcelHeckel:
👍

yudhisthira123:
in which version this fix is available? I am using 4.3.0 and getting this crash..

PR #308, opened 2019-04-04
Prevent extremely long reply timeouts from being set.

adiaholic:
Smack will throw an IllegalArguementException if extremely long reply
timeouts are tried to be set.I assumed currentTimeMillis() was the
boundary as per SMACK-718.


vanitasvitae:
I'm pretty sure comparing the value to `currentTimeMillis()` is not what @Flowdalic meant in the ticket.
You probably want to check, if `currentTimeMillis() + timeout` would be greater than `Long.MAX_VALUE`.

Hint: you have to reorder the equation in order to prevent overflows ;)

PR #316, opened 2019-05-01
Add support for XEP-0118: UserTune

adiaholic:
This commit will enable user to communicate
information about music to which he/she is listening.
This feature is less of a requirement and more like fun to me.
An attempt at solving SMACK-257.
This is my first attempt at implementing an XEP,
so i referred to previous XEP implementations.
I would love to have suggestions from everyone :)

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
@@ -1,0 +1,143 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune;
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.AsyncButOrdered;
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.SmackException.NoResponseException;
+import org.jivesoftware.smack.SmackException.NotConnectedException;
+import org.jivesoftware.smack.SmackException.NotLoggedInException;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPException.XMPPErrorException;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.pep.PepListener;
+import org.jivesoftware.smackx.pep.PepManager;
+import org.jivesoftware.smackx.pubsub.EventElement;
+import org.jivesoftware.smackx.pubsub.ItemsExtension;
+import org.jivesoftware.smackx.pubsub.PayloadItem;
+import org.jivesoftware.smackx.pubsub.PubSubException.NotALeafNodeException;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+
+import org.jxmpp.jid.BareJid;
+import org.jxmpp.jid.EntityBareJid;
+
+/**
+ * Entry point for Smacks API for XEP-0118: User Tune.
+ * <br>
+ * To publish a UserTune, please use {@link #publishUserTune(UserTuneElement)} method. This will publish the node.
+ * <br>
+ * To stop publishing a UserTune, please use {@link #clearUserTune()} method. This will send a disabling publish signal.
+ * <br>
+ * To add a UserTune listener in order to remain updated with other users UserTune, use {@link #addUserTuneListener(UserTuneListener)} method.
+ * <br>
+ * To link a UserTuneElement with {@link Message}, use 'message.addExtension(userTuneElement)'.
+ * <br>
+ * An example to illustrate is provided inside UserTuneElementTest inside the test package.
+ * <br>
+ * @see <a href="https://xmpp.org/extensions/xep-0118.html">
+ *     XEP-0118: User Tune</a>
+ */
+public final class UserTuneManager extends Manager {
+
+    public static final String USERTUNE_NODE = "http://jabber.org/protocol/tune";
+    public static final String USERTUNE_NOTIFY = USERTUNE_NODE + "+notify";
+
+    private static final Map<XMPPConnection, UserTuneManager> INSTANCES = new WeakHashMap<>();
+
+    private final Set<UserTuneListener> userTuneListeners = new HashSet<>();
+    private final AsyncButOrdered<BareJid> asyncButOrdered = new AsyncButOrdered<>();
+    private final PepManager pepManager;
+    private boolean ENABLE_USER_TUNE_NOTIFICATIONS_BY_DEFAULT = true;
+
+    public static synchronized UserTuneManager getInstanceFor(XMPPConnection connection) throws NotLoggedInException {
+        UserTuneManager manager = INSTANCES.get(connection);
+        if (manager == null) {
+            manager = new UserTuneManager(connection);
+            INSTANCES.put(connection, manager);
+        }
+        return manager;
+    }
+
+    private UserTuneManager(XMPPConnection connection) throws NotLoggedInException {
+        super(connection);
+        pepManager = PepManager.getInstanceFor(connection);
+        pepManager.addPepListener(new PepListener() {
+            @Override
+            public void eventReceived(EntityBareJid from, EventElement event, Message message) {
+                if (!USERTUNE_NODE.equals(event.getEvent().getNode())) {
+                    return;
+                }
+
+                final BareJid contact = from.asBareJid();
+                asyncButOrdered.performAsyncButOrdered(contact, () -> {
+                    ItemsExtension items = (ItemsExtension) event.getExtensions().get(0);
+                    PayloadItem<?> payload = (PayloadItem) items.getItems().get(0);
+                    UserTuneElement tune = (UserTuneElement) payload.getPayload();
+
+                    for (UserTuneListener listener : userTuneListeners) {
+                        synchronized (userTuneListeners) {
+                            listener.onUserTuneUpdated(contact, message, tune);
+                        }
+                    }
+                });
+            }
+        });
+        if (ENABLE_USER_TUNE_NOTIFICATIONS_BY_DEFAULT) {
+            enableUserTuneNotifications();
+        }
+    }
+
+    public void setUserTuneNotificationsEnabledByDefault(boolean bool) {
+        ENABLE_USER_TUNE_NOTIFICATIONS_BY_DEFAULT = bool;
+    }
+
+    public void enableUserTuneNotifications() {
+        ServiceDiscoveryManager.getInstanceFor(connection()).addFeature(USERTUNE_NOTIFY);
+    }
+
+    @SuppressWarnings("unused")
+    private void disableUserTuneNotifications() {
+        ServiceDiscoveryManager.getInstanceFor(connection()).removeFeature(USERTUNE_NOTIFY);
+    }
+
+    public void clearUserTune() throws NotLoggedInException, NotALeafNodeException, NoResponseException, NotConnectedException, XMPPErrorException, InterruptedException {
+        publishUserTune(UserTuneElement.EMPTY_USER_TUNE);
+    }
+
+    public void publishUserTune(UserTuneElement userTuneElement) throws NotLoggedInException, NotALeafNodeException, NoResponseException, NotConnectedException, XMPPErrorException, InterruptedException {
+        // TODO: To prevent a large number of updates when a user is skipping through tracks, an implementation SHOULD wait several seconds before publishing new tune information.
+        pepManager.publish(USERTUNE_NODE, new PayloadItem<>(userTuneElement));
+    }
+
+    public void addUserTuneListener(UserTuneListener listener) {
+        synchronized (userTuneListeners) {
+            userTuneListeners.add(listener);
+        }
+    }
+
+    public void removeUserTuneListener(UserTuneListener listener) {
+        synchronized (userTuneListeners) {
+            userTuneListeners.remove(listener);
+        }
+    }
+}
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
@@ -1,0 +1,267 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune.element;
+
+import java.net.URI;
+
+import org.jivesoftware.smack.datatypes.UInt16;
+import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.packet.XmlEnvironment;
+import org.jivesoftware.smack.util.EqualsUtil;
+import org.jivesoftware.smack.util.HashCode;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+
+/**
+ * {@link ExtensionElement} that contains the UserTune. <br>
+ * Instance of UserTuneElement can be created using {@link Builder#build()}
+ * method.
+ */
+public final class UserTuneElement implements ExtensionElement {
+
+    public static final String NAMESPACE = "http://jabber.org/protocol/tune";
+    public static final String ELEMENT = "tune";
+
+    private final String artist;
+    private final UInt16 length;
+    private final Integer rating;
+    private final String source;
+    private final String title;
+    private final String track;
+    private final URI uri;
+
+    public static final UserTuneElement EMPTY_USER_TUNE = null;
+
+    private UserTuneElement(Builder builder) {
+        this.artist = builder.artist;
+        this.length = builder.length;
+        this.rating = builder.rating;
+        this.source = builder.source;
+        this.title = builder.title;
+        this.track = builder.track;
+        this.uri = builder.uri;
+    }
+
+    @Override
+    public String getNamespace() {
+        return NAMESPACE;
+    }
+
+    @Override
+    public String getElementName() {
+        return ELEMENT;
+    }
+
+    public String getArtist() {
+        return artist;
+    }
+
+    public UInt16 getLength() {
+        return length;
+    }
+
+    public Integer getRating() {
+        return rating;
+    }
+
+    public String getSource() {
+        return source;
+    }
+
+    public String getTitle() {
+        return title;
+    }
+
+    public String getTrack() {
+        return track;
+    }
+
+    public URI getUri() {
+        return uri;
+    }
+
+    @Override
+    public XmlStringBuilder toXML(XmlEnvironment xmlEnvironment) {
+        XmlStringBuilder xml = new XmlStringBuilder(this);
+        if (isEmptyUserTune()) {
+            return xml.closeEmptyElement();
+        }
+        xml.rightAngleBracket();
+        xml.optElement("artist", artist);
+        xml.optElement("length", length);
+        xml.optElement("rating", rating);
+        xml.optElement("source", source);
+        xml.optElement("title", title);
+        xml.optElement("track", track);
+        xml.optElement("uri", uri);
+        return xml.closeElement(getElementName());
+    }
+
+    private boolean isEmptyUserTune() {
+        return this.equals(EMPTY_USER_TUNE);
+    }
+
+    public static boolean hasUserTuneElement(Message message) {
+        return message.hasExtension(ELEMENT, NAMESPACE);
+    }
+
+    public static UserTuneElement from(Message message) {
+        return message.getExtension(ELEMENT, NAMESPACE);
+    }
+
+    @Override
+    public int hashCode() {
+        return HashCode.builder()
+                .append(artist)
+                .append(length)
+                .append(rating)
+                .append(source)
+                .append(title)
+                .append(track)
+                .append(uri).build();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return EqualsUtil
+                .equals(this, obj, (equalsBuilder, otherTune) -> equalsBuilder
+                                                            .append(artist, otherTune.artist)
+                                                            .append(length, otherTune.length)
+                                                            .append(rating, otherTune.rating)
+                                                            .append(source, otherTune.source)
+                                                            .append(title, otherTune.source)
+                                                            .append(title, otherTune.title)
+                                                            .append(track, otherTune.track)
+                                                            .append(uri, otherTune.uri));
+    }
+
+    /**
+     * Returns a new instance of {@link Builder}.
+     * @return Builder
+     */
+    public static Builder getBuilder() {
+        return new Builder();
+    }
+
+    /**
+     * This class defines a Builder class for {@link UserTuneElement}. <br>
+     * {@link UserTuneElement} instance can be obtained using the {@link #build()} method as follows. <br>
+     * UserTuneElement.Builder builder = new UserTuneElement.Builder();
+     * builder.setSource("Yessongs"); builder.setTitle("Heart of the Sunrise");
+     * UserTuneElement userTuneElement = builder.build(); <br>
+     * Values such as title, source, artist, length, source, track and uri can be set using their respective setters through {@link Builder}
+     */
+    public static final class Builder {
+        private String artist;
+        private UInt16 length;
+        private Integer rating;
+        private String source;
+        private String title;
+        private String track;
+        private URI uri;
+
+        private Builder() {
+        }
+
+        /**
+         * Artist is an optional element in UserTuneElement.
+         * @param artist.
+         * @return builder.
+         */
+        public Builder setArtist(String artist) {
+            this.artist = artist;
+            return this;
+        }
+
+        /**
+         * Length is an optional element in UserTuneElement.
+         * @param length.
+         * @return builder.
+         */
+        public Builder setLength(int length) {
+            return setLength(UInt16.from(length));
+        }
+
+        /**
+         * Length is an optional element in UserTuneElement.
+         * @param length.
+         * @return builder.
+         */
+        public Builder setLength(UInt16 length) {
+            this.length = length;
+            return this;
+        }
+
+        /**
+         * Rating is an optional element in UserTuneElement.
+         * @param rating.
+         * @return builder.
+         */
+        public Builder setRating(int rating) {
+            this.rating = rating;
+            return this;
+        }
+
+        /**
+         * Source is an optional element in UserTuneElement.
+         * @param source.
+         * @return builder.
+         */
+        public Builder setSource(String source) {
+            this.source = source;
+            return this;
+        }
+
+        /**
+         * Title is an optional element in UserTuneElement.
+         * @param title.
+         * @return builder.
+         */
+        public Builder setTitle(String title) {
+            this.title = title;
+            return this;
+        }
+
+        /**
+         * Track is an optional element in UserTuneElement.
+         * @param track.
+         * @return builder.
+         */
+        public Builder setTrack(String track) {
+            this.track = track;
+            return this;
+        }
+
+        /**
+         * URI is an optional element in UserTuneElement.
+         * @param uri.
+         * @return builder.
+         */
+        public Builder setUri(URI uri) {
+            this.uri = uri;
+            return this;
+        }
+
+        /**
+         * This method is called to build a UserTuneElement.
+         * @return UserTuneElement.
+         */
+        public UserTuneElement build() {
+            return new UserTuneElement(this);
+        }
+    }
+}
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
@@ -1,0 +1,89 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune.provider;
+
+import java.io.IOException;
+import java.net.URI;
+
+import org.jivesoftware.smack.packet.XmlEnvironment;
+import org.jivesoftware.smack.parsing.SmackParsingException;
+import org.jivesoftware.smack.provider.ExtensionElementProvider;
+import org.jivesoftware.smack.util.ParserUtils;
+import org.jivesoftware.smack.xml.XmlPullParser;
+import org.jivesoftware.smack.xml.XmlPullParserException;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+
+/**
+ * This is the Provider Class for {@link UserTuneElement}.
+ */
+public class UserTuneProvider extends ExtensionElementProvider<UserTuneElement> {
+
+    public static final UserTuneProvider INSTANCE = new UserTuneProvider();
+
+    @Override
+    public UserTuneElement parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment)
+            throws XmlPullParserException, IOException, SmackParsingException {
+
+        UserTuneElement.Builder builder = UserTuneElement.getBuilder();
+        XmlPullParser.TagEvent tag = parser.nextTag();
+        outerloop: while (true) {
+            switch (tag) {
+            case START_ELEMENT:
+                String name = parser.getName();
+                String namespace = parser.getNamespace();
+                if (!UserTuneElement.NAMESPACE.equals(namespace)) {
+                    continue outerloop;
+                }
+                while (tag == XmlPullParser.TagEvent.START_ELEMENT) {
+                    switch (name) {
+                    case "artist":
+                        builder.setArtist(parser.nextText());
+                        break;
+                    case "length":
+                        builder.setLength(ParserUtils.getIntegerFromNextText(parser));
+                        break;
+                    case "rating":
+                        builder.setRating(ParserUtils.getIntegerFromNextText(parser));
+                        break;
+                    case "source":
+                        builder.setSource(parser.nextText());
+                        break;
+                    case "title":
+                        builder.setTitle(parser.nextText());
+                        break;
+                    case "track":
+                        builder.setTrack(parser.nextText());
+                        break;
+                    case "uri":
+                        URI uri = ParserUtils.getUriFromNextText(parser);
+                        builder.setUri(uri);
+                        break;
+                    }
+                    tag = parser.nextTag();
+                    name = parser.getName();
+                }
+                break;
+            case END_ELEMENT:
+                if (parser.getDepth() == initialDepth) {
+                    break outerloop;
+                }
+                break;
+            }
+        }
+        return builder.build();
+    }
+}
--- a/smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers
+++ b/smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers
@@ -589,4 +589,11 @@
           <className>org.jivesoftware.smackx.mood.provider.MoodProvider</className>
         </extensionProvider>
 
+    <!-- XEP-0118: User Tune -->
+    <extensionProvider>
+        <elementName>tune</elementName>
+        <namespace>http://jabber.org/protocol/tune</namespace>
+        <className>org.jivesoftware.smackx.usertune.provider.UserTuneProvider</className>
+    </extensionProvider>
+    
 </smackProviders>
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
@@ -1,0 +1,74 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune;
+
+import static org.jivesoftware.smack.test.util.XmlUnitUtils.assertXmlSimilar;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.jivesoftware.smack.parsing.SmackParsingException;
+import org.jivesoftware.smack.test.util.SmackTestSuite;
+import org.jivesoftware.smack.test.util.SmackTestUtil;
+import org.jivesoftware.smack.test.util.SmackTestUtil.XmlPullParserKind;
+import org.jivesoftware.smack.xml.XmlPullParser;
+import org.jivesoftware.smack.xml.XmlPullParserException;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+import org.jivesoftware.smackx.usertune.provider.UserTuneProvider;
+
+import org.junit.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+public class UserTuneElementTest extends SmackTestSuite {
+
+    private final String xml = "<tune xmlns='http://jabber.org/protocol/tune'>" +
+            "<artist>Yes</artist>" +
+            "<length>686</length>" +
+            "<rating>8</rating>" +
+            "<source>Yessongs</source>" +
+            "<title>Heart of the Sunrise</title>" +
+            "<track>3</track>" +
+            "<uri>http://www.yesworld.com/lyrics/Fragile.html#9</uri>" +
+            "</tune>";
+
+    @Test
+    public void toXmlTest() throws IOException, XmlPullParserException, SmackParsingException, URISyntaxException {
+
+        URI uri = new URI("http://www.yesworld.com/lyrics/Fragile.html#9");
+
+        UserTuneElement.Builder builder = UserTuneElement.getBuilder();
+        UserTuneElement userTuneElement = builder.setArtist("Yes")
+                .setLength(686)
+                .setRating(8)
+                .setSource("Yessongs")
+                .setTitle("Heart of the Sunrise")
+                .setTrack("3")
+                .setUri(uri)
+                .build();
+        assertXmlSimilar(xml, userTuneElement.toXML().toString());
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = SmackTestUtil.XmlPullParserKind.class)
+    public void userTuneElementProviderTest(XmlPullParserKind parserKind) throws XmlPullParserException, IOException, SmackParsingException {
+        XmlPullParser parser = SmackTestUtil.getParserFor(xml, parserKind);
+        UserTuneElement parsed = UserTuneProvider.INSTANCE.parse(parser);
+        assertXmlSimilar(xml, parsed.toXML().toString());
+    }
+}
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneManagerTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneManagerTest.java
@@ -1,0 +1,58 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune;
+
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.assertNotNull;
+import static junit.framework.TestCase.assertTrue;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.test.util.SmackTestSuite;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+
+import org.junit.Test;
+
+public class UserTuneManagerTest extends SmackTestSuite{
+
+    @Test
+    public void addMessage() throws URISyntaxException {
+
+        UserTuneElement.Builder builder = UserTuneElement.getBuilder();
+        builder.setArtist("Yes");
+        builder.setLength(686);
+        builder.setRating(8);
+        builder.setSource("Yessongs");
+        builder.setTitle("Heart of the Sunrise");
+        builder.setTrack("3");
+        URI uri = new URI("http://www.yesworld.com/lyrics/Fragile.html#9");
+        builder.setUri(uri);
+        UserTuneElement userTuneElement = builder.build();
+
+        Message message = new Message();
+        message.addExtension(userTuneElement);
+
+        assertTrue(message.hasExtension(UserTuneElement.ELEMENT, UserTuneElement.NAMESPACE));
+        assertTrue(UserTuneElement.hasUserTuneElement(message));
+
+        UserTuneElement element = UserTuneElement.from(message);
+        assertNotNull(element);
+        assertEquals(userTuneElement, element);
+    }
+}
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/usertune/UserTuneIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/usertune/UserTuneIntegrationTest.java
@@ -1,0 +1,89 @@
+/**
+ *
+ * Copyright 2019 Aditya Borikar.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.usertune;
+
+import java.net.URI;
+
+import org.jivesoftware.smack.SmackException;
+import org.jivesoftware.smack.SmackException.NotLoggedInException;
+import org.jivesoftware.smack.XMPPException;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+
+import org.igniterealtime.smack.inttest.AbstractSmackIntegrationTest;
+import org.igniterealtime.smack.inttest.SmackIntegrationTest;
+import org.igniterealtime.smack.inttest.SmackIntegrationTestEnvironment;
+import org.igniterealtime.smack.inttest.util.IntegrationTestRosterUtil;
+import org.igniterealtime.smack.inttest.util.SimpleResultSyncPoint;
+import org.junit.AfterClass;
+
+import org.jxmpp.jid.BareJid;
+
+public class UserTuneIntegrationTest extends AbstractSmackIntegrationTest {
+
+    private final UserTuneManager utm1;
+    private final UserTuneManager utm2;
+
+    public UserTuneIntegrationTest(SmackIntegrationTestEnvironment<?> environment) throws NotLoggedInException {
+        super(environment);
+        utm1 = UserTuneManager.getInstanceFor(conOne);
+        utm2 = UserTuneManager.getInstanceFor(conTwo);
+    }
+
+    @SmackIntegrationTest
+    public void test() throws Exception {
+        URI uri = new URI("http://www.yesworld.com/lyrics/Fragile.html#9");
+        UserTuneElement.Builder builder = UserTuneElement.getBuilder();
+        UserTuneElement userTuneElement1 = builder.setArtist("Yes")
+                                                  .setLength(686)
+                                                  .setRating(8)
+                                                  .setSource("Yessongs")
+                                                  .setTitle("Heart of the Sunrise")
+                                                  .setTrack("3")
+                                                  .setUri(uri)
+                                                  .build();
+
+        IntegrationTestRosterUtil.ensureBothAccountsAreSubscribedToEachOther(conOne, conTwo, timeout);
+
+        final SimpleResultSyncPoint userTuneReceived = new SimpleResultSyncPoint();
+
+        final UserTuneListener userTuneListener = new UserTuneListener() {
+            @Override
+            public void onUserTuneUpdated(BareJid jid, Message message, UserTuneElement userTuneElement) {
+                if (userTuneElement.equals(userTuneElement1)) {
+                    userTuneReceived.signal();
+                }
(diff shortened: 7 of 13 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Child elements described as in XEP-0118: User Tune.
+ * @see <a href="https://xmpp.org/extensions/xep-0118.html#table-1"> </a>
+ *
+ * TODO: If the user is listening to a stream, length should not be included.
+ */
+public class UserTune {
+
+    String artist;
```
vanitasvitae: Members of `ExtensionElements` should be private and final.

For complex elements with many members like this, where some members are optional, while others are not, it is advisable to create a `Builder` to construct them. The Builder can be used to require non-optional members.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
```diff
@@ -0,0 +1,82 @@
...
+    int rating;
+    String source;
+    String title;
+    int track;
+    String uri;
+
+    public String getArtist() {
+        return artist;
+    }
+    public void setArtist(String artist) {
```
vanitasvitae: If you have final members and make use of the Builder pattern, you dont need setters in the class any more.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
```diff
@@ -0,0 +1,82 @@
...
+    public void setTitle(String title) {
+        this.title = title;
+    }
+    public int getTrack() {
+        return track;
+    }
+    public void setTrack(int track) {
+        this.track = track;
+    }
+    public String getUri() {
```
vanitasvitae: I haven't read the XEP in a while, but wouldn't it make sense to use a `URI` here?

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private static final Map<XMPPConnection, UserTuneManager> INSTANCES = new WeakHashMap<>();
+
+    private final Set<UserTuneListener> userTuneListeners = new HashSet<>();
+    private final AsyncButOrdered<BareJid> asyncButOrdered = new AsyncButOrdered<>();
+    private PubSubManager pubSubManager;
+
+    private UserTuneManager(XMPPConnection connection) {
+        super(connection);
+        ServiceDiscoveryManager.getInstanceFor(connection).addFeature(USERTUNE_NOTIFY);
```
vanitasvitae: I'd swap this line and the line below. Even if it is super unlikely, that way you wont lose any incoming `UserTune` events that are received before you start listening for them.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,167 @@
...
+    private static UserTuneElement buildUserTune(UserTune userTune,UserTuneConcretisation concretisation,String text) {
+        return new UserTuneElement(
+                new UserTuneElement.UserTuneSubjectElement(userTune,concretisation));
+    }
+
+    public static void addUserTuneToMessage(Message message,UserTune userTune) {
+        addUserTuneToMessage(message, userTune, null);
+    }
+
+    public static void addUserTuneToMessage(Message message,UserTune userTune,UserTuneConcretisation concretisation) {
```
vanitasvitae: You might want to move methods like `addUserTuneToMessage()` into the UserTuneElement class and reference those methods from here. That way you can also create a method like `getUserTuneElementFromMessage` inside the `UserTuneElement` class and have both utility methods for adding and retrieving elements from a message in one place.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
vanitasvitae: Ah wait, this is not an Element :D Its too early for me :)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
vanitasvitae: I see you already did this further below :+1:

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,138 @@
...
+        return userTune != null ? userTune.getConcretisation() : null;
+    }
+
+    public boolean hasConcretisation() {
+        return getUserTuneConcretisation() != null;
+    }
+
+    @Override
+    public String getNamespace() {
+        // TODO Auto-generated method stub
```
vanitasvitae: Some remnants of you IDE ;)

adiaholic:
> My first impression is, that this code is quite solid. There are some things to change though.
> 
> First of all: I feel like you heavily oriented yourself on my implementation of XEP-0107: User Mood. This is totally fine of course, the XEPs are pretty similar and it would be a bad idea to reinvent the wheel. However, there are some differences between the XEPs, so there are some shortcuts you would have taken when implementing everything yourselves instead of transforming the mood code.
> 
> For example: User Tune doesn't have such a thing as UserTuneConcretisations. Same goes for UserMoodSubject. While it might be a good idea to have such features in the XEP, it does not belong into the implementation, as it makes your implementation incompatible with others. In the case of UserMood, the XEP states, that a Mood Concretisation and Subject exist.
> 
> You could remove the distinction between the UserTune class and the UserTuneElement class. As I see it, the UserTune class adds no value to the implementation and a UserTuneElement class should be sufficient for your needs. If you remove the UserTuneConcretisation stuff and the UserTuneSubject classes, you can save some lines. I'd then add a UserTuneBuilder, which can be used to build a UserTuneElement. The builder can be used inside your UserTuneProvider where it would make sure, that the incoming UserTuneElement is compliant to the XEP (i.e. throw an exception if a mandatory attribute like the titke is missing, or if attributes are being received twice etc). Additionally the builder can be used by the user to create a UserTuneElement for the UserTuneManager when publishing it.

I've made some changes as per your suggestions. Just let me know if I'm going right.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Removed remnants from code.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -67,10 +69,14 @@ public UserTuneElement parse(XmlPullParser parser, int initialDepth, XmlEnvironm
                     	parser.next();
                     	builder.setTrack(Integer.parseInt(parser.nextText()));
                     	parser.next();
-                    	builder.setUri(parser.nextText());
+                        URI uri = null;
+                		uri = new URI(parser.nextText());
```
vanitasvitae: This might be null, so watch out. Better check, if parser.nextText() is null.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -67,10 +69,14 @@ public UserTuneElement parse(XmlPullParser parser, int initialDepth, XmlEnvironm
                     	parser.next();
                     	builder.setTrack(Integer.parseInt(parser.nextText()));
```
vanitasvitae: I'm not sure, if this will work in real life. You have to check, what name the current element has. Right now you are randomly assigning `parser.nextText()` to your elements. If however, the XML is not in the order you expect, this will fail.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
adiaholic: URI used in making a UserTuneElement and pushed changes.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,152 @@
...
+
+    public static UserTuneElement fromMessage(Message message) {
+        return message.getExtension(ELEMENT,NAMESPACE);
+    }
+
+    public UserTuneElement getUserTune() {
+        return this;
+    }
+
+    public static class UserTuneElementBuilder{
```
vanitasvitae: Some things about the Builder:  
Usually (but not necessarily) a Builder is simply named `Builder`. In this case it would make sense, as `UserTuneElement.Builder` is specific enough and `UserTuneElement.UserTuneElementBuilder` gets a little bit too long ;)  
Also a Builder has a private constructor and is created via a static method within the class that it is used to build (in this case that would be `public static UserTuneElement.Builder builder()` which belongs into the `UserTuneElement` class.  
All setter methods of a Builder have a return value of the Builder itself (eg. `public Builder setArtist(String artist)`) and return `this`. This comes in handy to chain setter calls, which make the Builder pattern unfold its full potential. Notice how convenient `UserTuneElement.builder().setArtist("Adele").setLength(1234).setRating(9).setXYZ().build()` is ;)  
Lastly, I haven't yet seen the technique to pass a builder instance to the constructor in order to create the object, but I don't see how this is a bad idea either :D

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,152 @@
...
+        public void setLength(int length) {
+            this.length = length;
+        }
+
+        public void setRating(int rating) {
+            if (rating <= 10 && rating >= 0) {
+                this.rating = rating;
+            }
+            else {
+                throw new IllegalArgumentException("Rating tried to set was out of bounds");
```
vanitasvitae: This would be a good opportunity to inform the user, which bounds they should stay inside ;)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Oops, didn't see this before making a commit.
Will act upon this in the following commit.
:)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Proper documentation added to inform user about the rating range.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
adiaholic: Arranged as per the suggestion.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTune.java
adiaholic: Created a builder to build UserTuneElement along with static method to create a builder from inside UserTuneElement.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -67,10 +69,14 @@ public UserTuneElement parse(XmlPullParser parser, int initialDepth, XmlEnvironm
...
                     	parser.next();
-                    	builder.setUri(parser.nextText());
+                        URI uri = null;
```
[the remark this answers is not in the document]
adiaholic: I've been trying to avoid tabs and indentation violation as much as possible, but still this one got through ;(
Fixed in next commit.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -157,43 +160,53 @@ public UserTuneElement getUserTune() {
...
             } else {
                 throw new IllegalArgumentException("Rating tried to set was out of bounds");
             }
+            return this;
         }
 
-        public void setSource(String source) {
+        public Builder setSource(String source) {
             this.source = source;
             ORDER.put(++orderIndex,"source");
```
vanitasvitae: what do you need `ORDER` for? It doesn't matter in which order your elements are in the XML.
You can simply define your own order of elements in toXML() by serializing "source" first, then artist, etc. (in any order you like). There shouldn't be a reason to preserve the order in which the user placed the values in the builder.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -46,74 +45,73 @@
...
+                if (UserTuneElement.ELEMENT.equals(name)) {
+                    return builder.build();
                 }
             }
         }
     }
 
     private boolean isNameInvalid(String name) {
-        if (name.equals("artist") || name.equals("length") || name.equals("rating") || name.equals("source") || name.equals("title") || name.equals("track") || name.equals("uri") || name.equals("tune")) {
+        if (name.equals("artist") || name.equals("length") || name.equals("rating") || name.equals("source")
```
vanitasvitae: I'd try something like `if (Arrays.asList("artist", "length", ...).contains(name))`

adiaholic: Yes, this looks much bettew :)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Yes you are right that we do not need to preserve the order in which those elements are received or sent, but at the same time it is important to remember which elements are to be send and which are not.
Not storing them will result in an XML of the form
	  <tune xmlns='http://jabber.org/protocol/tune'>
	          <artist>Yes</artist>
	          <length></length>
	          <rating></rating>
	          <source>Yessongs</source>
	          <title>Heart of the Sunrise</title>
	          <track></track>
	   </tune>
Such a transmission will me meaningless.
Storing will result in
	<tune xmlns='http://jabber.org/protocol/tune'>
	          <artist>Yes</artist>
	          <source>Yessongs</source>
	          <title>Heart of the Sunrise</title>
	   </tune>

adiaholic: Anyways there will surely be some way to do it without storing. Will figure it out and shortly.

vanitasvitae: well, you can simply check
```
if (artist != null) {
    xml.apped("<artist>" + artist + "</artist>");
}
```
so you only add the opening and closing tags for attributes that are set.

vanitasvitae: On the other hand, I'm not sure, if @Flowdalic will be pleased with manual opening xml tags via Strings like that. The proper solution would probably be to create dedicated `NamedElement`s, but on the other hand that might be overkill for this case...

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,223 @@
...
+            if (artist != null) {
+                userTuneXML.element("artist", artist);
+            }
+            if (length  != -1) {
+                userTuneXML.element("length", Integer.toString(length));
+            }
+            if (rating  != -1) {
+                userTuneXML.element("rating",Integer.toString(rating));
+            }
+            userTuneXML.element("source",source);
```
vanitasvitae: What happens, if the user hasn't set a source?
Hint: To solve this, take a look at `XmlStringBuilder.optElement()` ;)

vanitasvitae: Same goes for all the other fields as well.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,223 @@
...
+        return xml.closeElement(getElementName());
+    }
+
+    private CharSequence getUserTuneXML() {
+
+        XmlStringBuilder userTuneXML = new XmlStringBuilder();
+            if (artist != null) {
+                userTuneXML.element("artist", artist);
+            }
+            if (length  != -1) {
```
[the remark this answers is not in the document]
vanitasvitae: Ah, I see, the values are initialized in the Builder. Still, I think using an Integer object is probably the cleaner solution.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,223 @@
...
+
+        private Builder() {
+        }
+
+        public Builder setArtist(String artist) {
+            this.artist = artist;
+            return this;
+        }
+
+        public Builder setLength(int length) {
```
vanitasvitae: You do not prevent setting negative lengths here.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,223 @@
...
+            this.source = source;
+            return this;
+        }
+
+        public Builder setTitle(String title) {
+            this.title = title;
+            return this;
+        }
+
+        public Builder setTrack(int track) {
```
vanitasvitae: What about negative track numbers?

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,223 @@
...
+                userTuneXML.element("track",Integer.toString(track));
+            }
+            if (uri != null) {
+                userTuneXML.element("uri",uri.toString());
+            }
+        return userTuneXML.toString();
+    }
+
+    private boolean isNullUserTune() {
+        if (this.artist == null && this.length == 0 && this.rating == 0 && this.source == null && this.title == null
```
vanitasvitae: 0 or -1?

adiaholic: Thanks for pointing this out :)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: If the user hasn't set source, firstly the `builder.build()` method will give throw an IllegalArguementException with a message that "UserTune should atleast have 'title' and 'source'".
This is because I've considered keeping source as a compulsory element because If the user is listening to a tune, it has to be from somewhere, maybe a website or the local machine. Which means that there will be no case in which the source is undefined. Of course, we can make source optional as well , like we did for other elements. 
Incase you think that there is a way in which source could be undefined , it can be set as an optional element.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Track numbers managed :)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Negative lengths managed.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
vanitasvitae: > This is because I've considered keeping source as a compulsory element because If the user is listening to a tune, it has to be from somewhere, maybe a website or the local machine. Which means that there will be no case in which the source is undefined.

Is this defined in the XEP somewhere? I would be cautious to prevent certain use cases without being able to point at the XEP.

adiaholic: No, It is not stated anywhere in the XEP. And I realized that it is not wise to violate it.
Will act on it by keeping only the `title` as a compulsory element and others as optional elements.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,254 @@
...
+        userTuneXML.optElement("length", length);
+        userTuneXML.optElement("rating",rating);
+        userTuneXML.element("source",source);
+        userTuneXML.element("title",title);
+        userTuneXML.optElement("track",track);
+        userTuneXML.optElement("uri",uri);
+        return userTuneXML.toString();
+    }
+
+    private boolean isNullUserTune() {
```
vanitasvitae: I'd rename this to `isEmptyUserTune()`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,254 @@
...
+         * @param length
+         * @return Builder
+         */
+        public Builder setLength(int length) {
+            if (length > 0) {
+                this.length = length;
+                return this;
+            }
+        	else {
+        		throw new IllegalArgumentException("Length shoud be positive. Incase of live streaming, length is expected to be null");
```
vanitasvitae: typo "should", also "In case".

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+
+        /**
+         * This method is called to build a null UserTuneElement while disabling publish.
+         * @throws IllegalArgumentException
+         * @return UserTuneElement
+         */
+        public UserTuneElement buildEmptyUserTune() {
+            if (title == null || source == null || uri == null || artist == null || length == null || rating == null || track == null) {
```
vanitasvitae: I would simply return a new UserTuneElement without any values. Maybe add another private constructor for this.

vanitasvitae:
The one thing that's missing from this PR is the ability to register listeners for incoming UserTunes, as well as maybe some integration tests ;)

adiaholic:
I'm working on the integration tests,
but I wonder if 

> ability to register listeners for incoming UserTunes,

refers to `UserTuneManager.addUserTuneListener` ?

vanitasvitae:
Oh, apparently I'm blind :D

smack-integration-test/src/main/java/org/jivesoftware/smackx/usertune/UserTuneIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+                .build();
+
+        IntegrationTestRosterUtil.ensureBothAccountsAreSubscribedToEachOther(conOne, conTwo, timeout);
+
+        final SimpleResultSyncPoint userTuneReceived = new SimpleResultSyncPoint();
+
+        final UserTuneListener userTuneListener = new UserTuneListener() {
+            @Override
+            public void onUserTuneUpdated(BareJid jid, Message message, UserTuneElement userTuneElement) {
+                userTuneReceived.signal();
```
adiaholic: @vanitasvitae Here, signal is triggered without comparing `userTuneElement1` and `userTuneElement`. I wish to know, if there should be public getters to compare the contents of both `UserTuneElement` s ?
Inside `UserTuneElement`, `public void compareElement(UserTuneElement  element)` should do the trick .

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,258 @@
...
+            return this;
+        }
+
+        /**
+         * Length is an optional element in UserTuneElement.
+         * @param length
+         * @return Builder
+         */
+        public Builder setLength(int length) {
+            if (length > 0) {
```
Flowdalic: This is not sufficient: 'length' is specified as xs:unsignedShort, you need to enforce that, which probably means that you want to add an according method to Smack's` `NumberUtil`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,258 @@
...
+                this.length = length;
+                return this;
+            }
+            else {
```
Flowdalic: The control flow of this method could be made easier readable by following the
```java
requireXsUnsignedShort(input);
this.input = input;
return this;
```

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,258 @@
...
+        /**
+         * Rating is an optional element in UserTuneElement.
+         * The rating to be given is suppossed to be out of 10.
+         * <br>
+         * Method will throw an Exception if rating is less than 0 or greater than 10.
+         * @param rating
+         * @return Builder
+         */
+        public Builder setRating(int rating) {
+            if (rating <= 10 && rating >= 0) {
```
Flowdalic: This allows for rating to be 0, which is invalid as far as I can tell from XEP-0118.

Also should be simplified to:
```
if (rating < 1 || rating > 10) {
    throw new IllegalArgumentException("Meaningful exception message");
}
this.rating = rating;
return this;
````

File: smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java [hunk 26]
```diff
@@ -0,0 +1,258 @@
...
+         * @param track
+         * @return Builder
+         */
+        public Builder setTrack(int track) {
+            if (track > 0) {
+                this.track = track;
+                return this;
+            }
+            else {
+                throw new IllegalArgumentException("Track number should be positive.");
```
Flowdalic: Same simplification to improve the readability of the method as above.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,258 @@
...
+         * This method is called to build a UserTuneElement.
+         * @throws IllegalArgumentException
+         * @return UserTuneElement
+         */
+        public UserTuneElement build() {
+            if (title != null) {
+                return new UserTuneElement(this);
+            }
+            else {
+                throw new IllegalArgumentException("UserTune should atleast have 'title'");
```
Flowdalic: Same simplification to improve the readability of the method as above.

Besides I don't see a requirement in XEP-0118 that user tune element's must have at least 'title' set (although it appears sensible). Did I miss it?

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                    continue outerloop;
+                }
+                try {
+                    while (true) {
+                        switch (name) {
+                        case "artist":
+                            builder.setArtist(parser.nextText());
+                            break;
+                        case "length":
+                            builder.setLength(Integer.parseInt(parser.nextText()));
```
Flowdalic: Should use `ParserUtils.getIntegerFromNextText()`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                            break;
+                        case "rating":
+                            builder.setRating(Integer.parseInt(parser.nextText()));
```
Flowdalic: Should use `ParserUtils.getIntegerFromNextText()`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                            break;
+                        case "source":
+                            builder.setSource(parser.nextText());
+                            break;
+                        case "title":
+                            builder.setTitle(parser.nextText());
+                            break;
+                        case "track":
+                            builder.setTrack(Integer.parseInt(parser.nextText()));
```
Flowdalic: Should use `ParserUtils.getIntegerFromNextText()`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                            builder.setUri(uri);
+                            break;
+                        case "tune":
+                            if (UserTuneElement.ELEMENT.equals(name) && (tag == Event.END_ELEMENT)) {
+                                return builder.build();
+                            }
+                        }
+                        tag = parser.next();
+                        name = parser.getName();
+                        if (isNameInvalid(name)) {
```
Flowdalic: Throwing on unknown (child) elements is atypical. Unless that protocol requires that only the specified elements have to appear, unknown elements are usually ignored. Which helps to easily extend XMPP (hence the 'X' in XMPP).

As I did not found such a requirement in XEP-0188, I believe this check can be removed.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                        }
+                    }
+                } catch (IllegalArgumentException e) {
+                    throw new XmlPullParserException("Unknown tune value: " + name + " encountered.");
+                } catch (URISyntaxException e) {
+                    throw new IllegalArgumentException("Invalid URI");
+                }
+
+            case END_ELEMENT:
+                if (UserTuneElement.ELEMENT.equals(name)) {
```
Flowdalic: This should take (only) `depth` into account, which ensures we actually hit the correct closing tag (and not a closing tag of a different depth which just happens to have the same name).

smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers
```diff
@@ -587,6 +587,13 @@
           <elementName>mood</elementName>
           <namespace>http://jabber.org/protocol/mood</namespace>
           <className>org.jivesoftware.smackx.mood.provider.MoodProvider</className>
-        </extensionProvider>
+    </extensionProvider>
```
Flowdalic: Unrelated whitespace change, please revert that.

smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
```diff
@@ -0,0 +1,88 @@
...
+                "<title>Heart of the Sunrise</title>" +
+                "<track>3</track>" +
+                "<uri>http://www.yesworld.com/lyrics/Fragile.html#9</uri>" +
+                "</tune>";
+
+        URI uri = null;
+        try {
+            uri = new URI("http://www.yesworld.com/lyrics/Fragile.html#9");
+        } catch (URISyntaxException e) {
+            throw new IllegalArgumentException("Invalid URI");
```
Flowdalic: Why do you catch the exception just to throw it again? Simply let the method throw the `URISyntaxException`. This makes the code a lot more readable. Also never swallow Exceptions! With your code we would loose all the information included in `e`.

smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
```diff
@@ -0,0 +1,88 @@
...
+        UserTuneElement userTuneElement = builder.setArtist("Yes")
+                .setLength(686)
+                .setRating(8)
+                .setSource("Yessongs")
+                .setTitle("Heart of the Sunrise")
+                .setTrack(3)
+                .setUri(uri)
+                .build();
+        assertXmlSimilar(xml, userTuneElement.toXML().toString());
+        XmlPullParser parser = TestUtils.getParser(xml);
```
Flowdalic: `TestUtils.getParser()` is going to be phased out in favor of the new parameterized test pattern for parsers. You want

```java
    @ParameterizedTest
    @EnumSource(SmackTestUtil.XmlPullParserKind.class)
    userTuneElementProviderTest(SmackTestUtil.XmlPullParserKind parserKind) {
        …
        XmlPullParser parser = SmackTestUtil.getParserFor(…, parserKind);
        …
    }
````

File: smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java [hunk 36]
```diff
@@ -0,0 +1,88 @@
...
+import org.jivesoftware.smack.xml.XmlPullParserException;
+import org.jivesoftware.smackx.usertune.element.UserTuneElement;
+import org.jivesoftware.smackx.usertune.provider.UserTuneProvider;
+
+import org.junit.Test;
+
+public class UserTuneElementTest extends SmackTestSuite {
+
+    @Test
+    public void toXmlTest() throws IOException, XmlPullParserException, SmackParsingException {
```
Flowdalic: This does actually test two things:
- transformation from the Java Object to XML
- transformation from XML to the Java Object

those are two independent aspects and hence should be tested in two different methods.

smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
```diff
@@ -0,0 +1,88 @@
...
+                .setUri(uri)
+                .build();
+        assertXmlSimilar(xml, userTuneElement.toXML().toString());
+        XmlPullParser parser = TestUtils.getParser(xml);
+        UserTuneElement parsed = UserTuneProvider.INSTANCE.parse(parser);
+        assertEquals(xml, parsed.toXML().toString());
+    }
+
+    @Test(expected = XmlPullParserException.class)
+    public void unknownUserTuneValueExceptionTest() throws Exception {
```
Flowdalic: Please avoid `throws Exception`, instead explicitly state the checked Exceptions this is potentially going to throw.

smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneManagerTest.java
```diff
@@ -0,0 +1,63 @@
...
+        builder.setLength(686);
+        builder.setRating(8);
+        builder.setSource("Yessongs");
+        builder.setTitle("Heart of the Sunrise");
+        builder.setTrack(3);
+        URI uri = null;
+        try {
+            uri = new URI("http://www.yesworld.com/lyrics/Fragile.html#9");
+        } catch (URISyntaxException e) {
+            throw new IllegalArgumentException("Invalid URI");
```
Flowdalic: Same as above.

smack-integration-test/src/main/java/org/jivesoftware/smackx/usertune/UserTuneIntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+        try {
+            utm1.publishUserTune(userTuneElement1);
+            userTuneReceived.waitForResult(timeout);
+        } finally {
+            utm2.removeUserTuneListener(userTuneListener);
+        }
+    }
+
+    private boolean isEqual(UserTuneElement usertuneElement,UserTuneElement userTuneElement1) {
+        if (usertuneElement.toXML().toString().contentEquals(userTuneElement1.toXML().toString())) {
```
Flowdalic: This can be simplified to `return usertuneElement.toXML().toString().contentEquals(userTuneElement1.toXML().toString();`

smack-integration-test/src/main/java/org/jivesoftware/smackx/usertune/UserTuneIntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+                                                  .build();
+
+        IntegrationTestRosterUtil.ensureBothAccountsAreSubscribedToEachOther(conOne, conTwo, timeout);
+
+        final SimpleResultSyncPoint userTuneReceived = new SimpleResultSyncPoint();
+
+        final UserTuneListener userTuneListener = new UserTuneListener() {
+            @Override
+            public void onUserTuneUpdated(BareJid jid, Message message, UserTuneElement userTuneElement) {
+                if (isEqual(userTuneElement,userTuneElement1)) {
```
Flowdalic: Why don't you add an `equals()` method to `UserTuneElement`?

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                        tag = parser.next();
+                        name = parser.getName();
+                        if (isNameInvalid(name)) {
+                            throw new XmlPullParserException("Invalid elements inside <tune/>");
+                        }
+                    }
+                } catch (IllegalArgumentException e) {
+                    throw new XmlPullParserException("Unknown tune value: " + name + " encountered.");
+                } catch (URISyntaxException e) {
+                    throw new IllegalArgumentException("Invalid URI");
```
Flowdalic: Should be `SmackUriSyntaxParsingException` instead.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                            }
+                        }
+                        tag = parser.next();
+                        name = parser.getName();
+                        if (isNameInvalid(name)) {
+                            throw new XmlPullParserException("Invalid elements inside <tune/>");
+                        }
+                    }
+                } catch (IllegalArgumentException e) {
+                    throw new XmlPullParserException("Unknown tune value: " + name + " encountered.");
```
Flowdalic: You don't have to catch this exception, simply let it throw.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,115 @@
...
+                            builder.setSource(parser.nextText());
+                            break;
+                        case "title":
+                            builder.setTitle(parser.nextText());
+                            break;
+                        case "track":
+                            builder.setTrack(Integer.parseInt(parser.nextText()));
+                            break;
+                        case "uri":
+                            URI uri = new URI(parser.nextText());
```
Flowdalic: If you use `ParserUtils.getUriFromNextText()` here, then you can remove the whole try-catch block.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: No, it is not stated anywhere. Though not sharing `title` would make less sense, I understand that it is not wise to contravene a XEP. I will make the corresponding changes.

smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers
adiaholic: This change was intentional.
[Master branch](https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers).

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,246 @@
...
+        userTuneXML.optElement("rating", rating);
+        userTuneXML.element("source", source);
+        userTuneXML.element("title", title);
+        userTuneXML.optElement("track", track);
+        userTuneXML.optElement("uri", uri);
+        return userTuneXML.toString();
+    }
+
+    private boolean isEmptyUserTune() {
+        if (this.artist == null && this.length == null && this.rating == null && this.source == null && this.title == null
```
Flowdalic: Ok, you seem to really like this, which is not uncommon for an inexperienced programmer. What you really want to write here is

```
return this.artist == null && this.length == null && this.rating == null && this.source == null && this.title == null 
           && this.track == null && this.uri == null;
```

Or, now that `UserTuneElement` has `hashCode()` and `equals()`, why not simply `this.equals(EMPTY_USER_TUNE)`? A lot more readable, no?

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public static final String NAMESPACE = "http://jabber.org/protocol/tune";
+    public static final String ELEMENT = "tune";
+
+    private final String artist;
+    private final Integer length;
+    private final Integer rating;
+    private final String source;
+    private final String title;
+    private final Integer track;
```
Flowdalic: Careful, XEP-0118 is a little bit nasty here: 'track' is specified as xs:string. Which means we must accept any string as 'track'.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,246 @@
...
+        return message.hasExtension(ELEMENT, NAMESPACE);
+    }
+
+    public static UserTuneElement from(Message message) {
+        return message.getExtension(ELEMENT, NAMESPACE);
+    }
+
+    @Override
+    public int hashCode() {
+        return super.hashCode();
```
Flowdalic: That is wrong because it breaks the API contract of `hashCode()` and `equals()`.

smack-extensions/src/main/resources/org.jivesoftware.smack.extensions/extensions.providers
Flowdalic: But it is unrelated to the xep118 implementation and hence should be in an extra commit.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Oh, that smoothly got unnoticed.
Thanks for pointing that out :)

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,142 @@
...
+
+    }
+
+    public void clearUserTune() throws NotLoggedInException, NotALeafNodeException, NoResponseException, NotConnectedException, XMPPErrorException, InterruptedException {
+        publishUserTune(UserTuneElement.EMPTY_USER_TUNE);
+    }
+
+    public void publishUserTune(UserTuneElement userTuneElement) throws NotLoggedInException, NotALeafNodeException, NoResponseException, NotConnectedException, XMPPErrorException, InterruptedException {
+
+        LeafNode node = pubSubManager.getOrCreateLeafNode(USERTUNE_NODE);
```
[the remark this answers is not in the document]
adiaholic: @Flowdalic I've looked at other XEP implementations based on PEP and every single one of them uses PubSub. It is alright to still go with PepManager instead of PubsubManager ?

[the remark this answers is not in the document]
Flowdalic: Which one? I'd say you should use `PepManager` whenever it is possible, which appears to be the case here. If `PepManager` does not provide the required functionality, then we should evaluate if we should put it there.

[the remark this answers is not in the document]
adiaholic: @Flowdalic 
Xep-0080 and xep-0107  seem to be based on pep along with xep-0108 which has not been implemented yet.
[Geolocation](https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/java/org/jivesoftware/smackx/geoloc/GeoLocationManager.java)
[User Mood](https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/java/org/jivesoftware/smackx/mood/MoodManager.java )
 Please take a look at these and let me know what you think should be my approach. You might want me to give minor touching to these other modules as well 😉

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public static synchronized UserTuneManager getInstanceFor(XMPPConnection connection) throws NotLoggedInException {
+        UserTuneManager manager = INSTANCES.get(connection);
+        if (manager == null) {
+            manager = new UserTuneManager(connection);
+            INSTANCES.put(connection, manager);
+        }
+        return manager;
+    }
+    private UserTuneManager(XMPPConnection connection) throws NotLoggedInException {
```
Flowdalic: nit: missing newline

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,130 @@
...
+                        UserTuneElement tune = (UserTuneElement) payload.getPayload();
+
+                        for (UserTuneListener listener : userTuneListeners) {
+                            listener.onUserTuneUpdated(contact, message, tune);
+                        }
+                    }
+                });
+            }
+        });
+        ServiceDiscoveryManager.getInstanceFor(connection).addFeature(USERTUNE_NOTIFY);
```
Flowdalic: This should not be done unconditionally in the constructor as it signals that the user wants to receive usertune notifications. How about creating `enableUserTuneNotifications()` and `disableUserTuneNotitications()` methods? Also there should be a setting `ENABLE_USER_TUNE_NOTIFICATIONS_BY_DEFAULT` which can be configured, defaults to `true` and is used in the constructor to enable the notifications.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,130 @@
...
+        super(connection);
+        PepManager.getInstanceFor(connection).addPepListener(new PepListener() {
+            @Override
+            public void eventReceived(EntityBareJid from, EventElement event, Message message) {
+                if (!USERTUNE_NODE.equals(event.getEvent().getNode())) {
+                    return;
+                }
+
+                final BareJid contact = from.asBareJid();
+                asyncButOrdered.performAsyncButOrdered(contact, new Runnable() {
```
Flowdalic: Feel free to use a lambda here instead of a `Runnable`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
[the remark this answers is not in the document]
Flowdalic: Yep, we should probably check if those shouldn't use `PepManager` instead. Possible that the code predates, at least some functionaly of, `PepManager`.

smack-core/src/main/java/org/jivesoftware/smack/util/NumberUtil.java
```diff
@@ -28,7 +28,21 @@ public static void checkIfInUInt32Range(long value) {
             throw new IllegalArgumentException("unsigned 32-bit integers can't be negative");
         }
         if (value > ((1L << 32) - 1)) {
-            throw new IllegalArgumentException("unsigned 32-bit integers can't be greater then 2^32 - 1");
+            throw new IllegalArgumentException("unsigned 32-bit integers can't be greater than 2^32 - 1");
```
Flowdalic: I'd really like to see those unrelated fixes - thanks for that - in an extra commit, since they are unrelated to this change.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
adiaholic: I wanted to use lambdas since ages.
😌

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,253 @@
...
+        return message.hasExtension(ELEMENT, NAMESPACE);
+    }
+
+    public static UserTuneElement from(Message message) {
+        return message.getExtension(ELEMENT, NAMESPACE);
+    }
+
+    @Override
+    public int hashCode() {
+        final int prime = 23;
```
Flowdalic: I have just pushed `o.j.s.util.HashCode`, you should use it here.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,253 @@
...
+        result = prime * result + (source == null ? 0 : source.length());
+        result = prime * result + (title == null ? 0 : title.length());
+        result = prime * result + (track == null ? 0 : track.length());
+        result = prime * result + (uri == null ? 0 : uri.toString().length());
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
```
Flowdalic: I have just pused `o.j.s.util.EqualsUtil`, you should use it here.

smack-core/src/main/java/org/jivesoftware/smack/util/NumberUtil.java
```diff
@@ -42,35 +42,6 @@ public static long requireUInt32(long value) {
         if (value > ((1L << 32) - 1)) {
             throw new IllegalArgumentException("unsigned 32-bit integers can't be greater than 2^32 - 1");
         }
-    }
```
adiaholic: @Flowdalic I think I have messed it up here. I think it is possible to undo changes on `NumberUtils.java` using `git rebase` by using `drop`. But I don't want to mess it up further, so thought better to ask you before I make it any worse.

Flowdalic: The commits of this PR surely need cleanup, but the good news is that the diff looks good. So I'd simply fix the remaining issues and then squash everything into a single commit.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,251 @@
...
+        XmlStringBuilder xml = new XmlStringBuilder(this);
+        if (isEmptyUserTune()) {
+            return xml.closeEmptyElement();
+        }
+        xml.rightAngleBracket();
+        xml.optAppend(getUserTuneXML());
+        return xml.closeElement(getElementName());
+    }
+
+    private CharSequence getUserTuneXML() {
```
Flowdalic: Why is that an extra method? Please simply write it into `toXML()`.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    public static UserTuneElement from(Message message) {
+        return message.getExtension(ELEMENT, NAMESPACE);
+    }
+
+    @Override
+    public int hashCode() {
+        HashCode.Builder builder = HashCode.builder();
+        builder.append(this);
```
Flowdalic: That will only calculate the hash code of the `Object` instance of `this`, i.e. the identity hash code. Which is not really what we want here, as the outcome would be the same as if the `hashCode()` method was not overridden.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,251 @@
...
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        EqualsUtil.EqualsComperator<UserTuneElement> comparator = new EqualsComperator<UserTuneElement>() {
+
+            @Override
+            public void compare(org.jivesoftware.smack.util.EqualsUtil.Builder equalsBuilder, UserTuneElement other) {
+                equalsBuilder.append(this, other);
```
Flowdalic: That will only compares `Object` instance of `this`. Which is not really what we want here, as the outcome would be the same as if the `equals()` method was not overridden.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
Flowdalic: I just consulted xep118, where track is specified as xs:string, so this needs to go away.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+        UserTuneElement.Builder builder = UserTuneElement.getBuilder();
+        XmlPullParser.TagEvent tag = parser.nextTag();
+        outerloop: while (true) {
+            switch (tag) {
+            case START_ELEMENT:
+                String name = parser.getName();
+                String namespace = parser.getNamespace();
+                if (!UserTuneElement.NAMESPACE.equals(namespace)) {
+                    LOGGER.log(Level.FINE, "Foreign namespace " + namespace + " detected.");
```
Flowdalic: Please remove that log statement

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                    continue outerloop;
+                }
+            while (tag == XmlPullParser.TagEvent.START_ELEMENT) {
```
Flowdalic: The indentation seems off here, which makes the already complicated logic way harder to read and comprehend.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/provider/UserTuneProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                        tag = parser.nextTag();
+                        name = parser.getName();
+                    }
+                break;
+            case END_ELEMENT:
+                if (parser.getDepth() == initialDepth) {
+                    break outerloop;
+                }
+                break;
+            default:
```
Flowdalic: The `default` case can now be removed since you use `nextTag()` and all values of the returned enum are already covered in the switch-case.

adiaholic: Yes Right !

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
adiaholic: Yes !

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,144 @@
...
+                    return;
+                }
+
+                final BareJid contact = from.asBareJid();
+                asyncButOrdered.performAsyncButOrdered(contact, () -> {
+                    ItemsExtension items = (ItemsExtension) event.getExtensions().get(0);
+                    PayloadItem<?> payload = (PayloadItem) items.getItems().get(0);
+                    UserTuneElement tune = (UserTuneElement) payload.getPayload();
+
+                    for (UserTuneListener listener : userTuneListeners) {
```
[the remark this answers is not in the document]
Flowdalic: That is still pending. Access to `userTuneListeners` is still not properly synchronized.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/element/UserTuneElement.java
```diff
@@ -0,0 +1,271 @@
...
+            return this;
+        }
+
+        /**
+         * Length is an optional element in UserTuneElement.
+         * @param length
+         * @return Builder
+         */
+        public Builder setLength(int length) {
+            this.length = UInt16.from(length);
```
Flowdalic: Should be `return setLength(UInt16.from(length))`, so that the actually setting of the length is delegated to a single method.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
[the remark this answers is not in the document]
Flowdalic: It is still broken :(
You either want
```
synchronized (userTuneListeners) {
   for (UserTuneLIstener listener : userTuneListeners) {
   …
   }
}
```
or make userTuneLIsteners a CopyOnWriteArraySet.

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
```diff
@@ -0,0 +1,146 @@
...
+                });
+            }
+        });
+        if (ENABLE_USER_TUNE_NOTIFICATIONS_BY_DEFAULT) {
+            enableUserTuneNotifications();
+        }
+        pubSubManager = PubSubManager.getInstanceFor(getAuthenticatedConnectionOrThrow(), connection().getUser().asBareJid());
+    }
+
+    public void setDefaultNotificationsValue(boolean bool) {
```
Flowdalic: The method name is not clear. Should probably be `setUserTuneNotificationsEnabledByDefault(boolean)`

smack-extensions/src/main/java/org/jivesoftware/smackx/usertune/UserTuneManager.java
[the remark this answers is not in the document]
Flowdalic: There is now `PepManager.publish()` which this method should use.

smack-extensions/src/test/java/org/jivesoftware/smackx/usertune/UserTuneElementTest.java
```diff
@@ -0,0 +1,75 @@
...
+                .build();
+        assertXmlSimilar(xml, userTuneElement.toXML().toString());
+    }
+
+    @ParameterizedTest
+    @EnumSource(value = SmackTestUtil.XmlPullParserKind.class)
+    public void userTuneElementProviderTest(XmlPullParserKind parserKind) throws XmlPullParserException, IOException, SmackParsingException {
+        XmlPullParser parser = SmackTestUtil.getParserFor(xml, parserKind);
+        UserTuneElement parsed = UserTuneProvider.INSTANCE.parse(parser);
+        assertEquals(xml, parsed.toXML().toString());
```
Flowdalic: Should use `XmlUnitUtils.assertXmlSimilar()`, so that slight changes in the XML serialization do not trigger the test.

Flowdalic:

Cherry picked as https://github.com/igniterealtime/Smack/commit/260c5539b55ef2c9186003a42064801c074f96af

Please note that some fixes where necessary: https://github.com/igniterealtime/Smack/commit/926c5892ad985b9fe5ef11104effa7400226304e

PR #317, opened 2019-05-03
WIP: OMEMO Media Sharing via HTTP File Upload

vanitasvitae:
This PR adds support for [OMEMO Media Sharing](https://xmpp.org/extensions/inbox/omemo-media-sharing.html) to Smacks HttpFileUploadManager.

The manager can now upload files encrypted to the server. As a result the user gets a aesgcm url-string which can be shared with others. There is also a function that retrieves a `Cipher` from an aesgcm string, which can then be used to decrypt the file/stream.

Since OMEMO Media Sharing is basically independent of OMEMO, I chose the HttpFileUploadManager as the best place to implement this feature.

This PR is work in progress. There are some rough edges, but the code works and is Integration-tested.
Missing things include:
* [ ] decide, whether or not Smack should provide a function to convert a aesgcm string into a URL
* [ ] figure out, whether a aesgcm:// url should be translated to http or https
* [ ] maybe create custom data type for aesgcm urls?
* [ ] code clean up
* [ ] decide, if a method to download + decrypt files would be necessary


vanitasvitae:
I'm not quite happy with this as of now. I'm not entirely sure, if I used the AlgorithmParameterSpecs correctly. Currently the decrypted file is 32 bytes longer than the original. This is probably the authentication tag and/or IV. However, since the tag length of AESGCM is configurable, I'm not sure, how one can calculate that length reliably when receiving a file.

cmeng-git:
aTalk v1.8.3 has implemented OMEMO Media Sharing using your source. Thanks for the contribution.

vanitasvitae:
The integration test does now no longer fail.

cmeng-git:
I have raised the PR as requested i.e.
[PR for XEP-0454 Media File Sharing #680](https://github.com/igniterealtime/Smack/pull/680)

PR #320, opened 2019-05-16
Add note about BookmarkManager to MUCManager

vanitasvitae:
Since this came up in the Forums, I thought it would be a good idea to add a note in the javadoc.

This PR adds a note about `BookmarkManager.getBookmarkedConferences()` to the Javadoc of `MultiUserChatManager.getJoinedRooms()`.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
@@ -282,6 +282,9 @@ public final class MultiUserChatManager extends Manager {
      * Returns a Set of the rooms where the user has joined. The Iterator will contain Strings where each String
      * represents a room (e.g. room@muc.jabber.org).
      *
+     * Note: In order to get a list of bookmarked (but not necessarily joined) conferences, use
+     * {@link org.jivesoftware.smackx.bookmarks.BookmarkManager#getBookmarkedConferences()}.
+     *
      * @return a List of the rooms where the user has joined using a given connection.
      */
     public Set<EntityBareJid> getJoinedRooms() {


smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -282,6 +282,9 @@ public boolean isServiceEnabled(Jid user) throws NoResponseException, XMPPErrorE
      * Returns a Set of the rooms where the user has joined. The Iterator will contain Strings where each String
      * represents a room (e.g. room@muc.jabber.org).
      *
+     * Note: A more modern way to determine which MUCs the user might be joined in on other clients is using
+     * {@link BookmarkManager#getBookmarkedConferences()}.
```
Flowdalic: Could you fully qualify `BookmarkManager` here so that the import statement becomes unnecessary?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -282,6 +282,9 @@ public boolean isServiceEnabled(Jid user) throws NoResponseException, XMPPErrorE
      * Returns a Set of the rooms where the user has joined. The Iterator will contain Strings where each String
      * represents a room (e.g. room@muc.jabber.org).
      *
+     * Note: A more modern way to determine which MUCs the user might be joined in on other clients is using
```
Flowdalic: Please remove the 'modern'.

Flowdalic: I also believe this should go into the javadoc of `getJoinedRooms(EntityJid)` instead, no?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
vanitasvitae: I wasn't sure, whether or not the import statement would do harm, but sure, I can remove it :)

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
vanitasvitae: As far as I understand, the bookmark manager is only used for the user themselves and cannot be used to determine bookmarks of contacts. `getJoinedRooms(EntityJid)` however can be used to determine, which mucs contacts are joined (right?). Therefore I decided to add the note only to the method that is used to query yourself.

Flowdalic: Right, I was confused.

PR #322, opened 2019-05-31
Adding range checker for unsignedShort in NumberUtil.

adiaholic:
Also corrected a typing error.


adiaholic:
Closing this because already implemented in #316 .
Another PR would make less sense.

PR #324, opened 2019-07-09
Correcting mini typing error

adiaholic:


adiaholic:
@vanitasvitae 
This is THE issue I faced multiple times, while building locally.
Also I am pretty sure that if I ran a build the second time, the project will build successfully.
What do you think is a solution to prevent such a situation ?

PR #325, opened 2019-07-26
Implement xml:lang support in ConnectionConfiguration and AbstractXMPPConnection

ge0rg:
This patch makes it possible to change the stream-level language as part
of the connection configuration, to allow a properly implemented
entities to provide i18n'ed response messages.

This code does not cover XMPPBOSHConnection!

X-Signed-Off-By: Georg Lukas <georg@op-co.de>

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
@@ -2012,7 +2012,7 @@ public abstract class AbstractXMPPConnection implements XMPPConnection {
             from = XmppStringUtils.completeJidFrom(localpart, to);
         }
         String id = getStreamId();
-        sendNonza(new StreamOpen(to, from, id));
+        sendNonza(new StreamOpen(to, from, id, config.getXmlLang(), StreamOpen.StreamContentNamespace.client));
     }
 
     public static final class SmackTlsContext {
--- a/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
@@ -24,6 +24,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -118,6 +119,8 @@ public abstract class ConnectionConfiguration {
     private final String password;
     private final Resourcepart resource;
 
+    private final Locale language;
+
     /**
      * The optional SASL authorization identity (see RFC 6120 § 6.3.8).
      */
@@ -164,6 +167,8 @@ public abstract class ConnectionConfiguration {
 
         // Resource can be null, this means that the server must provide one
         resource = builder.resource;
+
+        language = builder.language;
 
         xmppServiceDomain = builder.xmppServiceDomain;
         if (xmppServiceDomain == null) {
@@ -477,6 +482,34 @@ public abstract class ConnectionConfiguration {
     }
 
     /**
+     * Returns the stream language to use when connecting to the server.
+     *
+     * @return the stream language to use when connecting to the server.
+     */
+    public Locale getLanguage() {
+        return language;
+    }
+
+    /**
+     * Returns the xml:lang string of the stream language to use when connecting to the server.
+     *
+     * <p>If the developer sets the language to null, this will also return null, leading to
+     * the removal of the xml:lang tag from the stream. If a Locale("") is configured, this will
+     * return "", which can be used as an override.</p>
+     *
+     * @return the stream language to use when connecting to the server.
+     */
+    public String getXmlLang() {
+        // TODO: Change to Locale.toLanguageTag() once Smack's minimum Android API level is 21 or higher
+        // This will need a workaround for new Locale("").getLanguageTag() returning "und". Expected
+        // behavior of this function:
+        //  - returns null if language is null
+        //  - returns "" if language.getLanguage() returns the empty string
+        //  - returns language.toLanguageTag() otherwise
+        return (language != null) ? language.toString().replace("_", "-") : null;
+    }
+
+    /**
      * Returns the optional XMPP address to be requested as the SASL authorization identity.
      *
      * @return the authorization identifier.
@@ -564,6 +597,7 @@ public abstract class ConnectionConfiguration {
         private CharSequence username;
         private String password;
         private Resourcepart resource;
+        private Locale language = Locale.ENGLISH;
         private boolean sendPresence = true;
         private ProxyInfo proxy;
         private CallbackHandler callbackHandler;
@@ -685,6 +719,22 @@ public abstract class ConnectionConfiguration {
          */
         public B setResource(Resourcepart resource) {
             this.resource = resource;
+            return getThis();
+        }
+
+        /**
+         * Set the stream language.
+         * <p>
+         * (see RFC 6120 § 4.7.4).
+         * </p>
+         *
+         * @param language the language to use.
+         * @return a reference to this builder.
+         * @see <a href="https://tools.ietf.org/html/rfc6120#section-4.7.4">RFC 6120 § 4.7.4</a>
+         * @see <a href="https://www.w3.org/TR/xml/#sec-lang-tag">XML 1.0 § 2.12 Language Identification</a>
+         */
+        public B setLanguage(Locale language) {
+            this.language = language;
             return getThis();
         }
 


GitCop:

 * Commit: ffd3432cf74d871512e7fc61b60fdc7ad25a4ac2

---

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
```diff
@@ -118,6 +118,8 @@
     private final String password;
     private final Resourcepart resource;
 
+    private final String language;
```
Flowdalic: This should be of type `java.util.Locale`.

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
```diff
@@ -688,6 +702,21 @@ public B setResource(Resourcepart resource) {
...
+        /**
+         * Set the stream language.
+         * <p>
+         * (see RFC 6120 § 4.7.4).
+         * </p>
+         *
+         * @param language the language to use.
+         * @return a reference to this builder.
+         * @see <a href="https://tools.ietf.org/html/rfc6120#section-4.7.4">RFC 6120 § 4.7.4</a>
+         */
```
Flowdalic: I suggest to additionally add a `@see` to linking to "XML 1.0 § 2.12" https://www.w3.org/TR/xml/#sec-lang-tag

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
ge0rg: Isn't a locale more than just the language? Are we going to have issues of ambiguity here?

Flowdalic: > Isn't a locale more than just the language?

"A locale" yes, it describes how your date is formatted or your preferred input method

But think of `java.util.Locale` as a wrapper around BCP 47, which is what `xml:lang` also uses.

> Are we going to have issues of ambiguity here?

I don't see any.

GitCop:

 * Commit: ffd3432cf74d871512e7fc61b60fdc7ad25a4ac2

---

ge0rg:
So I've updated the code, just to realize that `Locale.toLanguageTag()` didn't exist in Android prior to r21. No idea how it even compiled with targetSdkLevel 17. I suppose the least ugly workaround would be to use `Locale.toString.replace("_", "-")`. Other ideas?

Flowdalic:
> No idea how it even compiled with targetSdkLevel 17.

What did you compile against an Android API 17 runtime jar? Smack, besides smack-android and smack-android-extensions is compiled with the Java SE runtime jar set bootstrapClasspath.

But do not worry, Smack also uses AnimalSniffer to ensure Android compatibility, which got you covered here and is the reason travis fails for the PR in the current state.



> I suppose the least ugly workaround would be to use `Locale.toString.replace("_", "-")`. Other ideas?

Works for me. We can always switch to a better approach if we come up with one. Please also add the following comment:
```
// TODO: Change to Locale.toLanguageTag() once Smack's minimum Android API level is 21 or higher
```

Thanks

GitCop:

 * Commit: ffd3432cf74d871512e7fc61b60fdc7ad25a4ac2

---

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
```diff
@@ -577,7 +578,7 @@ public boolean isEnabledSaslMechanism(String saslMechanism) {
         private CharSequence username;
         private String password;
         private Resourcepart resource;
-        private String language = "en";
+        private Locale language = Locale.ENGLISH;
```
Flowdalic: Is there any reason to not use `Locale.getDefault()` to initialize the value?

ge0rg: I didn't dare to change the default, which was "en". If you feel that using the default locale is good, and it's not going to break anything, I can patch it in.

vanitasvitae: I'm pretty sure, that there may be *some* tests that might break on non-english systems.
There was one such case pretty recently, but fixing those tests in the first place is probably a good idea :D

Flowdalic: What @vanitasvitae said :)

smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
```diff
@@ -2012,7 +2012,7 @@ protected void sendStreamOpen() throws NotConnectedException, InterruptedExcepti
             from = XmppStringUtils.completeJidFrom(localpart, to);
         }
         String id = getStreamId();
-        sendNonza(new StreamOpen(to, from, id));
+        sendNonza(new StreamOpen(to, from, id, config.getLanguage().toString().replace("_", "-"), StreamOpen.StreamContentNamespace.client));
```
Flowdalic: This could cause an NPE if `language` is `null`. I suggest to create a `ConnectionConfiguration.getLanguageTag()` method which returns `null` if language is `null`. This is also the method the TODO comment should go to, which I am still missing.

ge0rg: Oh, right. The TODO comment! Sorry about that! Is there a particular reason why `getLanguageTag` should return `null` and not the default "en" value?

Flowdalic: > Sorry about that! Is there a particular reason why `getLanguageTag` should return `null`

It is not `getLanguageTag()` that I am worried about returning `null`, it is `getLanguage()`. Currently it is possible to set the language to `null` in the configuration, and that is desirable given that `xml:lang` in stream open elements is optional. But the code should take that into consideration and not throw NPEs in that, arguably rare, case.

ge0rg: Understood. The questions that I'm looking at, are:
- is it reasonable to allow the user to supply a null Locale?
- what is the expected outcome in that case? "en", "" or no xml:lang tag at all?

Flowdalic: > * is it reasonable to allow the user to supply a null Locale?

I think so, especially for integration tests and the like. The point is, the standard is designed with an non-existing xml:lang in the stream open element in mind, and a library implementing the standard should allow for it. Of course, users deliberately providing a `null` value here, go down the "you should know what you are doing" road. And that is why Smack defaults to a non-empty xml:lang.

> * what is the expected outcome in that case? "en", "" or no xml:lang tag at all?

No xml:lang at all, otherwise the user would have provided `Locale.ENGLISH` for "en" or a `Locale` which returns the empty string as language tag for "".

The latter case is especially interesting: values where the empty string is allowed often cause pain, as, for example, their semantic is usually not as clearly defined as people believe to be or cause different behavior by different implementations. I really would like to be able to investigate how server implementations behave if a stream open with `xml:lang=""` is send (AFAIKT it is not explicitly forbidden by the standard). That said, given the current design we are aiming for, it is not possible to configure Smack to do so. That is because

```java
jshell> var l = new java.util.Locale("");
l ==> 
jshell> l.getLanguage()
$3 ==> ""
jshell> l.toLanguageTag()
$4 ==> "und"
```

so once we use `toLangaugeTag()` which is the goal, using a Locale constructed with the empty string as sole argument, will cause toLanguageTag() to return "und" (which I assume stands for 'undefined'). Hence the stream open will carry `xml:lang="und".

That means we should probably not introduce a method `ConnectionConfiguration.getLanguageTag()`, but `ConnectionConfiguration.getXmlLang()`, which
- returns `null` if `language` is null
- returns "" if `language.getLanguage()` returns the empty string
- returns `language.toLanguageTag()` otherwise

GitCop:

 * Commit: ffd3432cf74d871512e7fc61b60fdc7ad25a4ac2

---

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
```diff
@@ -490,6 +490,18 @@ public Locale getLanguage() {
...
 
+    /**
+     * Returns the xml:lang string of the stream language to use when connecting to the server.
+     *
+     * @return the stream language to use when connecting to the server.
+     */
+    public String getXmlLang() {
+        // TODO: use language.getLanguageTag() once we are on Java1.8 / Android r21+
+        // This will need a workaround for new Locale("").getLanguageTag() returning "und".
+        // Discussion in https://github.com/igniterealtime/Smack/pull/325
```
Flowdalic: I would prefer an in-code comment summarizing the discussion, but this is also ok.

smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java
ge0rg: The "nice" thing about that is that now we are leaking the environment of the user into the test setup, and from my experience, it will only break on locales that are not used by the maintainers in the first place, like Turkish (capital "i"), or Arabic (number parsing / formatting will bite you!). Feel free to rework the test setup to iterate through all existing locales, but I'm keeping this status-quo as is in this PR.

GitCop:

 * Commit: ffd3432cf74d871512e7fc61b60fdc7ad25a4ac2

---

Flowdalic:
This looks like it is good to go. Please squash the commits.

GitCop:

 * Commit: faaf8aaa48aa3312775c2603157a9ec970b5638f

---

Flowdalic:
Applied with minor modifications as https://github.com/igniterealtime/Smack/commit/6a0e0f0f67b3faaa4c4061ba4de6f0c298c424a8

Thanks for your contribution.

PR #328, opened 2019-07-30
Remove namespace from Uri's XmlStringBuilder

adiaholic:
The intended XML is 
` <uri type='image/jpeg'>
        http://www.shakespeare.lit/clients/exodus.jpg
 </uri> `
in place of
`<uri xmlns='urn:xmpp:media-element' type='image/jpeg'>
	http://www.shakespeare.lit/clients/exodus.jpg
</uri>`.
So remove `xmlEnvironment` as a parameter from XmlStringBuilder Constructor.


Flowdalic:
> The intended XML is
> `<uri type='image/jpeg'> http://www.shakespeare.lit/clients/exodus.jpg </uri>`
> in place of
> `<uri xmlns='urn:xmpp:media-element' type='image/jpeg'> http://www.shakespeare.lit/clients/exodus.jpg </uri>`.

That depends on the context where the element is embedded, no? You shouldn't need to remove the parameter. The XML serialization method should be able to determine if the namespace attribute is required or not. If this doesn't work as expected, then this has to be fixed.

PR #329, opened 2019-07-30
Improvements to the MUC join code

ge0rg:
This code addresses some corner cases when joining certain MUCs, like dino@chat.im, and provides `MUC.getRoomInfo()` to access the room information from the disco#info that is queried before join.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -329,7 +329,7 @@ public class MultiUserChat {
                     XMPPErrorException, InterruptedException, NotAMucServiceException {
         final DomainBareJid mucService = room.asDomainBareJid();
         if (!KNOWN_MUC_SERVICES.containsKey(mucService)) {
-            if (multiUserChatManager.providesMucService(mucService)) {
+            if (multiUserChatManager.providesMucService(mucService) || multiUserChatManager.providesMucService(room)) {
                 KNOWN_MUC_SERVICES.put(mucService, null);
             } else {
                 throw new NotAMucServiceException(this);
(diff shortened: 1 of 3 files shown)


Flowdalic:
Aren't some of the commits in this PR the same as we have already discussed in https://github.com/Flowdalic/Smack/pull/21 ?

ge0rg:
Yes, the first two were part of that PR. You've created a long-term battle plan in [SMACK-860](https://issues.igniterealtime.org/browse/SMACK-860) for the caching of disco#info elements. Until that is implemented, I'd like to move forward with this implementation.

This PR contains a feature and a bug fix (which you didn't pick up the last time, AFAICS):
- provide a MUC's `RoomInfo` to the application, after joining
- allow entering MUCs that are not hosted on a "typical" MUC domain, like chat@dino.im

It's hard (but not impossible) to separate those two things, as they are touching the same place in the code, and I have implemented them in the order I needed initially. Still, I'd like to reduce the divergence between my Smack fork and the master branch, so I'd love to see both applied.

Flowdalic:
> Until that is implemented, I'd like to move forward with this implementation.

I don't think that is is sensible to go for the not so good solution when we could go for the better solution. Especially if the better solution is not that hard to implement.

I will not pick up 
6d96ae1 because of my policy to not include workarounds for non-compliant implementations. Please ask the implementation vendor to ensure that all MUC services announce the fact that they are a MUC service by announcing the according feature.

I did not had a close look at the others. I'll try to do so once I find the time.

Please note that I would prefer separate pull requests for independent commits/changes in the future. That makes it easier to discuss and work on the individual changes.

ge0rg:
> I don't think that is is sensible to go for the not so good solution when we could go for the better solution.

Awesome! I'm sure you've used the five months that the issue has been open now to make a plan and prepare everything that's needed for the implementation, so all that's still left is pushing the better solution to the master branch.

> I will not pick up
6d96ae1 because of my policy to not include workarounds for non-compliant implementations.

I would agree with you if there was a place in XEP-0045 that a MUC *MUST* be hosted on a domain that advertises the MUC feature. However, I haven't seen such a requirement, and thus I consider that implementation as not violating the specification. Either way, it exists and locking out smack based implementations just isn't the right way.

Flowdalic:
Conflicts and a few commits of this PR have already been discussed.

PR #331, opened 2019-07-30
Locale: return a random body/subject if none matched language=null

ge0rg:
This is related to the time-out-closed thread on body language [0].

If no language is requested (language parameter is null), then return a
random subject / body from the respective Set. This is better than not
returning any body because of unexpected conditions.

However, this is just a workaround. The whole language code inside of
Message needs to be rewritten to properly do language tag matching
between the user's locale (which is not accessible from the Message
class), the stream language (same), and what's actually available in the
message.

Signed-off-by: Georg Lukas <georg@op-co.de>

[0] https://discourse.igniterealtime.org/t/message-getbody-returns-null-if-all-bodies-have-an-xml-lang/83923/7

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/Message.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/Message.java
@@ -201,6 +201,8 @@ public final class Message extends Stanza implements TypedCloneable<Message> {
                 return subject;
             }
         }
+        if (language == null && !subjects.isEmpty())
+            return subjects.iterator().next();
         return null;
     }
 
@@ -320,6 +322,8 @@ public final class Message extends Stanza implements TypedCloneable<Message> {
                 return body;
             }
         }
+        if (language == null && !getBodies().isEmpty())
+            return getBodies().iterator().next();
         return null;
     }
 


GitCop:

 * Commit: ede10ef0b54adbeb07bad82ea541d9a9e23746ec

---

ge0rg:
So the test fail because [getBodyLanguages()](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/packet/Message.java#L485-L494) and [getSubjectLanguages()](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/packet/Message.java#L319-L328) are removing the "default" element from the returned list based on the body/subject instead of based on the default language.

To be honest, I don't see how this can be changed meaningfully without a significant rewrite. The "easy" solution would be to expose a new function `getAnyBody()` but my intention of changing `getBody()` was to help legacy applications benefit.

Flowdalic:
I was thinking about removing that filtering of `getBodyLanguages()` and `getSubjectLanguages()` anyway, as I could not find a good reason why it is done in the first place. But that is ancient Smack code territory and I try to avoid breaking changes.
But assuming those methods would simply return all languages: would that help your cause?

ge0rg:
I *assume* the design decision behind it was that `getBody()` would return the "primary" body and `getBodyLanguages()` would allow you to iterate all the others.

Making it return *all* body languages will make the API consistent and the fix to the test case trivial, so yes, I'd like to go on with that.

Flowdalic:
> I _assume_ the design decision behind it was that `getBody()` would return the "primary" body and `getBodyLanguages()` would allow you to iterate all the others.

That would be my guess as well. But for various reasons I do not believe that such an API design is desirable.

> Making it return _all_ body languages will make the API consistent and the fix to the test case trivial, so yes, I'd like to go on with that.

Would you mind creating a PR (or updating this PR) with such a change?

Flowdalic:
I am to the conclusion that changing the behavior of existing methods is not desirable in that case and that instead, new methods should be added. And since this this PR also conflicts with the current master, I am closing this.

PR #333, opened 2019-08-02
DocFix in `NodeInformationProvider`

adiaholic:


smack-extensions/src/main/java/org/jivesoftware/smackx/disco/NodeInformationProvider.java
```diff
@@ -29,7 +29,6 @@
  * The NodeInformationProvider is responsible for providing supported identities, features
  * and hosted items (i.e. DiscoverItems.Item) about a given node. This information will be
  * requested each time this XMPPP client receives a disco info or items requests on the
```
Flowdalic: There is also an superfluous 'P' here. Could you remove it too?

adiaholic: Sure :)

PR #337, opened 2019-08-04
OMEMO: converse.js compatibility

vanitasvitae:
Fix some compatibility issues with converse.js and prevent NPE in provider class.

Converse was introducing linebreaks in the contents of the OMEMO bundle element, which would break base64 decoding.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/StringUtils.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/StringUtils.java
@@ -222,6 +222,10 @@ public class StringUtils {
         return out;
     }
 
+    public static String removeNewLines(String s) {
+        return s.trim().replaceAll("\n", "");
+    }
+
     /**
      * Hashes a String using the SHA-1 algorithm and returns the result as a
      * String of hexadecimal numbers. This method is synchronized to avoid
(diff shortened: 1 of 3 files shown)


Flowdalic:
Thanks. This makes we wonder if we shouldn't sanitize whitespace in general before passing the String to the Base64 decoder. Especially given XML Schema Part 2: Datatypes Second Edition § 3.2.16 says about `xs:base64Binary`, which is also used by XEP-0384:

> The lexical forms of base64Binary values are limited to the 65 characters of the Base64 Alphabet defined in [RFC 2045], i.e., a-z, A-Z, 0-9, the plus sign (+), the forward slash (/) and the equal sign (=), together with the characters defined in [XML 1.0 (Second Edition)] as white space. No other characters are allowed.
> 
> For compatibility with older mail gateways, [RFC 2045] suggests that base64 data should have lines limited to at most 76 characters in length. This line-length limitation is not mandated in the lexical forms of base64Binary data and must not be enforced by XML Schema processors.

Hence, right now, I believe we should go for the generic solution to this.

Also related: This commit also swaps the operands of a lot of String equals() methods. Is this related? What is the motivation in doing so? I can see an advantage of having the string constant, which is know to be not null, on the left side. Is there anything else to it?

vanitasvitae:
> This makes we wonder if we shouldn't sanitize whitespace in general before passing the String to the Base64 decoder.

You mean add the sanitizing to the Base64 util class itself?

> Also related: This commit also swaps the operands of a lot of String equals() methods. Is this related? What is the motivation in doing so? I can see an advantage of having the string constant, which is know to be not null, on the left side. Is there anything else to it?

That's basically it. I *think* you pointed this out in some other PR I did some time ago, so I noticed it right away :P

I don't have a stacktrace, as I haven't encountered this so far. So this change is probably more cosmetic.

Flowdalic:
Should be fixed with https://github.com/igniterealtime/Smack/commit/d7b7abc7ebd2efd04b9fcf835f97dbb45683e296

PR #342, opened 2019-08-28
Remove trailing semicolon in the header used in HttpFileUploadManager

fenek:
`HttpFileUploadManager` uses different content type for slot request and for the actual request to the upload URL. Extra semicolon results in signature error when min.io is used for the storage and quite probably doesn't work for Amazon S3 as well.

This PR fixes it.


vanitasvitae:
Hi @fenek!
Thanks for your contribution! It looks like the trailing semicolon can be omitted indeed. Good catch! :)

PR #343, opened 2019-08-31
Fix namespace contextualization in `MediaElement`

adiaholic:
This commit addresses #328 in respose to [review](https://github.com/igniterealtime/Smack/pull/328#issuecomment-523897472).

The intended XML is
`<uri type='image/jpeg'> http://www.shakespeare.lit/clients/exodus.jpg </uri>`
in place of
`<uri xmlns='urn:xmpp:media-element' type='image/jpeg'> http://www.shakespeare.lit/clients/exodus.jpg </uri>.`

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/mediaelement/element/MediaElement.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/mediaelement/element/MediaElement.java
@@ -87,6 +87,12 @@ public class MediaElement implements FormFieldChildElement {
             .optAttribute("width", width)
             .rightAngleBracket();
 
+        if (!NAMESPACE.contentEquals(xmlEnvironment.getNamespace())) {
+            xmlEnvironment = XmlEnvironment.builder()
+                    .withNamespace(MediaElement.NAMESPACE)
+                    .withNext(null)
+                    .build();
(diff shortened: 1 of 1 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/mediaelement/element/MediaElement.java
```diff
@@ -87,6 +87,12 @@ public XmlStringBuilder toXML(XmlEnvironment xmlEnvironment) {
             .optAttribute("width", width)
             .rightAngleBracket();
 
+        if (!xmlEnvironment.getNamespace().contentEquals(NAMESPACE)) {
```
vanitasvitae: Just to be sure that there are no NPEs, consider switching `getNamespace()` and `NAMESPACE`

adiaholic: Makes sense ;)

Flowdalic:
Thanks, but I believe I found a better fix with https://github.com/igniterealtime/Smack/commit/dd4df0a5ef2435f1c380a1c83c5b838b6986b2ce

Still appreciated that you pointed out the issue.

PR #344, opened 2019-08-31
XEP-0359: Fix filtering of messages applicable to SIDs.

vanitasvitae:
I noticed, that despite calling `sidManager.enable()`, the `StableAndUniqueStanzaIdManager` did not append any `<origin-id/>` elements to outgoing messages.

This patch fixes the filter logic, which was borked to begin with. The filter basically checked `A and not A` which never becomes true. I wonder how this happened :thinking:


vanitasvitae:
Btw: This fix comes courtesy to Mercury testing :yum:

Flowdalic:
Thanks :)

Any reason why we don't add origin-id to every outgoing message per default?

vanitasvitae:
@Flowdalic we actually already did (there was a static block which added a `ConnectionCreationListener` which got an instance of the `StableAndUniqueStanzaIdManager` for each new connection. On top of this, each manager enabled itself in its constructor.

I created #345 which makes this behaviour configurable, but also preserve it by default.

PR #348, opened 2019-09-27
IntegrationTest: allow configuration of hostname and ports for involved connections

vanitasvitae:
This PR adds additional properties for the IntegrationTest that allow specifying hostname and ports of involved test accounts.

The added properties are:

* `adminAccountHostname`: host name for the admin account
* `adminAccountPort`: port for the admin account
* `account{One,Two,Three}Hostname`: host to use for the n-th test account
* `account{One,Two,Three}Port`: port to use for the n-th test account

These options are only briefly tested (basically I made sure that connecting did *not* work when using invalid values) and I expect @guusdk to do the testing for me and to provide the pay check :P

### DIFF
--- a/documentation/developer/integrationtest.md
+++ b/documentation/developer/integrationtest.md
@@ -58,24 +58,34 @@ debugger=console
 
 ### Framework properties
 
-| Name                 |                                           |
-|----------------------|-------------------------------------------|
-| service              | XMPP service to run the tests on          |
+| Name                 |                                             |
+|----------------------|---------------------------------------------|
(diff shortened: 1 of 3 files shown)


Flowdalic:
I am confused why this is a per account settings and not per sinttest configuration (akin to e.g. 'service'), as I don't ever expect the hostname nor port, if set, to be different for a single configuration. Or am I missing some background / use-case here?

vanitasvitae:
The use case is testing of clustered services.
An example scenario would be to test whether pubsub items are properly propagated between the nodes of a cluster.

PR #349, opened 2019-10-10
Getgitcommit fix

jonathan-haubrich:
The assert on line 659 was causing my build to fail. Two issues caused gitCommit to be empty.

1. The cmd `'git describe --always --tags --dirty=+'` was not given enough time to complete and had not exited which meant no text in proc.text
2. The two git commands on lines 653 and 658 were run from the CWD of my Eclipse IDE, not the $projectDir which caused git to return an error 128.

To solve the two issues I added a `waitForOrKill` method call to proc (like the srCmd had) and I set the `execute` to run in $projectDir which I think was the intent/assumption in the original code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -650,13 +650,16 @@ def getGitCommit() {
 	def dotGit = new File("$projectDir/.git")
 	if (!dotGit.isDirectory()) return 'non-git build'
 
+	def projectDir = dotGit.getParentFile()
 	def cmd = 'git describe --always --tags --dirty=+'
-	def proc = cmd.execute()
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 685265ff482ff003f07bd67323150679c4131493

---

Flowdalic:
Cherry picked as https://github.com/igniterealtime/Smack/commit/7a5886279473c06d361d2903ee60f8b0eb0d13f9

At some point we may want to use https://github.com/ajoberstar/grgit for this.

PR #350, opened 2019-10-20
Make use of `pep` instead of `pubsub` in GeoLocations

adiaholic:
This mini assignment kicked-off with replacing `pubsub` with `pep`,
but later transformed into something more.

The alterations and additions in this commit:
- GeoLocation.

      1. Add Documentation.
      2. Add `EMPTY_GEO_LOCATION` to be used while `stopPublishingGeoLocation()` is called.
-  Add GeoLocation IntegrationTest.
-  Add GeoLocation Listener.
- GeoLocationManager.

      1. Add Documentation.
      2. Replace `pubsub` with `pep`.
      3. Add methods to add-and-remove GeoLocationListeners.
      4. Enable GeoLocation by default.
-  Add `package.info` for GeoLocation Integration Test.

What more can be done :
 - There is slight ambiguity in the package name (UserLocation vs GeoLocation). 
   If possible, I would prefer a unified nomenclature and change `GeoLocation` to `UserLocation`.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/geoloc/packet/GeoLocation.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/geoloc/packet/GeoLocation.java
@@ -48,6 +48,8 @@ public final class GeoLocation implements Serializable, ExtensionElement, FormFi
 
     public static final QName QNAME = new QName(NAMESPACE, ELEMENT);
 
+    public static final GeoLocation EMPTY_GEO_LOCATION = GeoLocation.builder().build();
+
     private static final Logger LOGGER = Logger.getLogger(GeoLocation.class.getName());
 
     private final Double accuracy;
@@ -264,6 +266,10 @@ public final class GeoLocation implements Serializable, ExtensionElement, FormFi
         return NAMESPACE;
     }
 
+    /**
+     * Returns a new instance of {@link Builder}.
+     * @return Builder
+     */
     public static Builder builder() {
         return new GeoLocation.Builder();
     }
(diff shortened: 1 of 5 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/geoloc/packet/GeoLocation.java
```diff
@@ -308,126 +335,255 @@ public static GeoLocation from(FormField formField) {
         private String tzo;
         private URI uri;
 
+        /**
+         * Sets accuracy.
+         * @param accuracy : Horizontal GPS error in meters.
```
vanitasvitae: For all those builder method documentations:
I'd probably put the detailed description of the parameter into the header of the documentation and replace the `@param` documentation with a short name.

Eg.
```
/**
 * Sets the accuracy of the GPS error in meters.
 * @param accuracy accuracy in meters
 * @return Builder
 */
```

smack-extensions/src/main/java/org/jivesoftware/smackx/geoloc/packet/GeoLocation.java
```diff
@@ -308,126 +335,255 @@ public static GeoLocation from(FormField formField) {
         private String tzo;
         private URI uri;
 
+        /**
+         * Sets accuracy of horizontal GPS error in meters.
+         * @param accuracy : accuracy accuracy in meters.
```
Flowdalic, line 340: Do not use the colon here. I doubt that it will render will. I would also prefer no period at the end.

Instead of
```@param accuracy : accuracy accuracy in meters.```
do
```@param accuracy the accuracy in meters```

When in doubt, look how the JRE is documented, .e.g http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/HashMap.java#l484

adiaholic:
@Flowdalic changes have been made in accordance with the review.

PR #351, opened 2019-10-20
Correct code example inside `UserTuneElement`.

adiaholic:


adiaholic:
> The correct pattern to format multi line code examples is `<pre>{@code}</pre>`. See https://stackoverflow.com/a/542142/194894
> 
> Please fix and rebase on current master. Thanks

@Flowdalic  Fixed and rebased 👍

PR #355, opened 2019-12-16
Message fastening

vanitasvitae:
This PR adds support for XEP-0422: Message Fastening version 0.1.1.

The manager class is rather empty, this is due to the nature of Message Fastening being a protocol intended to be used in combination with other protocols.
Therefore the focus of this PR lays on the elements and the provider.

The PR also contains a patch for OriginIdElement, which was lacking a proper equals() implementation.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
@@ -1,0 +1,107 @@
+/**
+ *
+ * Copyright 2019 Paul Schaub
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.message_fastening;
+
+import java.util.List;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.ConnectionCreationListener;
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPConnectionRegistry;
+import org.jivesoftware.smack.packet.MessageBuilder;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.message_fastening.element.FasteningElement;
+
+/**
+ * Smacks API for XEP-0422: Message Fastening.
+ * The API is still very bare bones, as the XEP intends Message Fastening to be used as a tool by other protocols.
+ *
+ * To enable / disable auto-announcing support for this feature, call {@link #setEnabledByDefault(boolean)} (default true).
+ *
+ * To fasten a payload to a previous message, create an {@link FasteningElement} using the builder provided by
+ * {@link FasteningElement#builder()}.
+ *
+ * You need to provide the {@link org.jivesoftware.smackx.sid.element.OriginIdElement} of the message you want to reference.
+ * Then add wrapped payloads using {@link FasteningElement.Builder#addWrappedPayloads(List)}
+ * and external payloads using {@link FasteningElement.Builder#addExternalPayloads(List)}.
+ *
+ * If you fastened some payloads onto the message previously and now want to replace the previous fastening, call
+ * {@link FasteningElement.Builder#isRemovingElement()}.
+ * Once you are finished, build the {@link FasteningElement} using {@link FasteningElement.Builder#build()} and add it to
+ * a stanza by calling {@link FasteningElement#applyTo(MessageBuilder)}.
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0422.html">XEP-0422: Message Fastening</a>
+ */
+public final class MessageFasteningManager extends Manager {
+
+    public static final String NAMESPACE = "urn:xmpp:fasten:0";
+
+    private static boolean ENABLED_BY_DEFAULT = true;
+
+    private static final WeakHashMap<XMPPConnection, MessageFasteningManager> INSTANCES = new WeakHashMap<>();
+
+    static {
+        XMPPConnectionRegistry.addConnectionCreationListener(new ConnectionCreationListener() {
+            @Override
+            public void connectionCreated(XMPPConnection connection) {
+                if (ENABLED_BY_DEFAULT) {
+                    MessageFasteningManager.getInstanceFor(connection).announceSupport();
+                }
+            }
+        });
+    }
+
+    private MessageFasteningManager(XMPPConnection connection) {
+        super(connection);
+    }
+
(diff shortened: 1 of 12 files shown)


Flowdalic:
Squash and needs JIRA assume (can you create one?)

vanitasvitae:
Done

smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/provider/FasteningElementProvider.java
```diff
@@ -0,0 +1,83 @@
...
+                    // Parse wrapped payload
+                    ExtensionElement wrappedPayload = PacketParserUtils.parseExtensionElement(name, namespace, parser, xmlEnvironment);
+                    builder.addWrappedPayload(wrappedPayload);
+                    break;
+
+                case END_ELEMENT:
+                    name = parser.getName();
+                    namespace = parser.getNamespace();
+
+                    if (FasteningElement.ELEMENT.equals(name) && MessageFasteningManager.NAMESPACE.equals(namespace)) {
```
Flowdalic, line 73: Strictly speaking, that is not robust, as it would also break the outerloop if there is a fastening element within a fastening element. I know, this is very unlikely, but the correct idiom is to break if the depth equals the initial depth, which is given as argument to the parse method.

In order to avoid other patters being used in places where it potentially matters me, we should simply use this one pattern everywhere.

smack-experimental/src/test/java/org/jivesoftware/smackx/message_fastening/MessageFasteningElementsTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Test XML serialization of the {@link FasteningElement} using the example provided by
+     * the XEP.
+     *
+     * @see <a href="https://xmpp.org/extensions/xep-0422.html#wrapped-payloads">XEP-0422 §3.1 Wrapped Payloads</a>
+     */
+    @Test
+    public void fasteningElementSerializationTest() {
+        String xml = "" +
```
Flowdalic, line 56: nit: superfluous `""` here.

Flowdalic: And in further places in this file.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
```diff
@@ -0,0 +1,100 @@
...
+            throw new IllegalArgumentException("Stanza cannot contain more than one apply-to elements.");
+        } else {
+            stanza.addExtension(applyTo);
+        }
+    }
+
+    /**
+     * Announce support for Message Fastening via Service Discovery.
+     */
+    public void announceSupport() {
```
Flowdalic, line 96: Shouldn't there also be a method `stopAnnoucingSupport()`? And if just for symmetry reasons.

smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Add the provided Message Fastening apply-to element to the provided stanza.
+     * Note: The stanza MUST NOT contain more than one apply-to elements at the same time.
+     *
+     * @see <a href="https://xmpp.org/extensions/xep-0422.html#rules">XEP-0422 §4: Business Rules</a>
+     *
+     * @param applyTo message fastening apply-to element
+     * @param stanza stanza
+     */
+    public void addFasteningElementToStanza(FasteningElement applyTo, Stanza stanza) {
```
Flowdalic, line 85: That method can be static. And I also think it potentially belongs into `FasteningElement`, since it is not really a Manager type of operation that it performs. It is more like a "modify stanza operation". That said, the type of the second parameter should be a builder. And I wonder if we should restrict it to message type stanzas, i.e. `MessageBuilder`.

Flowdalic: So maybe add a non static method `FasteningElement.addTo(MessageBuilder)`?

smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public static final String NAMESPACE = "urn:xmpp:fasten:0";
+
+    private static final WeakHashMap<XMPPConnection, MessageFasteningManager> INSTANCES = new WeakHashMap<>();
+
+    static {
+        XMPPConnectionRegistry.addConnectionCreationListener(new ConnectionCreationListener() {
+            @Override
+            public void connectionCreated(XMPPConnection connection) {
+                MessageFasteningManager.getInstanceFor(connection).announceSupport();
```
Flowdalic, line 58: That code is currently not necessary, no? I don't really mind having it, but I would remove it.

Flowdalic: Or alternatively, add a "enabled by default" feature. Then this code is necessary. But then `MessageFasteningManager` must be recorded as auto loaded class by smacks initialization routine.

smack-experimental/src/test/java/org/jivesoftware/smackx/message_fastening/MessageFasteningElementsTest.java
vanitasvitae: This is on purpose so that the XML is indented correctly and therefore readable :)

Flowdalic: I think if the newline would directly follow after `xml =`, the XML would be indented as well and the readability would even be increased (as there would be less noise).

vanitasvitae: Not with my IDE. It would be auto indented to
```
        String xml =
                "<apply-to xmlns='urn:xmpp:fasten:0' id='origin-id-1' clear='true'>" +
                        "      <i-like-this xmlns='urn:example:like'>Very much</i-like-this>" +
                        "</apply-to>";
```

Flowdalic: I am sure your IDE allows to change the layout after it has applied auto indention. Or simply `sed s; "" +$;; MessageFasteningElementsTest.java`

smack-experimental/src/main/java/org/jivesoftware/smackx/message_fastening/MessageFasteningManager.java
```diff
@@ -0,0 +1,107 @@
...
+ * Once you are finished, build the {@link FasteningElement} using {@link FasteningElement.Builder#build()} and add it to
+ * a stanza by calling {@link FasteningElement#applyTo(MessageBuilder)}.
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0422.html">XEP-0422: Message Fastening</a>
+ */
+public final class MessageFasteningManager extends Manager {
+
+    public static final String NAMESPACE = "urn:xmpp:fasten:0";
+
+    private static boolean ENABLED_BY_DEFAULT = true;
```
Flowdalic, line 54: Wouldn't it be sensible to default to 'false'? After all, if it is announced as feature, entities sending fastened messages expect the recipient to react somehow on those. And this is not the case if this is just enabled in Smack.

PR #356, opened 2020-01-06
Improve attention

vanitasvitae:
This is a small change that makes use of `XmlStringBuilder` instead of handcrafting XML in `AttentionElement.toXml()`.
I also added a `AttentionElementTest` test case that compares the resulting XML with the XML from the XEP.


Flowdalic:
Could you squash the last commit in the previous one?

vanitasvitae:
Done

PR #357, opened 2020-01-06
Add Nick JUnit test

vanitasvitae:
This PR adds a JUnit test for the Nick element and provider classes of Smacks implementation of XEP-0172: User Nickname.

### DIFF
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/nick/NickTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/nick/NickTest.java
@@ -1,0 +1,83 @@
+/**
+ *
+ * Copyright 2020 Paul Schaub
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


smack-extensions/src/test/java/org/jivesoftware/smackx/nick/NickTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import org.junit.Test;
+
+public class NickTest {
+    /**
+     * see <a href="https://xmpp.org/extensions/xep-0172.html#example-3">XEP-0172: User Nickname - Example 3</a>
+     */
+    private static final String XML = "<nick xmlns='http://jabber.org/protocol/nick'>Ishmael</nick>";
+
+    @Test(expected = IllegalArgumentException.class)
```
Flowdalic, line 39: Please use JUnit 5

smack-extensions/src/test/java/org/jivesoftware/smackx/nick/NickTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test
+    public void serializationTest() {
+        Nick nick = new Nick("Ishmael");
+
+        assertXmlSimilar(XML, nick.toXML());
+    }
+
+    @Test
+    public void deserializationTest() throws XmlPullParserException, IOException, SmackParsingException {
+        Nick nick = new NickProvider().parse(TestUtils.getParser(XML));
```
Flowdalic, line 59: This should use Smack's new idiom to test parsers, which will run the unit test for both stax and xpp3. See for example `CapsExtensionProviderTest`.

Flowdalic:
I note that this still uses the juni4 @Test annotation. Will merge anyways.

PR #361, opened 2020-01-28
Correct ALPHABET in `BASE32.java`

adiaholic:
As per [RFC 4648](https://tools.ietf.org/html/rfc4648#section-6) Table 3, no `Value` gives its encoding as `8`.


vanitasvitae:
Wow, sharp eye! Nice catch :)

PR #363, opened 2020-01-30
Disable SSL 3. Enable TLS 1.3. Deprecate TLS 1.0, TLS 1.1.

MilanKral:
In 2014, SSL 3.0 was found to be vulnerable to the POODLE attack.
SSL 3.0 was deprecated in June 2015 by RFC 7568.

According to RFC 8600 messages MUST be carried over TLS (minimally
TLS 1.2 and preferably TLS 1.3). According to stats from XMPP.net
over 97% of public XMPP servers support TLS 1.2.

In October 2018, Apple, Google, Microsoft, and Mozilla jointly
announced they would deprecate TLS 1.0 and 1.1 in March 2020.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/util/TLSUtils.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/util/TLSUtils.java
@@ -42,21 +42,16 @@ import org.jivesoftware.smack.SmackException.SecurityNotPossibleException;
 
 public class TLSUtils {
 
-    public static final String SSL = "SSL";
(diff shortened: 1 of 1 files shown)


Flowdalic:
Thanks for your contribution. I come up  with https://github.com/igniterealtime/Smack/commit/8b0de5ff85ceb52cfc71d9ac95ff2a211f7e2873

BTW, the referenced RFC 8600 only applies to XMPP Grid, not XMPP as a whole. But the recommendation within that RFC is still valid.

PR #364, opened 2020-02-08
Initial support for XEP-0249: Direct MUC Invitations

vanitasvitae:
This PR adds support for [XEP-0249](https://xmpp.org/extensions/xep-0249.html).

Future work would include unifying the API of this PR with the [existing legacy MUC invitation API](https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/MUCUser.java#L249).

Having support for XEP-0249 is required for advanced IM clients according to [XEP-0423: XMPP Compliance Suites 2020](https://xmpp.org/extensions/xep-0423.html#im).

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DirectMucInvitationManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DirectMucInvitationManager.java
@@ -1,0 +1,115 @@
+/**
+ *
+ * Copyright 2020 Paul Schaub.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.muc;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.SmackException;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPConnectionRegistry;
+import org.jivesoftware.smack.XMPPException;
+import org.jivesoftware.smack.filter.StanzaExtensionFilter;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.packet.MessageBuilder;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.muc.packet.GroupChatInvitation;
+
+import org.jxmpp.jid.EntityBareJid;
+
+/**
+ * Smacks API for XEP-0249: Direct MUC Invitations.
+ * Use this instead of {@link org.jivesoftware.smackx.muc.packet.MUCUser.Invite}.
+ *
+ * To invite a user to a group chat, use {@link #inviteToMuc(MultiUserChat, EntityBareJid)}.
+ *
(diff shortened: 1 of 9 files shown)


vanitasvitae:
I noticed, that support for this is already (partially?) implemented in https://github.com/igniterealtime/Smack/blob/master/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java

The implementation appears to be incomplete though.

vanitasvitae:
I now incorporated my implementation of XEP-0249 into the already existing (non-standard?) implementation of Smack's MUC Invitation code.

I tried to keep the code as backwards compatible as possible.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DirectMucInvitationManager.java
```diff
@@ -0,0 +1,113 @@
...
+        if (manager == null) {
+            manager = new DirectMucInvitationManager(connection);
+            INSTANCES.put(connection, manager);
+        }
+        return manager;
+    }
+
+    private DirectMucInvitationManager(XMPPConnection connection) {
+        super(connection);
+        ServiceDiscoveryManager.getInstanceFor(connection).addFeature(GroupChatInvitation.NAMESPACE);
```
Flowdalic, line 68: I think `ServiceDiscoveryManager` should be a field of this class.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/DirectMucInvitationManager.java
```diff
@@ -0,0 +1,113 @@
...
+
+        registerExtensionElementListener();
```
Flowdalic, line 70: For soundness, I'd switch this line with the feature announcement. That is, first register the listeners and then announce the feature.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -81,22 +71,85 @@
...
+        this._continue = _continue;
+        this.thread = thread;
     }
 
     /**
      * Returns the address of the group chat room. GroupChat room addresses
      * are in the form <code>room@service</code>, where <code>service</code> is
      * the name of group chat server, such as <code>chat.example.com</code>.
      *
+     * @deprecated use {@link #getRoomAddressJid()} instead.
```
Flowdalic, line 98: If we deprecate a method, then please add also a "TODO: Remove in Smack X.Y" comment, so that the method is eventually removed.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -108,11 +161,46 @@ public String getNamespace() {
...
+                .attribute(ATTR_JID, getRoomAddressJid())
+                .optAttribute(ATTR_PASSWORD, getPassword())
+                .optAttribute(ATTR_REASON, getReason())
+                .optAttribute(ATTR_THREAD, getThread())
+                .closeEmptyElement();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
```
Flowdalic, line 176: Could use `o.j.s.util.EqualsUtil` here.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/provider/GroupChatInvitationProvider.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jivesoftware.smackx.muc.packet.GroupChatInvitation;
+
+public class GroupChatInvitationProvider extends ExtensionElementProvider<GroupChatInvitation> {
+
+    public static final GroupChatInvitationProvider TEST_PROVIDER = new GroupChatInvitationProvider();
+
+    @Override
+    public GroupChatInvitation parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment)
+            throws XmlPullParserException, IOException, SmackParsingException {
+        return new GroupChatInvitation(
```
Flowdalic, line 42: I don't like the density here. I would prefer if all arguments of the constructor are declared as local variables.

samwatt6:
Hi there, our app uses Smack and we'd really benefit from this change. Is there any chance it could be merged? Thanks

PR #365, opened 2020-02-16
OMEMO: Switch to sending 12 byte IV

vanitasvitae:
This PR fixes https://github.com/xsf/xeps/pull/894 for smack-omemo.

A while ago, the community decided to switch from 16 bytes IV to 12 bytes, as that is what most AES-GCM libraries use internally anyways. Using 16 bytes as IV would not result in improved security, as the IV is getting hashed down to 12 bytes anyways.

For quite a while now, most implementations of OMEMO supported receiving both 12 and 16 byte IV, but kept sending 16 for a grace period. Now is the time to make the switch to sending 12.

### DIFF
--- a/smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/util/OmemoMessageBuilder.java
+++ b/smack-omemo/src/main/java/org/jivesoftware/smackx/omemo/util/OmemoMessageBuilder.java
@@ -274,13 +274,13 @@ public class OmemoMessageBuilder<T_IdKeyPair, T_IdKey, T_PreKey, T_SigPreKey, T_
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Neustradamus:
@vanitasvitae: Thanks!

Linked to:
- https://discourse.igniterealtime.org/t/a-correct-support-of-omemo-12-byte-for-initilization-vector-iv/87251

Neustradamus:
@vanitasvitae: Conversations has switched: https://github.com/siacs/Conversations/commit/d6ae9d8d14cca2afb0ddc7a66c68bcb89a64ab3b

Other softwares too.

Neustradamus:
@Flowdalic: Thanks!

@vanitasvitae: Thanks again for your PR.

PR #367, opened 2020-02-17
Allow for offline configuration of CarbonManager

vanitasvitae:
As discussed earlier, it would make sense to allow configuration of the CarbonManager while the connection is not yet established.

This PR adds such functionality by introducing the variable `shouldBeEnabled`, which is checked in the managers `ConnectionListener.onAuthenticated()` block. If true, the manager will automatically try to enable carbons for the connection. Exceptions are logged and the user is still able to check if carbons are successfully enabled using the old `isEnabled()` methods.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/carbons/CarbonManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/carbons/CarbonManager.java
@@ -1,6 +1,6 @@
 /**
  *
- * Copyright 2013-2014 Georg Lukas, 2017-2020 Florian Schmaus
+ * Copyright 2013-2014 Georg Lukas, 2017-2020 Florian Schmaus, 2020 Paul Schaub
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import org.jivesoftware.smack.AsyncButOrdered;
 import org.jivesoftware.smack.ConnectionCreationListener;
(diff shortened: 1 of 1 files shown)


vanitasvitae:
I tried to make the changes as less intrusive as possible. Someone who doesn't follow Smacks development should not notice any differences in behaviour, apart from that they only have to enable Carbons once ;)

smack-experimental/src/main/java/org/jivesoftware/smackx/carbons/CarbonManager.java
```diff
@@ -99,19 +105,30 @@ public void connectionCreated(XMPPConnection connection) {
     private final Set<CarbonCopyReceivedListener> listeners = new CopyOnWriteArraySet<>();
 
     private volatile boolean enabled_state = false;
+    private volatile boolean shouldBeEnabled = SHOULD_BE_ENABLED_DEFAULT;
```
Flowdalic, line 108: Can we drop the "should be" everywhere? I'd prefer "enabled by default".

smack-experimental/src/main/java/org/jivesoftware/smackx/carbons/CarbonManager.java
```diff
@@ -239,12 +263,10 @@ public boolean isSupportedByServer() throws NoResponseException, XMPPErrorExcept
      * You should first check for support using isSupportedByServer().
      *
      * @param new_state whether carbons should be enabled or disabled
-     * @throws NotConnectedException if the XMPP connection is not connected.
-     * @throws InterruptedException if the calling thread was interrupted.
      * @deprecated use {@link #enableCarbonsAsync(ExceptionCallback)} or {@link #disableCarbonsAsync(ExceptionCallback)} instead.
      */
     @Deprecated
-    public void sendCarbonsEnabled(final boolean new_state) throws NotConnectedException, InterruptedException {
+    public void sendCarbonsEnabled(final boolean new_state) {
```
Flowdalic, line 269: nice catch!

Flowdalic:
Squash into one commit please.

vanitasvitae:
Rebase in progress :)

PR #375, opened 2020-04-06
Replace JUnit4 with JUnit5 in `smack-extensions`.

adiaholic:
This PR concludes migration of `smack-extensions` module, from JUnit4 to JUnit5.
`mockito2` and `jUnit-vintage-engine` are required by both JUnit4 and JUnit5, hence remove them as `JUnit4` specific dependencies and provide them in project testing dependencies.

### DIFF
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/pubsub/ConfigureFormTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/pubsub/ConfigureFormTest.java
@@ -16,8 +16,9 @@
  */
 package org.jivesoftware.smackx.pubsub;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import java.io.IOException;
 
@@ -38,7 +39,7 @@ import org.jivesoftware.smackx.disco.packet.DiscoverInfoBuilder;
 import org.jivesoftware.smackx.pubsub.packet.PubSub;
 import org.jivesoftware.smackx.xdata.packet.DataForm;
 
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 
 /**
  * Configure form test.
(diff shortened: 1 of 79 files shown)


smack-extensions/src/test/java/org/jivesoftware/smackx/pubsub/ConfigureFormTest.java
```diff
@@ -95,12 +96,14 @@ public void getConfigFormWithTimeout() throws XMPPException, SmackException, Int
         DiscoverInfo discoverInfo = info.build();
         con.addIQReply(discoverInfo);
 
-        Node node = mgr.getNode("princely_musings");
+        assertThrows(Exception.class, () -> {
```
Flowdalic, line 99: Should be `SmackException` that is expected to be thrown here. And that by `getNodeConfiguration()` I'd assume, hence the lambda should be smaller.

adiaholic: Thanks for pointing that out. However, `getNode("pricely_musings");` has the potential to throw a `NotAPubSubNodeException` and hence I am keeping it inside the lambda. If this is not the case, please let me know and I'll make changes accordingly.

Flowdalic: Testing for expected exceptions is not about which exceptions could potentially be thrown, but which are actually thrown. So you test that, what is expected, actually happens, everything unforeseen happening is a violation of the test spec and should make the test fail.

adiaholic: Got it !

adiaholic: @Flowdalic 
Here, I think in master, the expected `SmackException` is raised by `mgr.getNode("princely_musings");` and not by `node.getNodeConfiguration();`.
This leaves the following code snippet unreachable and smoothly pass the test.
    `SmackConfiguration.setDefaultReplyTimeout(100);
     con.setTimeout();
     node.getNodeConfiguration();`.

adiaholic: A `TODO` has been added.

PR #377, opened 2020-04-11
Bouncy Castle 1.65 instead of 1.64

Neustradamus:
Bouncy Castle 1.65 instead of 1.64


Flowdalic:
Closed with https://github.com/igniterealtime/Smack/commit/321c91c264887eabda10a63a66906c1a637bd8b3

PR #378, opened 2020-04-11
Bump pgpainless to 0.0.1-alpha11 and test-depend on bcprov

vanitasvitae:
smack-openpgp is now transitively depending on bcpg through pgpainless 0.0.1-alpha9 (alpha11).
That means that only bcprov needs to be declared as a dependency in smack-openpgp's build file.

### DIFF
--- a/smack-openpgp/build.gradle
+++ b/smack-openpgp/build.gradle
@@ -8,7 +8,9 @@ dependencies {
 	compile project(':smack-extensions')
 	compile project(':smack-experimental')
 
-	compile 'org.pgpainless:pgpainless-core:0.0.1-alpha7'
+	api 'org.pgpainless:pgpainless-core:0.0.1-alpha11'
+
+	testImplementation "org.bouncycastle:bcprov-jdk15on:1.65"
 
 	testCompile project(path: ":smack-core", configuration: "testRuntime")
 	testCompile group: 'commons-io', name: 'commons-io', version: "$commonsIoVersion"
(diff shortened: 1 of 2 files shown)


smack-openpgp/build.gradle
```diff
@@ -8,7 +8,9 @@ dependencies {
 	compile project(':smack-extensions')
 	compile project(':smack-experimental')
 
-	compile 'org.pgpainless:pgpainless-core:0.0.1-alpha7'
+	api 'org.pgpainless:pgpainless-core:0.0.1-alpha9'
+
+	implementation "org.bouncycastle:bcprov-jdk15on:1.65"
```
Flowdalic, line 13: I'd like to keep bc out of the dependencies (if possible). I assume this is required for the unit tests? If so s/implementation/testImplementation/

vanitasvitae: Done :)

vanitasvitae:
Hum, I wonder why the build fails. The classes not found are part of bcpg, which is now [an API dependency of pgpainless](https://github.com/pgpainless/pgpainless/blob/master/pgpainless-core/build.gradle#L12), so it should be [transitively pulled in](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation) if I understand correctly.
It also worked on my local machine.

Flowdalic:
That doesn't seem to had any effect. Did you use an old gradle version to upload the artifacts?

https://search.maven.org/artifact/org.pgpainless/pgpainless-core/0.0.1-alpha9/jar
shows no dependency on bcpg-jdk15on, and so does `gradle -q smack-openpgp:dependencies`.

Flowdalic:
Should be fixed with https://github.com/pgpainless/pgpainless/pull/10

vanitasvitae:
This one should be good to go now :)

PR #388, opened 2020-04-28
Remove `TODO: method should not throw any exception but currently does` from `ConfigureFormTest`

adiaholic:
`mgr.getNode(...)` method is not supposed to throw any type of exception. It threw two kinds of exceptions: `NotConnectedException` and `SmackException`. By using `ThreadedDummyConnection.newInstance()` method, `NotConnectedException` can be removed. And `SmackException` can be removed by correctly building the DiscoverInfo packet.


Flowdalic:
Included as https://github.com/igniterealtime/Smack/commit/afa42a8b4105a969ef1c40df0009d510510ec93a

PR #389, opened 2020-05-12
pep: notification type can be normal or headline

guusdk:


vanitasvitae:
Can you link to where this is documented? A quick search yielded no results.

guusdk:
This is not explicitly documented as far as I can tell. XEP-0060 does only define a configurable option ('pubsub#notification_type'). The form lists two types ('normal' and 'headline'). The XEP doesn't use any type in most of the examples. XEP-0136 uses type 'headline' in its examples, but does not define any rules either. What I deduced is that only expecting 'headline' is to strict, hence this PR that at least adds 'normal'.

PR #392, opened 2020-05-13
Handle redundant API: getField(..)/read(..) inside FilledForm

adiaholic:


GitCop:

 * Commit: 0f4b582ae13c6808eb618100c7eeb00943ff54ed

---

vanitasvitae:
I'm not sure if this is desired.
The `read()` method comes from the `FormReader` class and is probably not really part of the API that `FilledForm` tries to provide. Removing the `getField()` method would make the API seem incomplete as there is `hasField()`, `getFieldOrThrow()`, but no more `getField()`. Having to use `read()` would break the "field-pattern" of the API.

Flowdalic:
I am torn on this too, but I was about to merge this. On the other hand, it was me who introduced getField() because I assumed that a method with that functionality was missing.

Another option would be to rename `read()` in `FormFieldReader` to `getField()`, but this method was named `read` because of the corresponding `write` in `FormFieldWriter`…. wait I just noticed something.

Flowdalic:
Closed with https://github.com/igniterealtime/Smack/commit/c5212c20b610e64eab125990c9fa67ee65e79864

PR #396, opened 2020-06-12
fix: Minimizes the number of disco-info on join room.

damencho:
Moves the known muc services to common place to avoid disco-info on every created MultiUserChat.

I cannot run gradle check because of having the same filename twice (ignorecase).


Flowdalic:
Thanks, applied with minor modifications as https://github.com/igniterealtime/Smack/commit/8e337d7810bbfb5d4a4bd1b5a68e471f865a4ea1

damencho:
Thanks.

PR #401, opened 2020-06-18
Add support for HTTP lookup method through XEP-0156

adiaholic:


Flowdalic:
Please also provide unit tests for this.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+         */
+        WEBSOCKET("urn:xmpp:alt-connections:websocket");
+        private final String attribute;
+        RelAttribute(String relAttribute) {
+            this.attribute = relAttribute;
+        }
+    }
+
+    private final int content_length;
+    private final HttpURLConnection httpConnection;
```
vanitasvitae, line 81: Why is the connection a member of the class?

vanitasvitae: Also, why is it final?

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+    String obtainedXml = "";
+
+    /**
+     * Constructs a HTTP connection with the server and retrieves complete host metadata.
+     *
+     * @param host host that needs to be connected
+     * @param usingHttps true if secure HTTP being used, false otherwise
+     * @throws IOException exception due to input/output operations
+     */
+    public HttpLookUp(String host, boolean usingHttps) throws IOException {
```
vanitasvitae, line 92: Don't throw exceptions in the constructor.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+        final String metadataUrl = (usingHttps ? "https://" : "http://") + host + "/.well-known/host-meta";
+        httpConnection = getHttpConnection(metadataUrl);
+        content_length = httpConnection.getContentLength();
+
+        InputStream inputStream = httpConnection.getInputStream();
```
vanitasvitae, line 97: Where do you close the connection+stream?

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+        int remainingBytes = content_length;
+
+        // Keep reading from the input stream until read completely
+        while (remainingBytes > 0) {
+            byte[] byteBuffer = new byte[BUFFER_SIZE];
+            int bytes_read = inputStream.read(byteBuffer);
+            obtainedXml += new String(byteBuffer, 0, bytes_read, Charset.defaultCharset());
```
vanitasvitae, line 104: Don't use string concatenation in a loop.
I'd either read the inputstream into a `ByteArrayOutputStream` and convert that to a String when I'm done with reading, or at least use a StringBuilder.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+            remainingBytes = remainingBytes - bytes_read;
+        }
+    }
+
+    private static HttpURLConnection getHttpConnection(String endpoint) throws IOException {
```
vanitasvitae, line 109: You are talking of an endpoint as the String argument here, but the whole point of this class is to retrieve endpoints, so the naming of the argument is not ideal :P

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Get remote endpoints from the web host metadata fetched from the server.
+     *
+     * @param relAttribute relation attribute for the type of connection
+     * @return List of {@link RemoteConnectionEndpoint}s
+     * @throws IOException exception due to input/output operations
+     * @throws XmlPullParserException exception encountered during XML parsing
+     * @throws URISyntaxException exception to indicate that a string could not be parsed as a URI reference
+     */
+    public List<RemoteConnectionEndpoint> getEndpoints(String relAttribute)
```
vanitasvitae, line 142: Can you see what might be non-optimal and will happen if the user does the following:
```
if (httpLookUp.getEndpoints("urn:xmpp:alt-connections:websocket").size() > 0) {
  for (RemoteConnectionEndpoint e : httpLookUp.getEndpoints("urn:xmpp:alt-connections:websocket")) {
    ...
  }
}
```

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: same for content_length

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: An idea: You might also think about removing the usingHttps argument and instead let your implementation optimistically try https first and then fall back to http?

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: Not closing those will cause you to leak memory.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: Also, while I like that this method is generic enough to enable use of other, currently unknown endpoint namespaces, I would also like to see some explicitly named methods like `getWebsocketEndpoints()`.

vanitasvitae: While `relAttribute` is probably the correct technical term, you may consider naming the argument something else, eg. `protocolNamespace` so that it is immediatelly clear what the method does.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: An idea: You might also think about removing the usingHttps argument and instead let your implementation optimistically try https first and then fall back to http?

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: Not closing those will cause you to leak memory.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
vanitasvitae: Also, while I like that this method is generic enough to enable use of other, currently unknown endpoint namespaces, I would also like to see some explicitly named methods like `getWebsocketEndpoints()`.
While `relAttribute` is probably the correct technical term, you may consider naming the argument something else, eg. `protocolNamespace` so that it is immediately clear what the method does.

vanitasvitae: A [hint](https://github.com/igniterealtime/Smack/blob/master/smack-core/src/main/java/org/jivesoftware/smack/util/MultiMap.java) for what you might use when parsing out the endpoints :)

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+import org.jivesoftware.smack.util.rce.RemoteConnectionEndpoint;
+import org.jivesoftware.smack.xml.XmlPullParser;
+import org.jivesoftware.smack.xml.XmlPullParserException;
+
+import org.minidns.util.InetAddressUtil;
+
+/**
+ * The HTTP lookup method uses web host metadata to list the URIs of alternative connection methods.<br>
+ * <br>
+ * To obtain endpoints for Bosh or Websocket connections, use {@link RelAttribute#BOSH} and <br>
```
Flowdalic, line 43: Way to many `<br/>`'s

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
Flowdalic: As discussed, should only perform https lookups.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+ * {@code
+ * HttpLookUp lookup = new HttpLookUp(host,usingHttps);
+ * List<RemoteConnectionEndpoint> endpoints = lookup.getEndpoints(RelAttribute.WEBSOCKET);
+ * }
+ * </pre>
+ * @see <a href="https://xmpp.org/extensions/xep-0156.html#http">
+*     HTTP Lookup Method.
+*     </a>
+ */
+public final class HttpLookUp {
```
Flowdalic, line 57: This is a xep156 implementation. As such, it should go into the `org.jivesoftware.smack.altconnections` package. The class name should be `HttpLookupMethod`. I'd design it with the following static methods:

```
InputStream getXrdStream(DomainBareJid xmppServiceAddress)
List<URI> parseXrdLinkReferencesFor(XmlPullParser xmlPullParser, String relation)
List<URI> lookup(DomainBareJid xmppServiceAddress, String relation)
List<URI> lookup(DomainBareJid xmppServiceAddress, LinkRelation relation)
```

where `LinkRelation` is basically your existing `RelAttribute`.

There is no point in using a constructor (yet, that may change). Nor should this API be based on RemoteConnectionEndpoints IMHO. That is to be done by the users of this API.

Flowdalic: And, of course, there should be a unit test for `parseXrdLinkReferencesFor()`. The other methods are not easily unit testable, as they perform network I/O.

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
```diff
@@ -0,0 +1,197 @@
...
+            obtainedXml += new String(byteBuffer, 0, bytes_read, Charset.defaultCharset());
+            remainingBytes = remainingBytes - bytes_read;
+        }
+    }
+
+    private static HttpURLConnection getHttpConnection(String endpoint) throws IOException {
+        final URL putUrl = new URL(endpoint);
+        HttpURLConnection httpURLConnection = (HttpURLConnection) putUrl.openConnection();
+        httpURLConnection.setRequestMethod("GET");
+        httpURLConnection.setUseCaches(false);
```
Flowdalic, line 113: why?

adiaholic: Oops!

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
adiaholic: > Why is the connection a member of the class?

This field needs to be accessed by another method at the time of closing connection. And hence it's a member of the class :)

smack-core/src/main/java/org/jivesoftware/smack/remoteEndpoint/HttpLookUp.java
adiaholic: > Can you see what might be non-optimal and will happen if the user does the following:
> 
> ```
> if (httpLookUp.getEndpoints("urn:xmpp:alt-connections:websocket").size() > 0) {
>   for (RemoteConnectionEndpoint e : httpLookUp.getEndpoints("urn:xmpp:alt-connections:websocket")) {
>     ...
>   }
> }
> ```

We end up parsing the same obtainedXml twice :)

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
+        final URL putUrl = new URL(metadataUrl);
+        httpURLConnection = (HttpURLConnection) putUrl.openConnection();
+        httpURLConnection.setRequestMethod("GET");
+        httpURLConnection.setUseCaches(true);
+        httpURLConnection.setDoOutput(false);
+        httpURLConnection.setRequestProperty("Content-Type", "xml");
```
Flowdalic, line 98: Is that necessary?

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
+        final URL putUrl = new URL(metadataUrl);
+        httpURLConnection = (HttpURLConnection) putUrl.openConnection();
+        httpURLConnection.setRequestMethod("GET");
+        httpURLConnection.setUseCaches(true);
+        httpURLConnection.setDoOutput(false);
```
Flowdalic, line 97: This is already the default value, no need to explicitly set it.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     * @param xmppServiceAddress address of host
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
+        final URL putUrl = new URL(metadataUrl);
+        httpURLConnection = (HttpURLConnection) putUrl.openConnection();
+        httpURLConnection.setRequestMethod("GET");
+        httpURLConnection.setUseCaches(true);
```
Flowdalic, line 96: That is already the default, no need to explicitly set it.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * @param xmppServiceAddress address of host
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
+        final URL putUrl = new URL(metadataUrl);
+        httpURLConnection = (HttpURLConnection) putUrl.openConnection();
+        httpURLConnection.setRequestMethod("GET");
```
Flowdalic, line 95: That is already the default, no need to explicitly set it.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+ * @see <a href="https://xmpp.org/extensions/xep-0156.html#http">
+ *     HTTP Lookup Method from XEP-0156.
+ *     </a>
+ */
+public final class HttpLookupMethod {
+
+    private static final int BUFFER_SIZE = 256;
+    private static final String XRD_NAMESPACE = "http://docs.oasis-open.org/ns/xri/xrd-1.0";
+
+    private static int contentLength;
```
Flowdalic, line 61: All static fields should be method local variables now.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     * and retrieves XRD element in the form of an InputStream.
+     *
+     * @param xmppServiceAddress address of host
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
+        final URL putUrl = new URL(metadataUrl);
+        httpURLConnection = (HttpURLConnection) putUrl.openConnection();
```
Flowdalic, line 94: Do you need to cast it to `HttpURLConnection`? Can't it just stay an `URLConnection`?

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+     *
+     * @param xmppServiceAddress address of host
+     * @param relation LinkRelation as a string specifying type of connection
+     * @return list of endpoints
+     * @throws IOException exception due to input/output operations
+     * @throws XmlPullParserException exception encountered during XML parsing
+     * @throws URISyntaxException exception to indicate that a string could not be parsed as a URI reference
+     */
+    public static List<URI> lookup(DomainBareJid xmppServiceAddress, String relation) throws IOException, XmlPullParserException, URISyntaxException {
+        // Check if we have this xmppServiceAddress cached for the given link relation.
```
Flowdalic, line 169: Should open a new url connection.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,218 @@
...
+ *     </a>
+ */
+public final class HttpLookupMethod {
+
+    private static final int BUFFER_SIZE = 256;
+    private static final String XRD_NAMESPACE = "http://docs.oasis-open.org/ns/xri/xrd-1.0";
+
+    private static int contentLength;
+    private static HttpURLConnection httpURLConnection;
+    private static MultiMap<DomainBareJid, MultiMap<String, URI>> CACHE = new MultiMap<>();
```
Flowdalic, line 63: Caches are never simple maps. As this would allow them to grow indefinitely. I am still not sure if we really need a cache, as I'd epxect that the HTTP library will do caching for you. I'd say we remove the cache for now and re-introduce it if we see a need for it.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
adiaholic: Changed `HttpURLConnection` to its superclass `URLConnection` :)

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
adiaholic: After changing static final fields into local variables, a new `URLConnection` will be opened everytime.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param xmppServiceAddress address of host
+     * @param relation LinkRelation as a string specifying type of connection
+     * @return list of endpoints
+     * @throws IOException exception due to input/output operations
+     * @throws XmlPullParserException exception encountered during XML parsing
+     * @throws URISyntaxException exception to indicate that a string could not be parsed as a URI reference
+     */
+    public static List<URI> lookup(DomainBareJid xmppServiceAddress, String relation) throws IOException, XmlPullParserException, URISyntaxException {
+        final int buffer_size = 256;
+        InputStream inputStream = getXrdStream(xmppServiceAddress);
```
Flowdalic, line 157: Should use try-with-resources to ensure the input stream is closed in any case.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,187 @@
...
+
+        // Keep reading from the input stream until read completely
+        while (remainingBytes > 0) {
+            byte[] byteBuffer = new byte[buffer_size];
+            int bytes_read = inputStream.read(byteBuffer);
+            obtainedXml.write(byteBuffer, 0, bytes_read);
+            remainingBytes = remainingBytes - bytes_read;
+        }
+        inputStream.close();
+        XmlPullParser parser = PacketParserUtils.getParserFor(obtainedXml.toString());
```
Flowdalic, line 169: Again, there is no need to additionally buffer the input stream, simply use `PacketParserUtils.getParserFor(InputStream)`.

smack-core/src/main/java/org/jivesoftware/smack/util/MultiMap.java
```diff
@@ -217,6 +218,13 @@ public void putAll(Map<? extends K, ? extends V> map) {
         }
     }
 
+    public void putAll(K key, List<V> list) {
```
Flowdalic, line 221: That method appears to be no longer used and can be removed.

smack-core/src/test/java/org/jivesoftware/smack/altConnections/HttpLookupMethodTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jivesoftware.smack.altConnections.HttpLookupMethod.LinkRelation;
+import org.jivesoftware.smack.util.PacketParserUtils;
+import org.jivesoftware.smack.xml.XmlPullParserException;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+public class HttpLookupMethodTest {
+    private final String xml = "<XRD xmlns=\"http://docs.oasis-open.org/ns/xri/xrd-1.0\">\n" +
```
Flowdalic, line 34: Please add another test where two websocket endpoints are announced.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,187 @@
...
+                }
+                String endpointUri = parser.getAttributeValue("href");
+                uriList.add(new URI(endpointUri));
+                tag = parser.nextTag();
+                break;
+            case END_ELEMENT:
+                if (parser.getDepth() == initialDepth) {
+                    break loop;
+                }
+                tag = parser.nextTag();
```
Flowdalic, line 124: That's an unusually parsing loop design. IIRC the call to nextTag() is once right at the start of the loop, and not twice, one time before the loop and in the end element case.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,187 @@
...
+     * Get remote endpoints for the given LinkRelation from host.
+     *
+     * @param xmppServiceAddress address of host
+     * @param relation LinkRelation as a string specifying type of connection
+     * @return list of endpoints
+     * @throws IOException exception due to input/output operations
+     * @throws XmlPullParserException exception encountered during XML parsing
+     * @throws URISyntaxException exception to indicate that a string could not be parsed as a URI reference
+     */
+    public static List<URI> lookup(DomainBareJid xmppServiceAddress, String relation) throws IOException, XmlPullParserException, URISyntaxException {
```
Flowdalic, line 155: Please move this method and the method below to the top of the class, since those are the most important high level methods.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,187 @@
...
+ * </pre>
+ * @see <a href="https://xmpp.org/extensions/xep-0156.html#http">
+ *     HTTP Lookup Method from XEP-0156.
+ *     </a>
+ */
+public final class HttpLookupMethod {
+    /**
+     * Specifies a link relation for the selected type of connection.
+     */
+    enum LinkRelation {
```
Flowdalic, line 58: I think this enum should be public.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,174 @@
...
+        loop: while (true) {
+            XmlPullParser.TagEvent tag = parser.nextTag();
+            switch (tag) {
+            case START_ELEMENT:
+                String name = parser.getName();
+                String namespace = parser.getNamespace();
+                String rel = parser.getAttributeValue("rel");
+
+                if (!namespace.equals(xrd_namespace) || !name.equals("Link") || !rel.equals(relation)) {
+                    tag = parser.nextTag();
```
Flowdalic, line 144: That appears wrong now, as there is a call to `nextTag()` at the start of the loop.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @param parser XmlPullParser that contains LinkRelations
+     * @param relation type of endpoints specified by the given LinkRelation
+     * @return list of endpoints
+     * @throws IOException exception due to input/output operations
+     * @throws XmlPullParserException exception encountered during XML parsing
+     * @throws URISyntaxException exception to indicate that a string could not be parsed as a URI reference
+     */
+    public static List<URI> parseXrdLinkReferencesFor(XmlPullParser parser, String relation) throws IOException, XmlPullParserException, URISyntaxException {
+        final String xrd_namespace = "http://docs.oasis-open.org/ns/xri/xrd-1.0";
```
Flowdalic, line 130: Should be a private static final String constant in the class.

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,174 @@
...
+                String name = parser.getName();
+                String namespace = parser.getNamespace();
+                String rel = parser.getAttributeValue("rel");
+
+                if (!namespace.equals(xrd_namespace) || !name.equals("Link") || !rel.equals(relation)) {
+                    tag = parser.nextTag();
+                    continue loop;
+                }
+                String endpointUri = parser.getAttributeValue("href");
+                uriList.add(new URI(endpointUri));
```
Flowdalic, line 148: nit: I'd love to see this URI created with its on variable

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,174 @@
...
+    /**
+     * Constructs a HTTP connection with the host specified by the DomainBareJid
+     * and retrieves XRD element in the form of an InputStream.
+     *
+     * @param xmppServiceAddress address of host
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
+        final String metadataUrl = "https://" + xmppServiceAddress.toString() + "/.well-known/host-meta";
```
Flowdalic, line 113: nit: calling `toString()` is not necessary here

smack-core/src/main/java/org/jivesoftware/smack/altConnections/HttpLookupMethod.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Constructs a HTTP connection with the host specified by the DomainBareJid
+     * and retrieves XRD element in the form of an InputStream.
+     *
+     * @param xmppServiceAddress address of host
+     * @return InputStream containing XRD element
+     * @throws IOException exception due to input/output operations
+     */
+    public static InputStream getXrdStream(DomainBareJid xmppServiceAddress) throws IOException {
```
Flowdalic, line 112: How does this method behave if no XRD document is published? This should be documented. I think it would ideally return `null` in this case.

Flowdalic: Forget that about 'null', simply let the method throw.

GitCop:

 * Commit: af90b07a668d2cd7280c265875bc0c149837e6d7

---

smack-core/src/test/java/org/jivesoftware/smack/altConnections/HttpLookupMethodTest.java
```diff
@@ -0,0 +1,48 @@
...
```
Flowdalic, line 34: Pro Tip: Use single quotes within the string to quote the attribute values, which do not need to be escaped.

smack-core/src/test/java/org/jivesoftware/smack/altConnections/HttpLookupMethodTest.java
```diff
@@ -0,0 +1,48 @@
...
+                    "<Link rel=\"urn:xmpp:alt-connections:xbosh\" href=\"https://igniterealtime.org:443/http-bind/\"/>\n" +
```
Flowdalic, line 35: Another test querying for the BOSH endpoint would be great!

documentation/extensions/index.md
```diff
@@ -15,6 +15,7 @@ Currently supported XEPs of Smack (all sub-projects)
 | Name                                        | XEP                                                    | Version   | Description |
 |---------------------------------------------|--------------------------------------------------------|-----------|----------------------------------------------------------------------------------------------------------|
 | Nonzas                                      | [XEP-0360](https://xmpp.org/extensions/xep-0360.html)  | n/a       | Defines the term "Nonza", describing every top level stream element that is not a Stanza.                                         |
+| Discovering Alternative XMPP Connection Methods | [XEP-0156](https://xmpp.org/extensions/xep-0156.html)  | n/a       | Defines ways to discover alternative connection methods.                                                              |
```
Flowdalic, line 18: This should be sorted in before xep360.

adiaholic: Aah right !

PR #403, opened 2020-06-26
Remove unrequired assignment of value to connectionEndpoint variable

adiaholic:
The current code would work just fine for a connection having multiple endpoints. However, when there is only one endpoint `ConnectionAttemptState.nextAddress()` would return `null`, since `connectionEndpointIterator` has already iterated over the only possible value in the contructor leading to a `NullPointerException`. This means that during establishment of a connection having multiple endpoints, the first value inside `connectionEndpointIterator` would always be overlooked.


GitCop:

 * Commit: b836588b348ae07a71235ee80d21b0409e146da4

---

PR #406, opened 2020-07-04
OX: Improvements to Secret key backup restore function

vanitasvitae:
I did some further experiments with Smacks OX implementation and noticed some strange behaviours.
Firstly, keys that are restored from a secret key backup are not marked trusted, so sending messages will fail because no trusted own key can be found. This is fixed by trusting the key after restoring.
Secondly, public keys are extracted from the backup twice, which is unnecessary work.


Flowdalic:
LGTM, one minor nit: Could you prefix your commit messages either with '[OX]' or 'OX:'. Especially the "Trust key upon import" commit message subject is very vague without this.

vanitasvitae:
Sure :)

PR #408, opened 2020-07-15
Allow user-supplied secret key backup passphrases

vanitasvitae:
Instead of passing the passphrase back to the user via a
DisplayBackupCodeCallback, we directly return the passphrase
which is now represented by a class.

Also we now allow the user to provide the passphrase
when backing up the key to the server.


smack-core/src/main/java/org/jivesoftware/smack/util/StringUtils.java
```diff
@@ -270,7 +270,7 @@ public static String encodeHex(byte[] bytes) {
     /**
      * 24 upper case characters from the latin alphabet and numbers without '0' and 'O'.
      */
-    private static final char[] UNAMBIGUOUS_NUMBERS_AND_LETTER = "123456789ABCDEFGHIJKLMNPQRSTUVWXYZ".toCharArray();
+    public static final char[] UNAMBIGUOUS_NUMBERS_AND_LETTER = "123456789ABCDEFGHIJKLMNPQRSTUVWXYZ".toCharArray();
```
Flowdalic, line 273: If this array is public, then users would modify its contents. I think it should stay private. I guess you idea was to use it in the regex below, which we definitely should!
How about:

```
public static final String UNAMBIGUOUS_NUMBERS_AND_LETTER_STRING = "123456789ABCDEFGHIJKLMNPQRSTUVWXYZ";
private static final char[] UNAMBIGUOUS_NUMBERS_AND_LETTER = UNAMBIGUOUS_NUMBERS_AND_LETTER_STRING.toCharArray();
```

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpManager.java
```diff
@@ -551,7 +577,7 @@ public void run() {
                 if (contentElement instanceof SigncryptElement) {
                     for (SigncryptElementReceivedListener l : signcryptElementReceivedListeners) {
                         l.signcryptElementReceived(contact, message, (SigncryptElement) contentElement,
-                                        decrypted.getMetadata());
```
Flowdalic, line 554: Unrelated whitespace change

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpManager.java
```diff
@@ -566,7 +592,7 @@ public void run() {
                 if (contentElement instanceof CryptElement) {
                     for (CryptElementReceivedListener l : cryptElementReceivedListeners) {
                         l.cryptElementReceived(contact, message, (CryptElement) contentElement,
-                                        decrypted.getMetadata());
+                                decrypted.getMetadata());
```
Flowdalic, line 595: Unrelated whitespace change

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpSecretKeyBackupPassphrase.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Represents a secret key backup passphrase whose format is described in XEP-0373 §5.3.
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0373.html#backup-encryption">
+ *      XEP-0373 §5.4 Encrypting the Secret Key Backup</a>
+ */
+public class OpenPgpSecretKeyBackupPassphrase implements CharSequence {
+
+    private static final Pattern PASSPHRASE_PATTERN = Pattern.compile(
+            "^([123456789ABCDEFGHIJKLMNPQRSTUVWXYZ]{4}-){5}[123456789ABCDEFGHIJKLMNPQRSTUVWXYZ]{4}$");
```
Flowdalic, line 30: Should re-use UNAMBIGUOUS_NUMBERS_AND_LETTER_STRING from StringUtils

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpSecretKeyBackupPassphrase.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private final String passphrase;
+
+    public OpenPgpSecretKeyBackupPassphrase(String passphrase) {
+        if (!PASSPHRASE_PATTERN.matcher(passphrase).matches()) {
```
Flowdalic, line 35: I'd love to see unit tests for this constructor. One with a valid passphrase and one with an invalid one would be a great start. :)

smack-core/src/main/java/org/jivesoftware/smack/util/StringUtils.java
vanitasvitae: Ah right, haven't thought of that.
Yeah I guess either the fix you suggested, or just make it private again, as I ended up constructing the regex without the array.

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpManager.java
vanitasvitae: Are you happy if I move this change to another commit?

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpSecretKeyBackupPassphrase.java
vanitasvitae: There is a test for a valid construction already, but I happily add another one as well :)

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpManager.java
vanitasvitae: Moved into separate commit

smack-openpgp/src/main/java/org/jivesoftware/smackx/ox/OpenPgpSecretKeyBackupPassphrase.java
vanitasvitae: Added dedicated junit tests.

Flowdalic:
Note that the "Fix indentation" commit message would also be nice if it was prefixed. But since it is just a code style change, I am merging this as is anyways.

PR #410, opened 2020-07-22
Add SimpleXmppConnectionIntegrationTest

adiaholic:

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
@@ -1,0 +1,67 @@
+/**
+ *
+ * Copyright 2020 Aditya Borikar
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smack.c2s;
+
+import java.util.concurrent.TimeoutException;
+
+import org.jivesoftware.smack.StanzaListener;
+import org.jivesoftware.smack.filter.MessageWithBodiesFilter;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.packet.Stanza;
+
+import org.igniterealtime.smack.inttest.AbstractSmackIntegrationTest;
+import org.igniterealtime.smack.inttest.SmackIntegrationTestEnvironment;
+import org.igniterealtime.smack.inttest.annotations.SmackIntegrationTest;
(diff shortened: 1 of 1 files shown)


smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class SimpleXmppConnectionIntegrationTest extends AbstractSmackIntegrationTest {
+
+    public SimpleXmppConnectionIntegrationTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+    }
+
+    @SmackIntegrationTest
+    public void createConnectionTest() throws TimeoutException, Exception {
+        IntegrationTestRosterUtil.ensureBothAccountsAreSubscribedToEachOther(conOne, conTwo, timeout);
```
Flowdalic, line 42: I don't think this is necessary.

smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+        EntityBareJid userTwo = conTwo.getUser().asEntityBareJid();
+
+        final String messageBody = "Hello from the other side!";
```
Flowdalic, line 46: To be extra paranoid, please mix in the integration run test id in the message body, so that it is unique across integration test runs. This is to prevent a stale message from a previous run being considered the right one.

smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+        final StanzaListener stanzaListener = new StanzaListener() {
+            @Override
+            public void processStanza(Stanza stanza) {
+                Message message = (Message) stanza;
+                if (message.getBody().equals(messageBody)) {
+                    messageReceived.signal();
+                }
+            }
+        };
+        conTwo.addAsyncStanzaListener(stanzaListener, MessageWithBodiesFilter.INSTANCE);
```
Flowdalic, line 63: I suggest using a lambda instead of `stanzaListener` here.

smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public SimpleXmppConnectionIntegrationTest(SmackIntegrationTestEnvironment environment) {
+        super(environment);
+    }
+
+    @SmackIntegrationTest
+    public void createConnectionTest() throws TimeoutException, Exception {
+        IntegrationTestRosterUtil.ensureBothAccountsAreSubscribedToEachOther(conOne, conTwo, timeout);
+
+        EntityBareJid userTwo = conTwo.getUser().asEntityBareJid();
```
Flowdalic, line 44: You don't have to convert this to an `EntityBareJid`, simply use the `EntityFullJid` returned by `getUser()`.

smack-integration-test/src/main/java/org/jivesoftware/smack/c2s/SimpleXmppConnectionIntegrationTest.java
vanitasvitae: "I must've called a thousand times" :P

PR #413, opened 2020-07-24
HttpLookupMethod: Position parser at START_ELEMENT before parsing

adiaholic:
This PR aims to provide `HttpLookupMethod.lookp(..)` method the ability to parse forward to the first `START_ELEMENT` tag. The `HttpLookupMethodTest` tests the `HttpLookupMethod` class by parsing `String`. This makes use of `PacketParserUtils.getParserFor(String)`, which already does forward winding to reach `START_ELEMENT`. However when fetching endpoints from a remote host meta data, `PacketParserUtils.getParserFor(InputStream)` is used which doesn't do winding in any form. And thus, even though `HttpLookupMethodTest` tests pass, this implementation would crash.

I feel that the change should be incorporated inside `PacketParserUtils` but it would correspond to changes in other modules as well making the change a bit more involved. Hence putting this PR to have feedback on the issue.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/altconnections/HttpLookupMethod.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/altconnections/HttpLookupMethod.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.jivesoftware.smack.util.PacketParserUtils;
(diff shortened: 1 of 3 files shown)


Flowdalic:
- Create new method `forwardToNextStartElement()` in ParserUtils
- Move code from `PacketParserUtils.getParserFor(Reader)` into there
- Use this method in parseXrdLinkReferencesFor().

PR #420, opened 2020-08-14
Introduce StreamOpenFactory for modular architecture

adiaholic:


smack-core/src/main/java/org/jivesoftware/smack/c2s/ModularXmppClientToServerConnection.java
```diff
@@ -1086,4 +1093,8 @@ public void appendStatsTo(ExtendedAppendable appendable) throws IOException {
         }
 
     }
+
+    public interface StreamOpenFactory {
```
Flowdalic, line 1097: This should be in its own file within o.j.s.c2s

I also wonder: Don't we need that also for the stream close? Hence I wonder if we should rename this to `StreamOpenAndCloseFactory` (or something?).

smack-core/src/main/java/org/jivesoftware/smack/c2s/ModularXmppClientToServerConnection.java
```diff
@@ -1086,4 +1093,8 @@ public void appendStatsTo(ExtendedAppendable appendable) throws IOException {
...
+        StreamOpen getStreamOpen();
```
Flowdalic, line 1098: Factory methods are usually prefixed with "create", e.g. `createFoo()`, not "get". Since they are factories, they create new objects. Hence this should be `createStreamOpen()`.

Furthermore, the method should have parameters, i.e. everything that is required to create a new stream open. Those are basically the same as [this StreamOpen concstructor](https://github.com/igniterealtime/Smack/blob/45e865757a7242ea4942855213b53a9670c06f9c/smack-core/src/main/java/org/jivesoftware/smack/packet/StreamOpen.java#L71) (minus the StreamContentNamespace, which is always 'jabber:client').

smack-core/src/main/java/org/jivesoftware/smack/c2s/ModularXmppClientToServerConnection.java
```diff
@@ -560,10 +561,16 @@ protected void waitForFeaturesReceived(String waitFor)
     protected void newStreamOpenWaitForFeaturesSequence(String waitFor) throws InterruptedException,
                     SmackException, XMPPException {
         prepareToWaitForFeaturesReceived();
-        sendStreamOpen();
+        StreamOpenFactory streamOpenFactory = activeTransport.getStreamOpenFactory();
+        sendStreamOpen(streamOpenFactory.getStreamOpen());
```
Flowdalic, line 565: Please add a new line with the declaration of `StreamOpen streamOpen = …`.

smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XmppTcpTransportModule.java
```diff
@@ -580,6 +582,24 @@ protected XmppTcpNioTransport(ModularXmppClientToServerConnectionInternal connec
...
+                @Override
+                public StreamOpen getStreamOpen() {
+                    CharSequence to = connectionInternal.connection.getXMPPServiceDomain();
+                    CharSequence from = null;
+                    CharSequence localpart = connectionInternal.connection.getConfiguration().getUsername();
+                    if (localpart != null) {
+                        from = XmppStringUtils.completeJidFrom(localpart, to);
+                    }
+                    String id = connectionInternal.connection.getStreamId();
+                    StreamOpen streamOpen = new StreamOpen(to, from, id, connectionInternal.connection.getConfiguration().getXmlLang(), StreamOpen.StreamContentNamespace.client);
```
Flowdalic, line 597: Now this duplicates code from `AbstractXMPPConnection.sendStreamOpen()`, but after thinking a little bit about it, I can't come up with a nice way to deduplicate this. I'd say we keep it for now as is, just wanted to point that out.

adiaholic: Thanks for pointing that out. However, I have tried removing `AbstractXMPPConnection.sendStreamOpen()` and found that it still is required for non-modular connections. So lets keep it until `XmppTCPConnection/ XMPPBOSHConnection` is deprecated :)

smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XmppTcpTransportModule.java
```diff
@@ -57,7 +57,9 @@
 import org.jivesoftware.smack.XmppInputOutputFilter;
 import org.jivesoftware.smack.c2s.ModularXmppClientToServerConnection.ConnectedButUnauthenticatedStateDescriptor;
 import org.jivesoftware.smack.c2s.ModularXmppClientToServerConnection.LookupRemoteConnectionEndpointsStateDescriptor;
+
```
vanitasvitae, line 60: Nit: Whitespace change

smack-core/src/main/java/org/jivesoftware/smack/c2s/ModularXmppClientToServerConnection.java
```diff
@@ -560,10 +562,25 @@ protected void waitForFeaturesReceived(String waitFor)
...
                     SmackException, XMPPException {
         prepareToWaitForFeaturesReceived();
-        sendStreamOpen();
+
+        // Create StreamOpen from StreamOpenAndCloseFactory via underlying transport.
+        StreamOpenAndCloseFactory streamOpenAndCloseFactory = activeTransport.getStreamOpenAndCloseFactory();
+        CharSequence from = null;
+        CharSequence localpart = connectionInternal.connection.getConfiguration().getUsername();
+        if (localpart != null) {
+            from = XmppStringUtils.completeJidFrom(localpart, getXMPPServiceDomain());
```
vanitasvitae, line 571: I'm not sure if this behaviour is correct in all cases, as the username can be different from the localpart of the jid.
However, I certainly wouldn't know how to fix this instead :D

Flowdalic: Servers will ignore the 'from' in c2s anyway, here it is merely a debug aid.

smack-core/src/main/java/org/jivesoftware/smack/c2s/StreamOpenAndCloseFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smack.c2s;
+
+import org.jivesoftware.smack.packet.StreamClose;
+import org.jivesoftware.smack.packet.StreamOpen;
+
+public interface StreamOpenAndCloseFactory {
+    StreamOpen createStreamOpen(CharSequence to, CharSequence from, String id, String lang);
```
Flowdalic, line 23: I just thought some more about it. Could you create `AbstractStreamOpen` and subclass `StreamOpen`, and then in your websocket branch `WebsocketOpen` from it. Same goes probably for `StreamClose` → `AbstractStreamClose`.

Flowdalic: And bonus points if you do the AbstractStreamOpen/StreamOpen splits in an extra commit. :)

adiaholic: Made changes accordingly in an extra commit :smiley:

smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractStreamOpen.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smack.packet;
+
+public abstract class AbstractStreamOpen implements Nonza {
```
Flowdalic, line 19: This should probably extend/implement `TopLevelStreamElement` and not `Nonza` (since it is not a Nonza).

Also all fields of StreamOpen should be moved into this class, together with the constructors and getters.

smack-core/src/main/java/org/jivesoftware/smack/packet/AbstractStreamOpen.java
```diff
@@ -0,0 +1,81 @@
...
+import org.jivesoftware.smack.util.StringUtils;
+
+/**
+ * AbstractStreamOpen is actually a {@link TopLevelStreamElement}, however we
+ * implement {@link Nonza} here. This is because, {@link XMPPConnection} doesn't
+ * yet support sending {@link TopLevelStreamElement} directly and the same can only
+ * be achieved through {@link XMPPConnection#sendNonza(Nonza)}.
+ */
+public abstract class AbstractStreamOpen implements Nonza {
+    public static final String ELEMENT = "stream:stream";
```
Flowdalic, line 30: That is a property of the concrete subclass, i.e., `StreamOpen`, not the abstract superclass.

smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XmppTcpTransportModule.java
```diff
@@ -580,6 +582,21 @@ protected XmppTcpNioTransport(ModularXmppClientToServerConnectionInternal connec
             super(connectionInternal);
         }
 
+        @Override
+        public StreamOpenAndCloseFactory getStreamOpenAndCloseFactory() {
+            return new StreamOpenAndCloseFactory() {
+                @Override
+                public StreamOpen createStreamOpen(CharSequence to, CharSequence from, String id, String lang) {
+                    StreamOpen streamOpen = new StreamOpen(to, from, id, connectionInternal.connection.getConfiguration().getXmlLang(), StreamOpen.StreamContentNamespace.client);
```
Flowdalic, line 590: That line is pretty long, and I don't like having method calls to fetch arguments. Please introduce a variable for xmlLang and use it. This also improves readability of the code.

smack-core/src/main/java/org/jivesoftware/smack/packet/StreamOpen.java
```diff
@@ -17,49 +17,12 @@
...
-     * RFC 6120 § 4.7.4.
-     */
-    private final String lang;
-
-    /**
-     * RFC 6120 § 4.8.2.
-     */
-    private final String contentNamespace;
-
+public class StreamOpen extends AbstractStreamOpen {
```
Flowdalic, line 25: This class should become `final`.

adiaholic: @Flowdalic , Aren't we going to inherit this class into `WebsocketOpen` ?

PR #431, opened 2020-09-02
Remove AbstractWebsocketNonza since it is no longer needed

adiaholic:
This class is a residue of older design pattern and is no longer needed.


Flowdalic:
Thanks, applied as https://github.com/igniterealtime/Smack

PR #434, opened 2020-09-14
Mac & Windows build instructions

Fishbowler:
Adds instructions for Mac & Windows to work with the case-sensitivity required by the Smack codebase.

Tested Mac 👍
Researched Windows 🤞


Flowdalic:

PR #435, opened 2020-10-03
Add missing right parenthesis in SINT output

guusdk:
While running the Smack integration tests, a line like this is printed to std-out somewhere

    Available tests: 21 (Disabled 12 classes and 6 tests

There's a missing character on the end of that line, which makes the author of this commit twitch.
This commit adds the missing character, resulting in a line like this:

    Available tests: 21 (Disabled 12 classes and 6 tests)


vanitasvitae (approved):
This clearly indicates that the tests are not run often enough :D

adiaholic (approved):

vanitasvitae:
@Flowdalic it looks like although Github shows the travis pipeline as stagnant, the build actually went through successfully.

Could you restart the build, so that we don't loose track of this PR due to Github showing incomplete checks?

PR #436, opened 2020-10-03
In tests, use specific assertion implementation

guusdk:
Although this does not functionally change anything, the error messages that are generated by the more specific assert* methods can be helpful.

Take, for example, this assertion:

    assertTrue(muc.getOccupantsCount()==1);

When this assertion fails (https://github.com/igniterealtime/Openfire/pull/1721#issuecomment-703112496) then the most likely first question that a developer is going to ask is: "what was the count?" The assertion above would have printed something along these lines:

    org.opentest4j.AssertionFailedError: expected: <true> but was: <false>

That is not really helpful here. If, instead, a more specific assertion was used (like what's done in this commit), then someting like this would have been printed:

    org.opentest4j.AssertionFailedError: expected: 1 but was: 2

That provides more helpful information! On top of that, the code argably expresses intent better when using the latter form, especially when the assertions become more complex.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
@@ -16,7 +16,9 @@
  */
 package org.jivesoftware.smackx.muc;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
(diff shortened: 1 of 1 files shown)


adiaholic (approved):

guusdk:
*sigh* Even after taking extra time to compose this commit, something is still blocking the contribution. I've now spent about ten times more time on trying to mold and shape a PR in the desired format than the time that actually went in the code changes.

I'm sure that I can eventually get this to pass. I'm also sure that I really do not want to waste more of my time. My primary goal here was to give some advice to @adiaholic on how to slightly improve the way he writes tests. Mission accomplished.

adiaholic:
Thanks a lot for going out of your way to help me improve :)

Flowdalic:
Could either @guusdk  or @vanitasvitae squash the commits? Thanks.

guusdk:
Squashed.

PR #437, opened 2020-10-04
Fix MUC enter race condition

guusdk:
When joining/entering a Multi User Chat, the 'enter' method should not return before the internal state of the instance has been updated to reflect the join.

The internal state of `MultiUserChat` is managed b[y a listener that triggers on presence stanzas](https://github.com/igniterealtime/Smack/blob/5782fff2a48f4f58acd8e2c81ea9921ffe31351f/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java#L188). A stanza filter, triggered by the same presence stanza, is used in `enter()` to detect a successful join. This introduces a race condition.

The race condition manifests itself in scenarios like these:

        int before = muc.getOccupantsCount();
        MultiUserChat muc = mucManager.getMultiUserChat(mucAddress);
        muc.join(Resourcepart.from("nick"));
        int after = muc.getOccupantsCount();

Although the `join` on the third line is blocking, the `after` and `before` count can be equal. Assuming no other occupant changes occur, you would expect the count to have increased with one.

The race condition can be verified by adding a delay in the processing of presence updates, somewhere in this code: https://github.com/igniterealtime/Smack/blob/5782fff2a48f4f58acd8e2c81ea9921ffe31351f/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java#L188

The commit in this PR prevents the race condition by modifying `enter`, registering a listener that is triggered only after the presence update has been processed.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -385,7 +385,7 @@ public class MultiUserChat {
         // occupant (eg: #getOccupantsCount might return an incorrect number). To prevent this, this 'enter' method does
         // not return until after the internal state of this class has been updated to reflect that the user joined the muc.
         final ArrayDeque<EntityFullJid> queue = new ArrayDeque<>();
-        final ParticipantStatusListener internalStateUpdateListener = new DefaultParticipantStatusListener() {
+        final ParticipantStatusListener internalStateUpdateListener = new ParticipantStatusListener() {
             @Override
             public void joined(EntityFullJid participant) {
                 synchronized (this) {
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -17,6 +17,7 @@
 
 package org.jivesoftware.smackx.muc;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Fishbowler:
I've seen this flaky test a few times. Rerunning the job usually solves it, but it degrades the usefulness of fast feedback from CI.

https://github.com/Fishbowler/Openfire/pull/5/checks?check_run_id=1204127397

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -377,16 +378,53 @@ private Presence enter(MucEnterConfiguration conf) throws NotConnectedException,
...
                             selfPresenceCollector).setStanzaFilter(presenceFromRoomFilter);
             // This stanza collector is used to reset the timeout of the selfPresenceCollector.
             presenceStanzaCollector = connection.createStanzaCollector(presenceStanzaCollectorConfguration);
+            final long waitStart = System.currentTimeMillis();
             presence = selfPresenceCollector.nextResultOrThrow(conf.getTimeout());
+
+            // Wait for the internal state of this instance of MultiUserChat to have been updated with the join.
+            long remainingWait = conf.getTimeout() - (System.currentTimeMillis() - waitStart);
+            outer:
+            while (remainingWait > 0) {
```
Flowdalic, line 415: This does nothing we never received the self-presence within the timeout period.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -377,16 +378,53 @@ private Presence enter(MucEnterConfiguration conf) throws NotConnectedException,
                         )
                     );
         // @formatter:on
+
+        // 'responseFilter' is triggered by the same stanza as org.jivesoftware.smackx.muc.MultiUserChat.presenceListener,
+        // a listener that is responsible for updating the internal state of this instance of MultiUserChat. This causes
+        // a race condition: after this 'enter' method returns, other methods might not yet be aware of the newly entered
+        // occupant (eg: #getOccupantsCount might return an incorrect number). To prevent this, this 'enter' method does
+        // not return until after the internal state of this class has been updated to reflect that the user joined the muc.
+        final ArrayDeque<EntityFullJid> queue = new ArrayDeque<>();
```
Flowdalic, line 387: Can't we simply use a blocking queue?

Flowdalic (commented):
Thanks, I am working on this based on my review.

Flowdalic:
Thanks for pointing this issue out. I opted for a slightly different solution in https://github.com/igniterealtime/Smack/commit/7e311ab9df129b843e4fbcf0905834673624f220

MartinHoerning:
> Thanks for pointing this issue out. I opted for a slightly different solution in [7e311ab](https://github.com/igniterealtime/Smack/commit/7e311ab9df129b843e4fbcf0905834673624f220)

Hi,
I had problems with this fix when switching to another MUC. When I call join() of the new MUC directly after leave() of the old MUC, it seams that sometimes the presenceListener of the new MUC gets some 'available' Presence that does not contain the PRESENCE_TO_SELF_110 status first. Only a second callback will then contain the PRESENCE_TO_SELF_110 status but this will not reach the 'processedReflectedSelfPresence = true;' line because of the previous presence. Thus the enter() will never finish.
I fixed it by checking for the PRESENCE_TO_SELF_110 even if an oldPresence is found.  
I'm completely new to xmpp and smack-extensions and found this while debugging foreign code. So I don't know if this is a bug or if I'm using it wrong.

Flowdalic:
Please post this in the smack community forums. It's better to discuss it there, instead of a closed PR.

PR #440, opened 2020-10-12
SINT: wait for notification filter to propagate

guusdk:
Several PEP-based Smack Integration Test Framework tests have been disabled in the CI pipeline for Openfire, as they frequently, but inconsistently, failed. The changes in this PR address that.

Each of these tests depend on a pubsub notification filter to be in place, to allow the intended recipient to receive a notification. A race condition exists where the setting of that filter competes with the publishing of data.

The changes in this PR add a synchronization point to explicitly wait for the filter to have been set, before proceeding to execute the test.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/geolocation/GeolocationIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/geolocation/GeolocationIntegrationTest.java
@@ -17,6 +17,7 @@
 package org.jivesoftware.smackx.geolocation;
 
 import java.net.URI;
+import java.util.concurrent.TimeoutException;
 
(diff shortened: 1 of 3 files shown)


vanitasvitae (changes requested):
What is that `tatus` file that got changed?

guusdk:
Unsure how exactly I managed to get that `tatus` file committed. It's gone now.

Flowdalic:
Could you rebase this on the latest master? I think this introduces changes to existing tests and adds new ones. Ideally this would be in two separate commits

guusdk:
Rebased to current master, and refactored to make use of the change introduced in https://github.com/igniterealtime/Smack/pull/441

Flowdalic:
Merged as https://github.com/igniterealtime/Smack/commit/819ed87a178e71c6c40b70d6a5f079751337b51f

PR #441, opened 2020-10-14
Add method to remove registered CAPS listener

guusdk:
ServiceDiscoveryManager allows a listener for CAPS changes to be
registered. It should also allow to remove a previously registered
listener.


vanitasvitae (approved):
Good catch!

PR #442, opened 2020-10-24
OpenPgpManager: Expose methods to generate and import keys

vanitasvitae:
This PR exposes methods to generate OpenPGP keys in the OpenPgpManager.

The motivation for this change is to allow key generation without the user having to access the store.


Flowdalic:
Applied to the 4.4 branch as https://github.com/igniterealtime/Smack/commit/1a9ac238e85f4a3a028b4a23dde671bcc2170272

PR #443, opened 2020-10-28
Add Github Actions workflow

Fishbowler:
Following on from work done for Openfire, reproduce the Travis workflow in Github Actions to reduce the number of CI systems involved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,87 @@
+name: Smack CI
+
+env:
+  GRADLE_VERSION: 6.2
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


Flowdalic:
I have looked into this and made some minor modifications to this PR in https://github.com/igniterealtime/Smack/pull/451

The main issue right now is that there is no Android SDK 19 installed. I guess we could use https://github.com/android-actions/setup-android/ for that.

However, I am not really sold on github actions right now, but OTOH I am also not really attached to travis. But eventually there should only be one CI system performing the same task.

Flowdalic:
Merged https://github.com/igniterealtime/Smack/pull/451

Closing this PR. Thanks for your contribution.

PR #444, opened 2020-11-09
Prevent premature reading of file transfer data

guusdk:
When reading from an inputstream, reading 0 bytes does not signal EOF.

I suspect that this is causing an issue where a file is (partially)
transferred, before being marked as 'cancelled'. Spark suffers from
this (SPARK-2200). Potentially affects files that contain multi-byte
characters more than others.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/FileTransfer.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/filetransfer/FileTransfer.java
@@ -209,7 +209,7 @@ public abstract class FileTransfer {
         int count = 0;
         amountWritten = 0;
 
-        while ((count = in.read(b)) > 0 && !getStatus().equals(Status.cancelled)) {
+        while ((count = in.read(b)) > -1 && !getStatus().equals(Status.cancelled)) {
             out.write(b, 0, count);
             amountWritten += count;
         }


Flowdalic:
`InputStream.read(b)` javadoc states

> If the length of b is zero, then no bytes are read and 0 is returned; otherwise, there is an attempt to read at least one byte. If no byte is available because the stream is at the end of the file, the value -1 is returned; otherwise, at least one byte is read and stored into b.

and since `b` size in this case, is non-zero, we should never see this method return zero.

So, while this does not apply the typical pattern of "read() until EOF/-1", I do not think the existing code is incorrect.

guusdk:
I agree with your reading of the javadoc that 0 cannot occur. However, what if it does? The InputStream that's being fed is one of our own creation that could have, or could in the future be modified to misbehave in this respect (would it, for example, be be possible to parse a block of data that is the equivalent of an empty bit of data?) Checking for -1 instead of 0 (or, alternatively, explicitly asserting that the value is not 0), would guard against this. Does it have a negative effect?

Flowdalic:
In that case I would use an `assert`:

```
        while ((count = in.read(b)) > 0) {
            // read() must never return 0, as it would violate read()'s API contract.
            assert(count != 0);

            if (getStatus().equals(Status.cancelled)) {
               break;
            }

            out.write(b, 0, count);
            amountWritten += count;
        }
```

Flowdalic:
Closing, feel free to re-open with an `assert` if you want.

PR #446, opened 2020-11-18
Add MUC tests

Fishbowler:
Following an issue in Openfire (https://igniterealtime.atlassian.net/browse/OF-2146) where self-presence wasn't being sent, added tests to Smack to prove the fix and give some love to our future selves and not have it break again.

Required one change to smack-extensions to have muc.join() mirror muc.leave() and return the self-presence that was returned by the server.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -629,6 +629,7 @@ public class MultiUserChat {
      * joining the room, the room will decide the amount of history to send.
      *
      * @param nickname the nickname to use.
+     * @return the leave self-presence as reflected by the MUC.
      * @throws NoResponseException if there was no response from the remote entity.
      * @throws XMPPErrorException if an error occurs joining the room. In particular, a
      *      401 error can occur if no password was provided and one is required; or a
@@ -641,10 +642,11 @@ public class MultiUserChat {
      * @throws InterruptedException if the calling thread was interrupted.
      * @throws NotAMucServiceException if the entity is not a MUC serivce.
      */
-    public void join(Resourcepart nickname) throws NoResponseException, XMPPErrorException,
+    public Presence join(Resourcepart nickname) throws NoResponseException, XMPPErrorException,
                     NotConnectedException, InterruptedException, NotAMucServiceException {
         MucEnterConfiguration.Builder builder = getEnterConfigurationBuilder(nickname);
-        join(builder.build());
+        Presence reflectedJoinPresence = join(builder.build());
+        return reflectedJoinPresence;
     }
 
     /**
@@ -691,6 +693,7 @@ public class MultiUserChat {
      * decide to create a new room or not.
      *
      * @param mucEnterConfiguration the configuration used to enter the MUC.
+     * @return the join self-presence as reflected by the MUC.
      * @throws XMPPErrorException if an error occurs joining the room. In particular, a
      *      401 error can occur if no password was provided and one is required; or a
      *      403 error can occur if the user is banned; or a
@@ -702,7 +705,7 @@ public class MultiUserChat {
      * @throws InterruptedException if the calling thread was interrupted.
      * @throws NotAMucServiceException if the entity is not a MUC serivce.
      */
-    public synchronized void join(MucEnterConfiguration mucEnterConfiguration)
+    public synchronized Presence join(MucEnterConfiguration mucEnterConfiguration)
         throws XMPPErrorException, NoResponseException, NotConnectedException, InterruptedException, NotAMucServiceException {
         // If we've already joined the room, leave it before joining under a new
         // nickname.
@@ -714,7 +717,8 @@ public class MultiUserChat {
                 LOGGER.log(Level.WARNING, "Could not leave MUC prior joining, assuming we are not joined", e);
             }
         }
-        enter(mucEnterConfiguration);
+        Presence reflectedJoinPresence = enter(mucEnterConfiguration);
+        return reflectedJoinPresence;
     }
 
     /**
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
@@ -23,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
 import java.util.concurrent.TimeoutException;
+import java.util.logging.Level;
 
 import org.jivesoftware.smack.MessageListener;
 import org.jivesoftware.smack.SmackException.NoResponseException;
@@ -33,8 +34,6 @@ import org.jivesoftware.smack.packet.Presence;
 import org.jivesoftware.smack.util.StringUtils;
 
 import org.jivesoftware.smackx.muc.MultiUserChat.MucCreateConfigFormHandle;
-import org.jivesoftware.smackx.muc.MultiUserChatException.MucNotJoinedException;
-import org.jivesoftware.smackx.muc.MultiUserChatException.NotAMucServiceException;
 import org.jivesoftware.smackx.muc.packet.MUCUser;
 
 import org.igniterealtime.smack.inttest.AbstractSmackIntegrationTest;
@@ -45,6 +44,7 @@ import org.igniterealtime.smack.inttest.util.ResultSyncPoint;
 import org.igniterealtime.smack.inttest.util.SimpleResultSyncPoint;
 import org.jxmpp.jid.DomainBareJid;
 import org.jxmpp.jid.EntityBareJid;
+import org.jxmpp.jid.EntityFullJid;
 import org.jxmpp.jid.impl.JidCreate;
 import org.jxmpp.jid.parts.Localpart;
 import org.jxmpp.jid.parts.Resourcepart;
@@ -56,6 +56,7 @@ public class MultiUserChatIntegrationTest extends AbstractSmackIntegrationTest {
 
     private final MultiUserChatManager mucManagerOne;
     private final MultiUserChatManager mucManagerTwo;
+    private final MultiUserChatManager mucManagerThree;
     private final DomainBareJid mucService;
 
     public MultiUserChatIntegrationTest(SmackIntegrationTestEnvironment environment)
@@ -64,6 +65,7 @@ public class MultiUserChatIntegrationTest extends AbstractSmackIntegrationTest {
         super(environment);
         mucManagerOne = MultiUserChatManager.getInstanceFor(conOne);
         mucManagerTwo = MultiUserChatManager.getInstanceFor(conTwo);
+        mucManagerThree = MultiUserChatManager.getInstanceFor(conThree);
 
         List<DomainBareJid> services = mucManagerOne.getMucServiceDomains();
         if (services.isEmpty()) {
@@ -74,27 +76,72 @@ public class MultiUserChatIntegrationTest extends AbstractSmackIntegrationTest {
         }
     }
 
+    /**
+     * Asserts that when a user joins a room, they are themselves included on the list of users notified (self-presence).
+     *
+     * <p>From XEP-0045 § 7.2.2:</p>
+     * <blockquote>
+     * ...the service MUST also send presence from the new participant's occupant JID to the full JIDs of all the
+     * occupants (including the new occupant)
+     * </blockquote>
+     *
+     * @throws Exception when errors occur
+     */
     @SmackIntegrationTest
-    public void mucJoinLeaveTest() throws XmppStringprepException, NotAMucServiceException, NoResponseException,
-            XMPPErrorException, NotConnectedException, InterruptedException, MucNotJoinedException {
-        EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from("smack-inttest-join-leave-" + randomString),
-                mucService.getDomain());
+    public void mucJoinTest() throws Exception {
+        EntityBareJid mucAddress = getRandomRoom("smack-inttest-join-leave");
 
         MultiUserChat muc = mucManagerOne.getMultiUserChat(mucAddress);
+        try {
+            Presence reflectedJoinPresence = muc.join(Resourcepart.from("nick-one"));
 
-        muc.join(Resourcepart.from("nick-one"));
+            MUCUser mucUser = MUCUser.from(reflectedJoinPresence);
 
-        Presence reflectedLeavePresence = muc.leave();
+            assertNotNull(mucUser);
+            assertTrue(mucUser.getStatus().contains(MUCUser.Status.PRESENCE_TO_SELF_110));
+            assertEquals(mucAddress + "/nick-one", reflectedJoinPresence.getFrom().toString());
+            assertEquals(reflectedJoinPresence.getTo().toString(), conOne.getUser().asEntityFullJidIfPossible().toString());
+        } finally {
+            tryDestroy(muc);
+        }
+    }
 
-        MUCUser mucUser = MUCUser.from(reflectedLeavePresence);
-        assertNotNull(mucUser);
+    /**
+     * Asserts that when a user leaves a room, they are themselves included on the list of users notified (self-presence).
+     *
+     * <p>From XEP-0045 § 7.14:</p>
+     * <blockquote>
+     * The service MUST then send a presence stanzas of type "unavailable" from the departing user's occupant JID to
+     * the departing occupant's full JIDs, including a status code of "110" to indicate that this notification is
+     * "self-presence"
+     * </blockquote>
+     *
+     * @throws Exception when errors occur
+     */
+    @SmackIntegrationTest
+    public void mucLeaveTest() throws Exception {
+        EntityBareJid mucAddress = getRandomRoom("smack-inttest-join-leave");
 
-        assertTrue(mucUser.getStatus().contains(MUCUser.Status.PRESENCE_TO_SELF_110));
+        MultiUserChat muc = mucManagerOne.getMultiUserChat(mucAddress);
+        try {
+            muc.join(Resourcepart.from("nick-one"));
+
+            Presence reflectedLeavePresence = muc.leave();
+
+            MUCUser mucUser = MUCUser.from(reflectedLeavePresence);
+            assertNotNull(mucUser);
+
+            assertTrue(mucUser.getStatus().contains(MUCUser.Status.PRESENCE_TO_SELF_110));
+            assertEquals(mucAddress + "/nick-one", reflectedLeavePresence.getFrom().toString());
+            assertEquals(reflectedLeavePresence.getTo().toString(), conOne.getUser().asEntityFullJidIfPossible().toString());
+        } finally {
+            tryDestroy(muc);
+        }
     }
 
     @SmackIntegrationTest
     public void mucTest() throws Exception {
-        EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from("smack-inttest-" + randomString), mucService.getDomain());
+        EntityBareJid mucAddress = getRandomRoom("smack-inttest");
 
         MultiUserChat mucAsSeenByOne = mucManagerOne.getMultiUserChat(mucAddress);
         MultiUserChat mucAsSeenByTwo = mucManagerTwo.getMultiUserChat(mucAddress);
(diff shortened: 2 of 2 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -621,6 +621,7 @@ public MucCreateConfigFormHandle createOrJoinIfNecessary(Resourcepart nickname,
      * joining the room, the room will decide the amount of history to send.
      *
      * @param nickname the nickname to use.
+     * @return the leave presence as reflected by the MUC.
```
Flowdalic, line 624: XEP-0045 uses the term "self-presence", hence I think this should also read self-presence (instead of just "presence).

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -683,6 +685,7 @@ public void join(Resourcepart nickname, String password) throws XMPPErrorExcepti
      * decide to create a new room or not.
      *
      * @param mucEnterConfiguration the configuration used to enter the MUC.
+     * @return the join presence as reflected by the MUC.
```
Flowdalic, line 688: XEP-0045 uses the term "self-presence", hence I think this should also read self-presence (instead of just "presence).

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -74,27 +75,69 @@ public MultiUserChatIntegrationTest(SmackIntegrationTestEnvironment environment)
         }
     }
 
+    /**
+     * Asserts that when a user joins a room, they are themselves included on the list of users notified (self-presence).
+     *
+     * <p>From XEP-0045 § 7.2.2:</p>
+     * <blockquote>
+     * ...the service MUST also send presence from the new participant's occupant JID to the full JIDs of all the occupants (including the new occupant)
```
Flowdalic, line 83: That line appears overly long, could you add a line break after 120 chars?

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -74,27 +75,69 @@ public MultiUserChatIntegrationTest(SmackIntegrationTestEnvironment environment)
...
+            tryDestroy(muc);
+        }
+    }
+
+    /**
+     * Asserts that when a user leaves a room, they are themselves included on the list of users notified (self-presence).
+     *
+     * <p>From XEP-0045 § 7.14:</p>
+     * <blockquote>
+     * The service MUST then send a presence stanzas of type "unavailable" from the departing user's occupant JID to the departing occupant's full JIDs, including a status code of "110" to indicate that this notification is "self-presence"
```
Flowdalic, line 112: That line appears overly long, could you add a line break after 120 chars?

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -113,23 +156,514 @@ public void processMessage(Message message) {
...
+        MucCreateConfigFormHandle handle = mucAsSeenByOne.createOrJoin(Resourcepart.from("one-" + randomString));
+        try {
+            if (handle != null) {
+                handle.makeInstant();
+            }
+
+            final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+            mucAsSeenByTwo.join(nicknameTwo);
+
+            // This implicitly tests "The service MUST add the user to the moderator list and then inform the admin of success" in §9.6, since it'll throw on either an error IQ or on no response.
```
Flowdalic, line 211: That line appears overly long, could you add a line break after 120 chars?

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -113,23 +156,514 @@ public void processMessage(Message message) {
...
+     * Asserts that a user who is present when another user undergoes an affiliation change receives that change as a presence update
+     *
+     * <p>From XEP-0045 § 5.2.2:</p>
+     * <blockquote>
+     * ...a MUC service implementation MUST change the user's affiliation to reflect the change and communicate that to all occupants...
+     * </blockquote>
+     *
+     * <p>From XEP-0045 § 10.6:</p>
+     * <blockquote>
+     * If the user is in the room, the service MUST then send updated presence from this individual to all occupants, indicating the granting of admin status...
```
Flowdalic, line 524: That line appears overly long, could you add a line break after 120 chars?

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -113,23 +156,514 @@ public void processMessage(Message message) {
...
+     * Asserts that a user who is present when another user undergoes a role change receives that change as a presence update
+     *
+     * <p>From XEP-0045 § 5.2.2:</p>
+     * <blockquote>
+     * ...a MUC service implementation MUST change the user's affiliation to reflect the change and communicate that to all occupants...
+     * </blockquote>
+     *
+     * <p>From XEP-0045 § 10.7:</p>
+     * <blockquote>
+     * If the user is in the room, the service MUST then send updated presence from this individual to all occupants, indicating the loss of admin status by sending a presence element...
```
Flowdalic, line 622: That line appears overly long, could you add a line break after 120 chars?

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -161,4 +695,29 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason) {
...
+        return JidCreate.entityBareFrom(Localpart.from(roomNameLocal), mucService.getDomain());
+    }
+
+    /**
+     * Destroys a MUC room, ignoring any exceptions.
+     *
+     * @param muc The room to destroy (can be null).
+     */
+    static void tryDestroy(final MultiUserChat muc) {
+        try {
```
Flowdalic, line 715: Try to keep the try-block as small as possible, never swallow exceptions, and never catch `Exception`, instead try to catch as less exception types as possible. This should become

```java
if (muc == null) {
    return;
}
try {
    muc.destory("test fixture teardown", null);
} catch (XMPPErrorException | NoResponseException | NotConnectedException | InterruptedException e) {
    LOGGER.log(Level.FINE, "Exception while destroing MUC", e);
}
```

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -113,23 +156,514 @@ public void processMessage(Message message) {
...
+     * Asserts that a user who undergoes a role change receives that change as a presence update
+     *
+     * <p>From XEP-0045 § 5.2.2:</p>
+     * <blockquote>
+     * ...a MUC service implementation MUST change the user's affiliation to reflect the change and communicate that to all occupants...
+     * </blockquote>
+     *
+     * <p>From XEP-0045 § 10.7:</p>
+     * <blockquote>
+     * If the user is in the room, the service MUST then send updated presence from this individual to all occupants, indicating the loss of admin status by sending a presence element...
```
Flowdalic, line 574: Overly long line.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -161,4 +695,29 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason) {
...
         assertNull(muc.getNickname());
     }
+
+    /**
+     * Gets a random room name
+     *
+     * @param prefix A prefix to add to the room name for descriptive purposes
+     */
+    private EntityBareJid getRandomRoom(String prefix) throws XmppStringprepException {
+        final String roomNameLocal = prefix + StringUtils.insecureRandomString(6);
```
Flowdalic, line 705: I know that this was not strictly introduced by you, but while at it: Objects created during a sinttest run should carry the `testRunId`, which is a field of `AbstractSmackIntTest`. Hence this should probably be

```
final String roomNameLocal = prefix + "-" + testRunId + "-" StringUtils.insecureRandomString(6);
```

Note that I added the dash after prefix, which means call sites of this function must remove it. For example
`getRandomRoom("smack-inttest-join-leave-")`
becomes
`getRandomRoom("smack-inttest-join-leave")`

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -113,23 +156,514 @@ public void processMessage(Message message) {
         });
 
         MucCreateConfigFormHandle handle = mucAsSeenByOne.createOrJoin(Resourcepart.from("one-" + randomString));
-        if (handle != null) {
-            handle.makeInstant();
+        try {
+            if (handle != null) {
+                handle.makeInstant();
```
Flowdalic, line 161: I know this pattern was used before, but since this PR repeats this pattern I wonder if it is actually a good idea. All (or maybe just most?) MUCs used by the integration test run should be freshly created, shouldn't they? So why do we use `MultiUserChat.createOrJoin()` when you could simply use `MulitUserChat.create()` and get rid of the `if (handle != null)`?

If you agree and are motivated to give it a shot, then please do. Otherwise we could look into it after your PR has been merged.

Flowdalic (changes requested):

The changes look good. Besides the minor nits mentioned in my inline comments I have to ask you for one more thing: Please rebase and squash this in a sensible manner. I think maybe two commits, one for the API changes and one for the integration test. But just one commit for all changes would also be acceptable.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: I avoided this pattern, since whilst it should work, it's semi-optimistic teardown code (in that we don't _really_ care if it succeeds or fails, it'd just be better if it did), and as such wanted to keep it clean of logging code and we don't care about what exception it throws, because that isn't the test. 

Is this a pattern you're really keen on starting?
Presently, there's only one other use of LOGGER in the tests.

Flowdalic: > Is this a pattern you're really keen on starting?

I think I've already use it everywhere in Smack.

> we don't care about what exception it throws, because that isn't the test.

Right now we don't care, hence my example put the log level at FINE. You can even make it FINER if you want. The point is, if we silently swallow exceptions, then we never have an option to observe those without an debugger. But if each and every exception is feed into the logging API, JUL in this case, then we can configure JUL so that it will show each and every exception if we ever need to. And by default, simply not show those exceptions: It would be the same as an empty `catch` block.

Fishbowler: I think my point is that these are tests, rather than regular code, and as such have a key objective to ensure the reader is aware of what's going on. Abstractions should be clean, flow should be linear and non-branching, and the intent always surfaced and clear and obvious. (Sure, some of these arguments could be made for "regular" code too)

Catching any exception is a indication to the reader that you don't care if this succeeds or fails.

Conversely, adding a logger is to say "someone could need to know about this", but in this instance they don't. Given the stuff in the first paragraph, tests shouldn't ever require debugging. 

I can though imagine a scenario where you're using these tests for granular observability of another system. That feels like a separate use case, where you might want more significant logging than just this one line, and are likely to simply fork the code in a throwaway branch with a plethora of log lines, be they LOGGER, System.out or other.

Fishbowler:
Rebased.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Flowdalic: > tests shouldn't ever require debugging.
But they do: Tests are code, code requires debugging sometimes

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: Tried this. Broke plenty. I've re-added it, but abstracted it into a method to make it read cleaner.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: To be debuggable there should be way more log lines throughout this test. I worry that adopting this as a pattern throughout future tests too might put off future contributions (although won't affect me - I have a wonderful employer paying me to do this!)

I maintain that quietly ignoring exceptions that aren't part of the important flow is a common pattern where IO/networking is involved
e.g. https://github.com/apache/commons-io/blob/commons-io-2.1/src/main/java/org/apache/commons/io/IOUtils.java#L311

Fishbowler: Or indeed https://github.com/igniterealtime/Smack/blob/master/smack-bosh/src/main/java/org/jivesoftware/smack/bosh/XMPPBOSHConnection.java#L344

Fishbowler: Maybe we could move debugging to a new issue in the tracker? I'd be interested to have a crack at it, and see what could be done to make tests both highly debuggable and highly readable at the same time. Perhaps there's a whole new pattern in test design to had from this.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -56,45 +55,91 @@
...
     private final MultiUserChatManager mucManagerOne;
     private final MultiUserChatManager mucManagerTwo;
+    private final MultiUserChatManager mucManagerThree;
     private final DomainBareJid mucService;
 
     public MultiUserChatIntegrationTest(SmackIntegrationTestEnvironment environment)
-                    throws NoResponseException, XMPPErrorException, NotConnectedException,
-                    InterruptedException, TestNotPossibleException {
+            throws NoResponseException, XMPPErrorException, NotConnectedException,
+            InterruptedException, TestNotPossibleException {
```
Flowdalic, line 63: This appears to be an unrelated whitespace change. If so, please revert that.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -56,45 +55,91 @@
...
         mucManagerOne = MultiUserChatManager.getInstanceFor(conOne);
         mucManagerTwo = MultiUserChatManager.getInstanceFor(conTwo);
+        mucManagerThree = MultiUserChatManager.getInstanceFor(conThree);
 
         List<DomainBareJid> services = mucManagerOne.getMucServiceDomains();
         if (services.isEmpty()) {
             throw new TestNotPossibleException("No MUC (XEP-45) service found");
-        }
-        else {
+        } else {
```
Flowdalic, line 72: This is an unrelated style change, please revert that.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
-        }
+        createMUC(mucAsSeenByOne, "one-" + randomString);
         mucAsSeenByTwo.join(Resourcepart.from("two-" + randomString));
-
         mucAsSeenByOne.sendMessage(mucMessage);
-        resultSyncPoint.waitForResult(timeout);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 166: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+
+        // This implicitly tests "The service MUST add the user to the moderator list and then inform the admin of
+        // success" in §9.6, since it'll throw on either an error IQ or on no response.
+        mucAsSeenByOne.grantModerator(nicknameTwo);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 215: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        createMUC(mucAsSeenByOne, "one-" + randomString);
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+
+        mucAsSeenByOne.grantModerator(nicknameTwo);
+        mucAsSeenByOne.revokeModerator(nicknameTwo);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 314: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        final Resourcepart nicknameThree = Resourcepart.from("three-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+        mucAsSeenByThree.join(nicknameThree);
+
+        mucAsSeenByOne.grantModerator(nicknameTwo);
+        mucAsSeenByOne.revokeModerator(nicknameTwo);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 365: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        });
+
+        createMUC(mucAsSeenByOne, "one-" + randomString);
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+        mucAsSeenByOne.revokeVoice(nicknameTwo);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 411: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        final Resourcepart nicknameThree = Resourcepart.from("three-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+        mucAsSeenByThree.join(nicknameThree);
+
+        mucAsSeenByOne.revokeVoice(nicknameTwo);
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 461: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        createMUC(mucAsSeenByOne, "one-" + randomString);
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+
+        // This implicitly tests "The service MUST add the user to the admin list and then inform the owner of success" in §10.6, since it'll throw on either an error IQ or on no response.
+        mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 510: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        final Resourcepart nicknameThree = Resourcepart.from("three-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+        mucAsSeenByThree.join(nicknameThree);
+
+        mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 561: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+
+        final Resourcepart nicknameTwo = Resourcepart.from("two-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+
+        mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
+        mucAsSeenByOne.revokeAdmin(conTwo.getUser().asEntityBareJid());
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 610: Always wrap the causing exception.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -112,24 +157,517 @@ public void processMessage(Message message) {
...
+        final Resourcepart nicknameThree = Resourcepart.from("three-" + randomString);
+        mucAsSeenByTwo.join(nicknameTwo);
+        mucAsSeenByThree.join(nicknameThree);
+        mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
+
+        mucAsSeenByOne.revokeAdmin(conTwo.getUser().asEntityBareJid());
+        try {
+            resultSyncPoint.waitForResult(timeout);
+        } catch (TimeoutException e) {
+            throw new AssertionError("Failed to receive presence");
```
Flowdalic, line 661: Always wrap the causing exception.

Come to think of it. The exception that is thrown because of the timeout should mark the test as failed and not abort the whole sinttest run. I am not sure if this is the case for `AssertionError`. In any way, why don't you simply *not* catch the `TimeoutException` and let it throw (I'd expect this exception type to mark the test as failed).

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -161,4 +699,37 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason) {
...
+     * @param muc The room to destroy (can be null).
+     */
+    static void tryDestroy(final MultiUserChat muc) {
+        if (muc == null) {
+            return;
+        }
+        try {
+            muc.destroy("test fixture teardown", null);
+        } catch (Exception ignore) {
+        }
```
Flowdalic, line 725: Always try to catch the most specific exception(s). Furthermore, no empty `catch` blocks. Either log the exceptions, with `FINEST` level if you want, or rethrow. I am currently updating ErrorProne in Smack, and with the new version, there the new [EmptyCatch bug pattern](https://errorprone.info/bugpattern/EmptyCatch) will be active.

Flowdalic (changes requested):

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: I've added the log line. Good call.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: I've removed the catch in hopes that it throws, although there's still a `finally` there too (which if I understand correctly means the cleanup will happen before the test bombs).

The AssertionError was because of this:
https://github.com/igniterealtime/Smack/blob/1df0763f92ce1775a1fb508701c3a9e5b91d2a1c/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java#L496-L500
Although on reading through this time, I'm not sure that it has any deep effect.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: Done

Fishbowler:
Think I've resolved all of the complaints.
I've retested this against a local Openfire running Monitoring, and they still all pass :)

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Flowdalic: This does not wrap the causing exception, i.e. `throw new AssertionError("Timeout waiting for result", e)`.

But after looking again at it, why do you catch it anyways? Simply let the method throw the `TimeoutException`. This is already done in the other test methods.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -161,4 +681,38 @@ public void roomDestroyed(MultiUserChat alternateMUC, String reason) {
...
+     * @param muc The room to destroy (can be null).
+     */
+    static void tryDestroy(final MultiUserChat muc) {
+        if (muc == null) {
+            return;
+        }
+        try {
+            muc.destroy("test fixture teardown", null);
+        } catch (Exception ignore) {
+            LOGGER.log(Level.FINEST, "Failed to tear down a randomly generated MUC (" + muc.getRoom().toString() + ") used as a test fixture. Continuing anyway.");
```
Flowdalic, line 707: Always pass the exception to the logging subsystem. And please don't call the `Exception` variable `ignore`.

LOGGER.log(Level.FINEST, "Failed to tear down a randomly generated MUC (" + muc.getRoom().toString() + ") used as a test fixture. Continuing anyway.", exception);

Flowdalic (changes requested):

Flowdalic:
Please also rebase on the lastest master while at it.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
Fishbowler: First exception has been wrapped. Second exception passed to logging. Thanks for the tips!

Fishbowler:
I've made the requested changes, and rebased 👍

Flowdalic:
You want to squash the last two fixup commits in the according commits to not clutter the git history.

Fishbowler:
Squashed and rebased 👍

PR #447, opened 2020-12-10
SINT: MAM test does not require XEP-0441

guusdk:
The MAM integration test setup attepts to set MAM preferences (XEP-0441).
When a server does not support this XEP, the setup phase errors out, preventing
the tests from being executed.

There is no functional reason why tests shouldn't be run when XEP-0441 is not
supported: the tests can run against the default configuration of the MAM
service.

This commit ignores any errors that occur when MAM preferences are set.


Flowdalic:
Thanks, applied with modifications as https://github.com/igniterealtime/Smack/commit/252cea1149c3967f57033f9c28efce5ff6fdcbce

PR #448, opened 2020-12-30
Initial support for XEP-0446: File Metadata Element

vanitasvitae:
Solves SMACK-894

[XEP-0446: File Metadata Element]() defines a generic extension element format that can be used to describe file metadata.
This PR adds initial support for this protocol by implementing the element and provider classes.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
@@ -1,0 +1,312 @@
+/**
+ *
+ * Copyright 2020 Paul Schaub
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.file_metadata.element;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jivesoftware.smack.packet.ExtensionElement;
+import org.jivesoftware.smack.packet.XmlEnvironment;
+import org.jivesoftware.smack.util.CollectionUtil;
+import org.jivesoftware.smack.util.EqualsUtil;
+import org.jivesoftware.smack.util.HashCode;
+import org.jivesoftware.smack.util.StringUtils;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+import org.jivesoftware.smackx.hashes.HashManager;
+import org.jivesoftware.smackx.hashes.element.HashElement;
+import org.jivesoftware.smackx.thumbnails.element.ThumbnailElement;
+
+/**
+ * File metadata element as defined in XEP-0446: File Metadata Element.
+ * This element is used in a generic way to provide information about files, e.g. during file sharing.
+ */
+public final class FileMetadataElement implements ExtensionElement {
+
+    public static final String ELEMENT = "file";
+    public static final String NAMESPACE = "urn:xmpp:file:metadata:0";
+    public static final String ELEM_DATE = "date";
+    public static final String ELEM_HEIGHT = "height";
+    public static final String ELEM_WIDTH = "width";
+    public static final String ELEM_DESC = "desc";
+    public static final String ELEM_LENGTH = "length";
+    public static final String ELEM_MEDIA_TYPE = "media-type";
+    public static final String ELEM_NAME = "name";
+    public static final String ELEM_SIZE = "size";
+
+
+    private final Date date;
+    private final Integer height;
+    private final Integer width;
+    private final Map<String, String> descriptions;
+    private final Map<HashManager.ALGORITHM, HashElement> hashElements;
+    private final Long length;
+    private final String mediaType;
+    private final String name;
+    private final Long size;
+    private final List<ThumbnailElement> thumbnails;
+
+    private FileMetadataElement(Date date, Integer height, Integer width, Map<String, String> descriptions,
+                               Map<HashManager.ALGORITHM, HashElement> hashElements, Long length,
+                               String mediaType, String name, Long size,
+                               List<ThumbnailElement> thumbnails) {
+        this.date = date;
+        this.height = height;
+        this.width = width;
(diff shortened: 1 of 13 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,262 @@
...
+ * File metadata element as defined in XEP-0446: File Metadata Element.
+ *
+ * This element is used in a generic way to provide information about files, eg. during file sharing.
+ */
+public final class FileMetadataElement implements ExtensionElement {
+
+    public static final String ELEMENT = "file";
+    public static final String NAMESPACE = "urn:xmpp:file:metadata:0";
+
+    private final Set<NamedElement> children = new HashSet<>();
```
Flowdalic, line 55: It is very uncommon that every child element gets its own class in such a case. XEP-0446 § 3 has a nice description of all child elements, which are all optional and, as far as I can tell, can only appear at most once. So why not simply create 8 member fields for those in `FileMetadataElement` (note that the situation may be different for `<thumbnail/>`? I think this would make the code much more readable, accessible and maintainable.

smack-experimental/src/test/java/org/jivesoftware/smackx/file_metadata/FileMetadataElementTest.java
```diff
@@ -0,0 +1,137 @@
...
+                "<hash xmlns='urn:xmpp:hashes:2' algo='sha-256'>2XarmwTlNxDAMkvymloX3S5+VbylNrJt/l5QyPa+YoU=</hash>" +
+                "<length>63000</length>" +
+                "<media-type>text/plain</media-type>" +
+                "<name>text.txt</name>" +
+                "<size>6144</size>" +
+                "</file>";
+        assertXmlSimilar(expectedXml, metadataElement.toXML().toString());
+
+        // Test deserialization
+        FileMetadataElement parsed = FileMetadataElementProvider.TEST_INSTANCE.parse(TestUtils.getParser(expectedXml));
```
Flowdalic, line 78: This should be split into two tests (serialization and deserialization) and use `@EnumSource(SmackTestUtil.XmlPullParserKind.class)` so that deserialization is performed with the two parsers available by Smack. See for example https://github.com/igniterealtime/Smack/blob/35a71f0131862b4cfdf3c4f5d31750245a3fc846/smack-extensions/src/test/java/org/jivesoftware/smackx/caps/provider/CapsExtensionProviderTest.java#L33

Flowdalic (changes requested):

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
vanitasvitae: Yeah I wasn't sure about the cardinality of the child elements. HashElement for example may occur multiple times.
But for name etc. I can definitely see your point.

smack-experimental/src/test/java/org/jivesoftware/smackx/file_metadata/FileMetadataElementTest.java
vanitasvitae: Ah right, I was aware of the parametrization, but somehow writing tests the old way was too much of a routine for me :D

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,308 @@
...
+    public Date getDate() {
+        return date;
+    }
+
+    public String getDimensions() {
+        return dimensions;
+    }
+
+    public Map<String, String> getDescriptions() {
+        return new ConcurrentHashMap<>(descriptions);
```
Flowdalic, line 121: Drop `ConcurrentHashMap` here and make `descriptions` an unmodifiable Map.

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,308 @@
...
+    public String getDescription() {
+        return getDescription(getLanguage());
+    }
+
+    public String getDescription(String lang) {
+        return descriptions.get(lang != null ? lang : "");
+    }
+
+    public Map<HashManager.ALGORITHM, HashElement> getHashElements() {
+        return new ConcurrentHashMap<>(hashElements);
```
Flowdalic, line 133: As with `descriptions`.

Flowdalic (changes requested):

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,308 @@
...
+    private final Long size;
+    private final Set<ExtensionElement> otherElements = new HashSet<>();
+
+    private FileMetadataElement(Date date, String dimensions, Map<String, String> descriptions,
+                               Map<HashManager.ALGORITHM, HashElement> hashElements, Long length,
+                               String mediaType, String name, Long size,
+                               Set<ExtensionElement> otherElements) {
+        this.date = date;
+        this.dimensions = dimensions;
+        this.descriptions.putAll(descriptions);
```
Flowdalic, line 71: In such cases, you may want to consider to use Smack's `CollectionUtil.cloneAndSeal()` helper method, to ensure that the build class is truly immutable

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,308 @@
...
+    public static final String ELEM_MEDIA_TYPE = "media-type";
+    public static final String ELEM_NAME = "name";
+    public static final String ELEM_SIZE = "size";
+    public static final String ELEM_THUMBNAIL = "thumbnail";
+
+
+    private final Date date;
+    private final String dimensions;
+    private final Map<String, String> descriptions = new ConcurrentHashMap<>();
+    private final Map<HashManager.ALGORITHM, HashElement> hashElements = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
vanitasvitae: I'm not sure I understand this comment. HashElement is already from Smacks XEP-0300 implementation.

smack-experimental/src/main/java/org/jivesoftware/smackx/file_metadata/element/FileMetadataElement.java
```diff
@@ -0,0 +1,308 @@
...
+                               Set<ExtensionElement> otherElements) {
+        this.date = date;
+        this.dimensions = dimensions;
+        this.descriptions.putAll(descriptions);
+        this.hashElements.putAll(hashElements);
+        this.length = length;
+        this.mediaType = mediaType;
+        this.name = name;
+        this.size = size;
+        this.otherElements.addAll(otherElements);
```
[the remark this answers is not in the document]
vanitasvitae: Implemented XEP-0264.

Flowdalic:
Thanks, please see also my follow-up commit https://github.com/igniterealtime/Smack/commit/b5180f819f422d3d6d3d6ff277f7fb8eee2ccbcc

PR #466, opened 2021-03-17
Added proxy support for XEP-0363

Vshnv:
HttpFileUploadManager did not use proxy info from connection to open the connection. Changed it so it uses the proxy info from the XMPP connection config to connect through the specified proxy where possible.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
@@ -16,6 +16,10 @@
  */
 package org.jivesoftware.smack.proxy;
 
+import java.net.InetSocketAddress;
+import java.net.Proxy;
+import java.net.SocketAddress;
+
 /**
  * Class which stores proxy information such as proxy type, host, port,
  * authentication etc.
@@ -44,19 +48,10 @@ public class ProxyInfo {
         this.proxyPort = pPort;
         this.proxyUsername = pUser;
         this.proxyPassword = pPass;
-        switch (proxyType) {
-        case HTTP:
-            proxySocketConnection = new HTTPProxySocketConnection(this);
-            break;
-        case SOCKS4:
-            proxySocketConnection = new Socks4ProxySocketConnection(this);
-            break;
-        case SOCKS5:
-            proxySocketConnection = new Socks5ProxySocketConnection(this);
-            break;
-        default:
-            throw new IllegalStateException();
-        }
+        this.proxySocketConnection =
+                ProxySocketConnection
+                        .forProxyType(proxyType)
+                        .apply(this);
     }
 
     public static ProxyInfo forHttpProxy(String pHost, int pPort, String pUser,
@@ -72,6 +67,18 @@ public class ProxyInfo {
     public static ProxyInfo forSocks5Proxy(String pHost, int pPort, String pUser,
                                     String pPass) {
         return new ProxyInfo(ProxyType.SOCKS5, pHost, pPort, pUser, pPass);
+    }
+
+    public Proxy.Type getJavaProxyType() {
+        switch (proxyType) {
+            case HTTP:
+                return Proxy.Type.HTTP;
+            case SOCKS4:
+            case SOCKS5:
+                return Proxy.Type.SOCKS;
+            default:
+                throw new AssertionError("Invalid proxy type: " + proxyType);
+        }
     }
 
     public ProxyType getProxyType() {
@@ -97,4 +104,11 @@ public class ProxyInfo {
     public ProxySocketConnection getProxySocketConnection() {
         return proxySocketConnection;
     }
+
+    public Proxy toJavaProxy() {
+        final SocketAddress proxySocketAddress = new InetSocketAddress(proxyAddress, proxyPort);
+        final Proxy.Type type = getJavaProxyType();
+        return new Proxy(type, proxySocketAddress);
+    }
+
 }
--- a/smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxySocketConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxySocketConnection.java
@@ -19,6 +19,8 @@ package org.jivesoftware.smack.proxy;
 import java.io.IOException;
 import java.net.Socket;
 
+import org.jivesoftware.smack.util.Function;
+
 public interface ProxySocketConnection {
 
     /**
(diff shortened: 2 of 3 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -502,6 +507,46 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
         }
     }
 
+    private static HttpURLConnection createURLConnection(XMPPConnection connection, URL putUrl) throws IOException {
+        Objects.requireNonNull(putUrl);
+        ProxyInfo proxyInfo = fetchProxyInfo(connection);
+        if (proxyInfo != null) {
+            return createProxiedURLConnection(proxyInfo, putUrl);
+        } else {
```
Flowdalic, line 515: Drop the `else` here, as you return in the `if` case.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -502,6 +507,46 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
...
+
+    private static Proxy.Type convertProxyType(ProxyInfo.ProxyType type) {
+        switch (type) {
+            case HTTP:
+                return Proxy.Type.HTTP;
+            case SOCKS4:
+            case SOCKS5:
+                return Proxy.Type.SOCKS;
+        }
+        return Proxy.Type.DIRECT;
```
Flowdalic, line 539: This can never happen, right? So I'd add a `default` case and throw an `AssertionError` in this case. I also think that this method should be added to `ProxyInfo`, e.g. as `ProxyInfo.getJavaNetProxyType()`.

Flowdalic (changes requested):
Thanks for your contribution. This appears to be a sensible addition. Please find my comments inline.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
Vshnv: Makes sense. Ill make the required changes and push in a bit.

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
```diff
@@ -24,17 +28,18 @@
  */
 
 public class ProxyInfo {
+
```
Flowdalic, line 31: Please do not include unrelated whitespace changes in a commit. If you really think there should be a newline, then create an extra commit and pull request for this.

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
```diff
@@ -24,17 +28,18 @@
...
         SOCKS4,
         SOCKS5
     }
 
-    private String proxyAddress;
-    private int proxyPort;
-    private String proxyUsername;
-    private String proxyPassword;
-    private ProxyType proxyType;
+    private final String proxyAddress;
```
Flowdalic, line 38: This and the following changes adding `final` seem very sensible, but should be an extra commit and pull request.

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxySocketConnection.java
```diff
@@ -34,4 +36,17 @@
...
+    static Function<ProxySocketConnection, ProxyInfo> forProxyType(ProxyInfo.ProxyType proxyType) {
+        switch (proxyType) {
+            case HTTP:
+                return HTTPProxySocketConnection::new;
+            case SOCKS4:
+                return Socks4ProxySocketConnection::new;
+            case SOCKS5:
+                return Socks5ProxySocketConnection::new;
+            default:
+                throw new IllegalStateException();
```
Flowdalic, line 49: If anything this should be an `AssertionError("Unknown proxy type: " + proxyType)`. But since you cover all cases of the`ProxyInfo.ProxyType` enum, do you need a default case at all?

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
```diff
@@ -74,6 +70,18 @@ public static ProxyInfo forSocks5Proxy(String pHost, int pPort, String pUser,
...
 
+    public Proxy.Type getJavaProxyType() {
+        switch (proxyType) {
+            case HTTP:
+                return Proxy.Type.HTTP;
+            case SOCKS4:
+            case SOCKS5:
+                return Proxy.Type.SOCKS;
+            default:
+                throw new AssertionError("Unreachable case.");
```
Flowdalic, line 81: Exception messages should always strive to include sensible context, e.g. in this case `"Unkown proxy type: "  + proxyType`. But since you cover all cases of theProxyInfo.ProxyType enum, do you need a default case at all?

Flowdalic (changes requested):
Thanks, this is clearly an improvement over the first version.

But please see if you can not rebase the PR into two commits: One that makes the `o.j.smack.proxy` related changes, and one that makes the `o.j.smackx.httpfileupload` related changes.

If I am not mistaken, then this would cleanly apply to the 4.4 branch, and also be a good candidate for the 4.4. branch. So maybe consider rebasing your commits onto the 4.4 branch and changing the target branch of this PR to 4.4. It should be trivial.

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
Vshnv: Ah, my bad. Did not notice the extra line. Reverting that

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxyInfo.java
Vshnv: Alrighty, will do

smack-core/src/main/java/org/jivesoftware/smack/proxy/ProxySocketConnection.java
Vshnv: Ah, I just move the IllegalStateException part from the other constructor as is, Ill make the change to an Assertion error. 

> do you need a default case at all?

Its just considered easier to understand exhaustive cases without going into the enum declaration with the explicit exception.

Flowdalic:
It appears your rebate went wrong, there should only be the two commits of you in the PR. Hint: maybe hard reset the source branch of this PR to 4.4, then cherry pick the two commits on top of it and then force push.

Vshnv:
Ah alright, thanks, was a bit stuck with that.

Vshnv:
Completed requested changes. Re-opening due to auto-close from the force-push.

Flowdalic:
Much better. Muchas gracias :)

PR #470, opened 2021-04-24
[sinttest] Tweaks to the muc test javadoc

Fishbowler:
* Fix 1 incorrect XEP reference
* Add 1 XEP reference for a previously undocumented test


Fishbowler:
Rebased.

PR #471, opened 2021-04-26
Unnecessary Exception is thrown when pub sub configure node is empty,…

annovanvliet:
… for instance in case of an auth error.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/provider/FormNodeProvider.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/provider/FormNodeProvider.java
@@ -21,7 +21,6 @@ import java.util.Map;
 
 import org.jivesoftware.smack.packet.ExtensionElement;
 import org.jivesoftware.smack.provider.EmbeddedExtensionProvider;
-
 import org.jivesoftware.smackx.pubsub.FormNode;
 import org.jivesoftware.smackx.pubsub.FormNodeType;
 import org.jivesoftware.smackx.xdata.packet.DataForm;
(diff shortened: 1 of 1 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/pubsub/provider/FormNodeProvider.java
```diff
@@ -35,6 +34,11 @@
 public class FormNodeProvider extends EmbeddedExtensionProvider<FormNode> {
     @Override
     protected FormNode createReturnExtension(String currentElement, String currentNamespace, Map<String, String> attributeMap, List<? extends ExtensionElement> content) {
-        return new FormNode(FormNodeType.valueOfFromElementName(currentElement, currentNamespace), attributeMap.get("node"), (DataForm) content.iterator().next());
+      
+        if ( !content.isEmpty() ) {
```
Flowdalic, line 38: "If content is empty, then return null" is easier to read.

Flowdalic (changes requested):
Can you share the stanza that triggers this? Also ideally add a test case for this.

annovanvliet:
An example is the following returned from OF:
<iq type="error" id="6LXNC-48" from="pubsub.openfire.xmpp.test" to="anno@openfire.xmpp.test/5dsi4g084a">
  <pubsub xmlns="http://jabber.org/protocol/pubsub#owner">
    <configure node="fdp/submitted/spot_report"/>
  </pubsub>
  <error code="403" type="auth">
    <forbidden xmlns="urn:ietf:params:xml:ns:xmpp-stanzas"/>
  </error>
</iq>

Flowdalic:
It appears to me the correct fix to this would be to allow `null` as `DataForm` argument to `FormNode`.

PR #472, opened 2021-05-10
MAM - added support for multiple versions of MAM

fenuz:
The versions of the MAM extension that a chat server supports are exposed as features.
Version 1 of the MAM is exposed as feature 'urn:xmpp:mam:1', and version 2 as feature 'urn:xmpp:mam:2'.

The v2 of the extension is largely identical to v1, except for a number of additional filtering options.
These filtering options are optional for clients, and are not implemented in the Smack implementation of MAM.

When we ignore the optional fields, the only difference between MAMv1 and MAMv2 is the use of a different
extension namespace (urn:xmpp:mam:2 instead of urn:xmpp:mam:1) and a different FORM_TYPE (again urn:xmpp:mam:2 instead of
urn:xmpp:mam:1).

Smack currently only supports MAMv2 with namespace urn:xmpp:mam:2.

This commit changes Smack to support multiple versions of MAM, and to do discovery of the version of MAM supported by
the archive.
The default behavior is to do no discovery, and to use MAMv2. The behavior and API is unchanged.
When MamManager.enableMamVersionDiscovery() is called, newly created MamManager instances will
discover which version of MAM is supported by the archive, and use the highest version supported.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
@@ -16,26 +16,10 @@
  */
 package org.jivesoftware.smackx.mam;
 
-import java.text.ParseException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.WeakHashMap;
-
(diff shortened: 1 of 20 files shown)


fenuz:
Hi, a comment on the PR would be much appreciated.

vanitasvitae:
I can't see any obvious problems with this PR :+1:.

Flowdalic:
Closing as I received no response to https://discourse.igniterealtime.org/t/mam-support-of-multiple-mam-versions/90136/3?u=flow

PR #475, opened 2021-06-15
[sinttest] Refactor MUC tests ready to add more

Fishbowler:
Relates to OF-2145, a desire to have more of the MUC capability of Openfire testable.

Following a conversation in open_chat, this is the first of a number of PRs to add more coverage of XEP-0045 to the Smack Integration Tests.

It splits some of the tests into a separate file, and makes both tests extend an Abstract class that contains utility methods. After that, it's tweaks and javadoc.


Fishbowler:
Realised I've left the createModeratedMUC method in AbstractMultiUserChatIntegrationTest that was done as part of testing my use cases against the refactor. It's a useful method, and about to be used, but happy to refactor it if it's a big problem to anyone.

PR #477, opened 2021-06-16
Additional tests covering s5 of XEP-0045

Fishbowler:
~Follows #475, so draft until that's complete.~

Tests passing against Openfire master.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
@@ -90,22 +90,22 @@ public class AbstractMultiUserChatIntegrationTest extends AbstractSmackIntegrati
         muc.destroy("test fixture teardown", null);
     }
 
-    static void createMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
-        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(Resourcepart.from(resourceName));
+    static void createMuc(MultiUserChat muc, Resourcepart resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException {
+        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(resourceName);
         if (handle != null) {
             handle.makeInstant();
         }
     }
 
-    static void createMembersOnlyMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.MucConfigurationNotSupportedException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
-        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(Resourcepart.from(resourceName));
+    static void createMembersOnlyMuc(MultiUserChat muc, Resourcepart resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.MucConfigurationNotSupportedException, MultiUserChatException.NotAMucServiceException {
+        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(resourceName);
         if (handle != null) {
             handle.getConfigFormManager().makeMembersOnly().submitConfigurationForm();
         }
     }
 
-    static void createModeratedMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
-        muc.create(Resourcepart.from(resourceName));
+    static void createModeratedMuc(MultiUserChat muc, Resourcepart resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException {
+        muc.create(resourceName);
         Form configForm = muc.getConfigurationForm();
         FillableForm answerForm = configForm.getFillableForm();
         answerForm.setAnswer("muc#roomconfig_moderatedroom", true); //TODO Add this to the MucConfigFormManager?
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
@@ -97,6 +97,13 @@ public class AbstractMultiUserChatIntegrationTest extends AbstractSmackIntegrati
         }
     }
 
+    static void createMembersOnlyMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.MucConfigurationNotSupportedException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
+        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(Resourcepart.from(resourceName));
+        if (handle != null) {
+            handle.getConfigFormManager().makeMembersOnly().submitConfigurationForm();
+        }
+    }
+
     static void createModeratedMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
         muc.create(Resourcepart.from(resourceName));
         Form configForm = muc.getConfigurationForm();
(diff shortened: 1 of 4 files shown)


smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
```diff
@@ -619,4 +621,295 @@ public void mucPresenceTestForWitnessingKick() throws Exception {
...
+                resultSyncPoint.signal("done");
+            }
+        });
+        mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
+        resultSyncPoint.waitForResult(timeout);
+
+        try {
+            assertEquals(mucAsSeenByOne.getOccupantsCount(), 3);
+            assertEquals(MUCRole.moderator, mucAsSeenByOne.getOccupant(
+                            JidCreate.entityFullFrom(mucAddress + "/" + nicknameOne.toString())).getRole());
```
Flowdalic, line 800: Use the `JidCreate.entityFullFrom(EntityBareJid, Resourcepart)` factory method here and everywhere else.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -97,6 +97,13 @@ static void createMuc(MultiUserChat muc, String resourceName) throws SmackExcept
         }
     }
 
+    static void createMembersOnlyMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.MucConfigurationNotSupportedException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
```
Flowdalic, line 100: The type of `resourceName` should be `Resourcepart`. This is also true for the `createModeratedMuc()` method below. It is called everywhere with a resourcepart that is transformed to a string, just to be re-transformed to a resourcepart within the method.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
```diff
@@ -619,4 +621,295 @@ public void mucPresenceTestForWitnessingKick() throws Exception {
...
+            assertEquals(MUCRole.moderator, mucAsSeenByOne.getOccupant(
+                            JidCreate.entityFullFrom(mucAddress + "/" + nicknameTwo.toString())).getRole());
+            assertEquals(MUCRole.visitor, mucAsSeenByOne.getOccupant(
```
Flowdalic, line 852: This is the only test that fails for me on ejabberd, as the role is 'participant' not 'visitor'. Had not yet the time to look into it.

Flowdalic (changes requested):

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
guusdk: Changed in 91897831dd520863fa5adf31565fb73aa836a78e

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
guusdk: Changed in 91897831dd520863fa5adf31565fb73aa836a78e

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
guusdk: In this implementation, the entity with nickname `nicknameThree` has joined a moderated room. 

Unlike the other two entities (`nicknameOne` created the room, `nicknameTwo` was granted Admin), no affiliation changes are applied to `nicknameThree`. The entity therefore doesn't have an affiliation.

According to [section 5.1.2, table 4](https://xmpp.org/extensions/xep-0045.html#table-4), the default role of such an entity SHOULD be `Visitor`. To me, it seems that this test asserts the correct status.

`Participant` would be appropriate for a similar scenario that plays out in an unmoderated room. That scenario is tested in a different test, `mucTestDefaultRoleForAffiliationInUnmoderatedRoom` (just above the one that we're discussing here).

Of note is that these tests assert a SHOULD definition. It's debatable if it warrants a test failure.

Flowdalic: I also think that this test asserts a SHOULD correctly. However ejabberd I my primary test platform. One way to move forward is to have a `static void setModeratedRoomDefaultRole(MUCRole mucRole)` method and a `static MUCRole moderedRoomDefaultRole = MUCRole.participant` field.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
```diff
@@ -619,4 +621,295 @@ public void mucPresenceTestForWitnessingKick() throws Exception {
...
+
+        final ResultSyncPoint<String, Exception> adminResultSyncPoint = new ResultSyncPoint<>();
+        mucAsSeenByOne.addParticipantStatusListener(new ParticipantStatusListener() {
+            @Override
+            public void adminGranted(EntityFullJid participant) {
+                adminResultSyncPoint.signal("done");
+            }
+        });
+
+        try {
```
Flowdalic, line 901: See my previous comment regarding the try block.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -90,15 +90,22 @@ static void tryDestroy(final MultiUserChat muc) throws SmackException.NoResponse
...
+        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(resourceName);
         if (handle != null) {
             handle.makeInstant();
         }
     }
 
-    static void createModeratedMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
-        muc.create(Resourcepart.from(resourceName));
+    static void createMembersOnlyMuc(MultiUserChat muc, Resourcepart resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.MucConfigurationNotSupportedException, MultiUserChatException.NotAMucServiceException {
```
Flowdalic, line 100: That has become a very long line. Please break at 120 chars.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -90,15 +90,22 @@ static void tryDestroy(final MultiUserChat muc) throws SmackException.NoResponse
         muc.destroy("test fixture teardown", null);
     }
 
-    static void createMuc(MultiUserChat muc, String resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, XmppStringprepException {
-        MultiUserChat.MucCreateConfigFormHandle handle = muc.create(Resourcepart.from(resourceName));
+    static void createMuc(MultiUserChat muc, Resourcepart resourceName) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException {
```
Flowdalic, line 93: Even though the line become shorter, it is still very long. Please break it at 120 chars.

Flowdalic (changes requested):

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
Fishbowler: How about exposing a sinttest setting for this, so that you can override it for ejabberd?

Flowdalic: I would usually prefer default compatibility with ejabberd than openfire. Simply because I use ejabberd as my primary and integration test server. However, I may be talked into defaulting to the XEPs SHOULD, if you create a new sinttest configuration setting that I can put into my `~/.config/smack-integration-test/properties`, e.g.

```
compatibilityMode=ejabberd
```

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatIntegrationTest.java
```diff
@@ -131,7 +131,7 @@ public void processMessage(Message message) {
             }
         });
 
-        createMuc(mucAsSeenByOne, "one-" + randomString);
+        createMuc(mucAsSeenByOne, Resourcepart.from("one-" + randomString));
```
Flowdalic, line 134: Since this is a re-occuring pattern, I think it would improve the readability of the tests if there was a 

```
public createMuc(MultiUserChat muc, String nick) {
   createMuc(muc, Resourcepart.from("one-" + randomString)
}
```
wrapper. This would also decrease the modified LoC count.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
Fishbowler: How do you feel about this? Feel more horizontally scalable:

```
    @SmackIntegrationTest
    public void mucTestDefaultRoleForAffiliationInUnmoderatedRoom() throws Exception {
        if(!environment.configuration.enableNonMandatoryXepChecks){
            return; //Don't test this
        }
```

Fishbowler: I also prefer this as it's more deterministic. A third XMPP server may have a third behaviour, and a fourth might have a fourth behaviour. Instead, this either enables or disables testing against the XEP's recommended behaviour.

Fishbowler: But... should this be a TestNotPossibleException? It seems to fit, except that it's not "NotPossible" so much as it is "NotDesirable" :)

Fishbowler:
Rebased

Flowdalic:
Thanks, applied with small modifications as https://github.com/igniterealtime/Smack/commit/48c057ab10e7f82a6af718c942863188c0b18437 and added a sinttest compatibility mode in https://github.com/igniterealtime/Smack/commit/a0b9279441d82747181e38847fab156f50868982

PR #480, opened 2021-06-25
[sint] increase stability

guusdk:
Some roster-based tests depend on there not being any prior subscription state beteween entities. The utility method that
tries to guarantee that, acts on the state of the roster that's cached in memory, but acts on the one that's stored on the
server. This occasionally causes issues, as both representations might be different.

Stability is added in this commit by:
- refreshing the roster from the server prior to evaluating it
- ignoring an 'item-not-found' as returned by the server, when the code tries to remove that item.

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/IntegrationTestRosterUtil.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/IntegrationTestRosterUtil.java
@@ -24,6 +24,7 @@ import org.jivesoftware.smack.SmackException.NotLoggedInException;
 import org.jivesoftware.smack.XMPPConnection;
 import org.jivesoftware.smack.XMPPException.XMPPErrorException;
 import org.jivesoftware.smack.packet.Presence;
+import org.jivesoftware.smack.packet.StanzaError;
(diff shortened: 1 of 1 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/IntegrationTestRosterUtil.java
```diff
@@ -95,11 +96,21 @@ public static void ensureBothAccountsAreNotInEachOthersRoster(XMPPConnection con
     private static void notInRoster(XMPPConnection c1, XMPPConnection c2) throws NotLoggedInException,
             NoResponseException, XMPPErrorException, NotConnectedException, InterruptedException {
         Roster roster = Roster.getInstanceFor(c1);
+        roster.reloadAndWait();
```
Flowdalic, line 99: That shouldn't be necessary, and if it is, then it's a bug (which we want to uncover with sinttest). Hence this would cover bugs.

guusdk: I'm not sure if I agree fully (a roster update might be in-flight), but it doesn't rule out that after the call, another update is in flight - so it presumably doesn't add much. I have removed the reload.

PR #481, opened 2021-06-30
[sint] new presence subscription tests

guusdk:
This adds a few basic tests that verify that presence subscription stanzas are received 'as intended', notably when they include extension elements.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/LowLevelSubscriptionIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/LowLevelSubscriptionIntegrationTest.java
@@ -1,6 +1,6 @@
 /**
  *
- * Copyright 2021-2023 Guus der Kinderen
+ * Copyright 2023 Guus der Kinderen
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
  */
 package org.jivesoftware.smack.subscription;
 
+import org.jivesoftware.smack.AbstractXMPPConnection;
 import org.jivesoftware.smack.filter.AndFilter;
 import org.jivesoftware.smack.filter.FromMatchesFilter;
 import org.jivesoftware.smack.filter.PresenceTypeFilter;
@@ -24,7 +25,7 @@ import org.jivesoftware.smack.filter.StanzaFilter;
 import org.jivesoftware.smack.packet.Presence;
 import org.jivesoftware.smack.packet.StandardExtensionElement;
 
-import org.igniterealtime.smack.inttest.AbstractSmackIntegrationTest;
+import org.igniterealtime.smack.inttest.AbstractSmackLowLevelIntegrationTest;
 import org.igniterealtime.smack.inttest.SmackIntegrationTestEnvironment;
 import org.igniterealtime.smack.inttest.annotations.SmackIntegrationTest;
 import org.igniterealtime.smack.inttest.util.IntegrationTestRosterUtil;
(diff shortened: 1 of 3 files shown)


guusdk:
After grudgingly applying two other 'improvements', this PR fails once again on a Checkstyle-based rule. I'm giving up.

smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/SubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,196 @@
...
+                .ofType(Presence.Type.subscribe)
+                .to(conOne.getUser())
+                .addExtension(new StandardExtensionElement("test", "org.example.test"))
+                .build();
+
+        final SimpleResultSyncPoint received = new SimpleResultSyncPoint();
+
+        conOne.addAsyncStanzaListener(p -> received.signal(),
+                stanza -> {
+                    if (!(stanza instanceof Presence)) {
```
Flowdalic, line 133: Use idiomatic Smack filters.

smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/SubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,196 @@
...
+                .build();
+
+        conTwo.sendStanza(subscriptionRequest);
+
+        ((AbstractXMPPConnection) conOne).connect();
+
+        final SimpleResultSyncPoint received = new SimpleResultSyncPoint();
+
+        conOne.addAsyncStanzaListener(p -> received.signal(),
+                stanza -> {
```
Flowdalic, line 178: Use idiomatic Smack filters.

smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/SubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,196 @@
...
+     * was made.
+     *
+     * @see <a href="https://igniterealtime.atlassian.net/browse/OF-2244">Openfire issue OF-2244</a>
+     * @throws Exception on anything unexpected or undesired.
+     */
+    @SmackIntegrationTest
+    public void testSubscriptionRequestOfflineWithExtension() throws Exception {
+        IntegrationTestRosterUtil.ensureBothAccountsAreNotInEachOthersRoster(conOne, conTwo);
+
+        ((AbstractXMPPConnection) conOne).disconnect();
```
Flowdalic, line 163: As above, this should probably be a low level integration test.

Flowdalic (changes requested):
Please find my review inline.

I think I've mentioned this before, but some of those tests are more towards testing server behavior and not really about integration tests. I guess a few simple ones are fine for Smack itself, but if you keep the integration tests coming that primarly test server behavior, then we really should think about establishing a separate xmpp-c2s-server-integration-test project where those tests reside in. Such a project can easily be based on, and exploit, Smack's integration test framework.

guusdk:
I've applied the feedback and rebased to the current HEAD of the Master branch for good measure.

Using a low-level test is less straight-forward if it's desirable to be able to make use of provisioned user accounts, or a mix of connected and unconnected connections. I've experimented with using `UnconnectedConnectionSource` and taking usernames from the configuration, but could not get that to work. As a result, this test still depends on two _connected_ resources, one of which is disconnected at the start of the tests. Those tests now _do_ inherit from Low-Level, which is at least something, I suppose.

smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/LowLevelSubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * This test verifies that a subscription request is received, in a scenario where the intended recipient was
+     * offline when the request was made.
+     *
+     * @param conOne Connection used to receive subscription request.
+     * @param conTwo Conenction used to send subscription request.
+     * @throws Exception on anything unexpected or undesired.
+     */
+    @SmackIntegrationTest
+    public void testSubscriptionRequestOffline(final AbstractXMPPConnection conOne,
```
Flowdalic, line 53: Please correct me if I am wrong (for this and all potential following statements), but this seems to test primarily a server functionality (i.e., that a presence subscription test send while the recipient is offline, is delivered once the recipient comes online).

As such, it is probably not a good fit for Smack and instead for smack-sinttest-server-extensions

smack-integration-test/src/main/java/org/jivesoftware/smack/subscription/LowLevelSubscriptionIntegrationTest.java
```diff
@@ -0,0 +1,121 @@
...
+     * such extension elements are received, in a scenario where the intended recipient was offline when the request
+     * was made.
+     *
+     * @param conOne Connection used to receive subscription request.
+     * @param conTwo Conenction used to send subscription request.
+     * @see <a href="https://igniterealtime.atlassian.net/browse/OF-2244">Openfire issue OF-2244</a>
+     * @throws Exception on anything unexpected or undesired.
+     */
+    @SmackIntegrationTest
+    public void testSubscriptionRequestOfflineWithExtension(final AbstractXMPPConnection conOne,
```
Flowdalic, line 92: Dito

Flowdalic (changes requested):
I think all of this belongs rather into smack-sint-server-extensions.

guusdk:
The tests to be added in this PR are now being added to the smack-sint-server-extensions project instead: https://github.com/XMPP-Interop-Testing/smack-sint-server-extensions/pull/6

PR #484, opened 2021-07-08
SMACK-908: Don't use components to count tabs in Debugger

guusdk:
This PR holds the same commit as the one in https://github.com/igniterealtime/Smack/pull/483, but cherry-picked against the 4.4 branch.


Flowdalic:
Please note that I usually avoid issue keys in the commit subject, as they are a waste of precious space in the subject. Especially if they are the first word of the subject.

PR #485, opened 2021-07-14
Bookmark parsing should not require non-empty nick

guusdk:
XEP-0048 defines an optional nickname for a bookmarked conference room, using an element named 'nick'.

Some clients have been observed to include an empty 'nick' element (instead of no such element) when a nickname is not stored as part of the bookmark.

Smack should not fail to parse these elements. Without the change in this commit, it fails with the following exception:

```
org.jxmpp.stringprep.XmppStringprepException: Argument can't be the empty string
    at org.jxmpp.stringprep.XmppStringPrepUtil.throwIfEmptyString(XmppStringPrepUtil.java:131)
    at org.jxmpp.stringprep.XmppStringPrepUtil.resourceprep(XmppStringPrepUtil.java:101)
    at org.jxmpp.jid.parts.Resourcepart.from(Resourcepart.java:91)
    at org.jivesoftware.smackx.bookmarks.Bookmarks.getConferenceStorage(Bookmarks.java:285)
    ...
```

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/bookmarks/Bookmarks.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/bookmarks/Bookmarks.java
@@ -282,7 +282,9 @@ public class Bookmarks implements PrivateData {
             XmlPullParser.Event eventType = parser.next();
             if (eventType == XmlPullParser.Event.START_ELEMENT && "nick".equals(parser.getName())) {
                 String nickString = parser.nextText();
-                conf.setNickname(Resourcepart.from(nickString));
+                if (nickString != null && !nickString.isEmpty()) {
+                    conf.setNickname(Resourcepart.from(nickString));
+                }
             }
             else if (eventType == XmlPullParser.Event.START_ELEMENT && "password".equals(parser.getName())) {
                 conf.setPassword(parser.nextText());


Flowdalic:
A bookmark with an `nick` element that contains the empty string is ambiguous: it could either mean that the client does not want a nick to be set, or that the client believes that the empty string is a valid nick (which it is not). Ambiguity is the enemy of interoperability: Someone send that you a message that could either mean this or that. If you are continue to operate after receiving such a message, chances are 50/50 that you pick the wrong one. And continuing from on a wrong assumption is deadly (ex falso sequitur quodlibet).

That said, thanks to Smack's modular architecture, users can easily replace the provider if they feel like must use a different one.

guusdk:
I disagree with Smack not accepting an empty `nick` element:
- I'd argue that there is no ambiguity since the 'empty-string nick' interpretation is clearly illegal.
- The XEP allows for an empty `nick` element. Smack should not reject valid stanzas.
- As for interoperability: _not_ allowing this hurts interoperability, as it breaks compatibility with implementations that adhere to the specification.

This is not a theoretical issue either: there is at least one client in the wild that add empty elements like this. They currently break Smack-based clients. This is especially visible and hard to work around for end users, since bookmarks are typically obtained and parsed upon login (to find auto-join bookmarked items).

Please reconsider merging this.

Flowdalic:
I thought about it and still stand by my points.

Try to fix the broken clients and/or use a custom parser if you are affected by this.

PR #487, opened 2021-07-26
MAM- added support for multiple versions of MAM

fenuz:
The versions of the MAM extension that a chat server supports are exposed as features.
Version 1 of the MAM is exposed as feature 'urn:xmpp:mam:1', and version 2 as feature 'urn:xmpp:mam:2'.

The v2 of the extension is largely identical to v1, except for a number of additional filtering options.
These filtering options are optional for clients, and are not implemented in the Smack implementation of MAM.

When we ignore the optional fields, the only difference between MAMv1 and MAMv2 is the use of a different
extension namespace (urn:xmpp:mam:2 instead of urn:xmpp:mam:1) and a different FORM_TYPE (again urn:xmpp:mam:2 instead of
urn:xmpp:mam:1).

Smack currently only supports MAMv2 with namespace urn:xmpp:mam:2.

This commit changes Smack to support multiple versions of MAM, and to do discovery of the version of MAM supported by
the archive.


smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamV1ElementFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.forward.packet.Forwarded;
+import org.jivesoftware.smackx.rsm.packet.RSMSet;
+import org.jivesoftware.smackx.xdata.packet.DataForm;
+
+import org.jxmpp.jid.Jid;
+
+public class MamV1ElementFactory implements MamElementFactory {
+
+    MamV1ElementFactory() {
```
Flowdalic, line 31: Shouldn't this constructor be declared `private`? or otherwise remove it.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamV2ElementFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.forward.packet.Forwarded;
+import org.jivesoftware.smackx.rsm.packet.RSMSet;
+import org.jivesoftware.smackx.xdata.packet.DataForm;
+
+import org.jxmpp.jid.Jid;
+
+public class MamV2ElementFactory implements MamElementFactory {
+
+    MamV2ElementFactory() {
```
Flowdalic, line 31: As with `MamV1ElementFactory`'s constructor.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -43,6 +45,12 @@
 
     @Override
     public MamPrefsIQ parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException {
+        MamElementFactory elementFactory = null;
+        try {
+            elementFactory = MamVersion.fromNamespace(parser.getNamespace()).newElementFactory();
+        } catch (IllegalArgumentException e) {
+            // do nothing, try to get the namespace from the parser later
```
Flowdalic, line 52: Can you elaborate why something in this `try` block would throw an IAE? I would have expected that `XmlPullParser.getNamespace()` at the starte of a `Provider`'s `parse()` method always returns a namespace. If so, then this should only throw unexpected exceptions (which we usually don't want to catch).

Also please consider only having those parts in the `try` block whos exceptions you intent to throw. For example: Can `parser.getNamespace()` be moved out of this `try` block, because we don't expect to throw it an IAE? What about `newElementFactory`?

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
```diff
@@ -63,6 +71,9 @@ public MamPrefsIQ parse(XmlPullParser parser, int initialDepth, XmlEnvironment x
             case START_ELEMENT:
                 final String name = parser.getName();
                 switch (name) {
+                case "prefs":
+                    elementFactory = MamVersion.fromNamespace(parser.getNamespace()).newElementFactory();
```
Flowdalic, line 75: As above, I am not sure, just from reading the code, why this is needed. Can you elaborate what's going on here? We don't do this dance in `MamQueryIQProvider` for example.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamResultProvider.java
```diff
@@ -43,6 +45,7 @@
     @Override
     public MamResultExtension parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment)
                     throws XmlPullParserException, IOException, SmackParsingException, ParseException {
+        MamElementFactory elementFactory = MamVersion.fromNamespace(parser.getNamespace()).newElementFactory();
```
Flowdalic, line 48: It looks like `MamVersion.fromNamespace(parser.getNamespace()).newElementFactory()` is a repeated pattern, which suggests what we may want to create a static convenience method like `MamElementFactory.for(XmlPullParser)`.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -250,6 +254,47 @@ public Jid getArchiveAddress() {
...
+     */
+    public String getMamNamespace() throws XMPPErrorException, NotConnectedException, NoResponseException, InterruptedException {
+        MamVersion mamVersion = getSupportedMamVersionOrNull();
+        return mamVersion == null ? null : mamVersion.getNamespace();
+    }
+
+    private MamVersion getSupportedMamVersionOrNull() throws XMPPErrorException, NotConnectedException, NoResponseException, InterruptedException {
+        if (mamVersion == null) {
+            DiscoverInfo info = serviceDiscoveryManager.discoverInfo(getArchiveAddress());
+            for (MamVersion v : MamVersion.values()) {
```
Flowdalic, line 275: Shouldn't we somehow ensure that we always detect the best, which usually means latest, version in case multiple are supported. I am not sure about the contract of an Enum's `values()`, but I think it could be implementation dependent. If so, then this would currently potentially yield different results in different Java runtime environments.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
```diff
@@ -250,6 +254,47 @@ public Jid getArchiveAddress() {
...
+     * @throws NotConnectedException if the XMPP connection is not connected.
+     * @throws InterruptedException if the calling thread was interrupted.
+     */
+    public String getMamNamespace() throws XMPPErrorException, NotConnectedException, NoResponseException, InterruptedException {
+        MamVersion mamVersion = getSupportedMamVersionOrNull();
+        return mamVersion == null ? null : mamVersion.getNamespace();
+    }
+
+    private MamVersion getSupportedMamVersionOrNull() throws XMPPErrorException, NotConnectedException, NoResponseException, InterruptedException {
+        if (mamVersion == null) {
```
Flowdalic, line 273: This should use the early return pattern to keep the level of indentation low.

Flowdalic (changes requested):
First late evening 20-minute review. Looks good in general :) I only left some comments that need to be addressed.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
fenuz: Good point, changed.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
fenuz: According to the Java Language Specification values() returns in the order the values were declared. So I think we can safely assume the order to be as we expect. See: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.3.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamResultProvider.java
fenuz: Good idea, changed.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
fenuz: When running the `MamPrefIQProviderTest` unit test, the initial `parser.getNamespace()` returns `null`, this causes the IAE to be thrown by `MamVersion.fromNamespace()`.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/provider/MamPrefsIQProvider.java
fenuz: As the parser initially returns `null` in the `MamPrefIQProviderTest` we call `getNamespace()` again when we encounter the `<prefs>` element.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamV1ElementFactory.java
fenuz: I changed the class visibility to package private and removed the (empty) constructor.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/element/MamV2ElementFactory.java
fenuz: I changed the class visibility to package private and removed the (empty) constructor.

smack-experimental/src/main/java/org/jivesoftware/smackx/mam/MamManager.java
fenuz: I did add some code comments to clarify this.

PR #489, opened 2021-09-07
Jingle and Jingle File Transfer

gkfabs:
This PR adds initial support (elements and providers) for XEP-0166 and XEP-0234

I took the work from https://github.com/vanitasvitae/Smack/tree/jftPR and made the modifications to the new code base.


Flowdalic (changes requested):
I've stopped the review half though, since this appears to be a half baked pull request (why isn't it marked WIP?). I think, if you want to continue working on Jingle support, then the initial work should be done outside of the Smack repository. Create an extra library with your code, that pulls in Smack as dependency. And, once the code is ready, it can be donated back to Smack.

gkfabs:
Thanks for running again the CI tools against my PR.

The 'TODO: Implement.' comments you spot in the code are wrong. The class are definitively implemented. I'll do the correction. I'll changes also to junit 5, the other comment readability problem and the logging.

If I do an extra library with this code, can I still use the same package names as yours ? ie org.jivesoftware.smackx ?

Flowdalic:
> If I do an extra library with this code, can I still use the same package names as yours ? ie org.jivesoftware.smackx ?

I'd suggest that you substitute `org.jivesoftware` for a domain under your control (or, if you don't have one, something else). Then, if the code gets donated back, we switch the classes back into a package under `org.jivesoftware`, while keeping the trailing package names. This way, users can simply switch by performing a global search and replace in their codebase.

PR #496, opened 2021-10-12
fix: Fixes processing destroy unavailable presence.

damencho:
Processes destroy event even when there is status in the unavailable presence stanza.


damencho:
Is it ok now :) ?

PR #503, opened 2021-10-20
Add missing stream namespace to xml declaration

ibauersachs:
Follow-up to PR #498

Sorry @Flowdalic that I didn't catch this in the first PR, I must have messed up the imported versions. Anyway, without this the parser instantiation fails with an unbounded prefix error message.


Flowdalic:
No worries. I'd like to avoid mentioning GitHub PR numbers in git commit messages. Could you instead simply add a 'Fixes:' git trailer? Thanks

PR #507, opened 2021-12-21
Increase performance of EnhancedDebugger

guusdk:
When using the Enhanced Debugger, a UI application is very noticably slow, especially around establishing a new
session. Profiling shows that much of the overhead is caused by XMPP-data being added to the text areas that are
on the "raw sent packets" and "raw received packets" tabs of the debugger.

This commit improves performance (considerably) by:
- properly limiting the amount of lines in those text areas (this was intended but broken in the old implementation)
- buffering data to be added in batches, to reduce the amount of invocations to JTextChat.append()

As an aside: some newline-based formatting was removed. As the provided data is now already formatted, retaining that
did not make much sense anymore.


guusdk:
This reduces the process of starting Spark and having everything logged in/initialized by roughly two minutes for me. Obviously, YMMV, depending on hardware and possibly OS and JRE version.

PR #511, opened 2022-02-03
Allow configurable ping timeout value

wferrell-ios:
Currently the minimum ping timeout value is hardcoded to 2 minutes. It would be nice to be able to configure this value. Using pingInterval as this is the lowest impact change to the code.


Flowdalic:
Thanks for your contribution.

Some connection kinds and network stack can have a large latency.  `PingManager` has a minimum ping timeout of 2 minutes to avoid false negatives (i.e., where the ping runs into a time out, even though a pong would arrive). If you really need a lower ping timeout, then the periodic checks of `PingManager` are not what you are looking for. Instead, you should implement your own logic.

PR #513, opened 2022-02-07
Conditionally reduce severity of roster reload error logging

maniac103:
If the roster feature is not supported by the server, there's no need to
log this at SEVERE log level.

### DIFF
--- a/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
+++ b/smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
@@ -469,11 +469,14 @@ public final class Roster extends Manager {
             @Override
             public void processException(Exception exception) {
                 rosterState = RosterState.uninitialized;
-                Level logLevel;
+                Level logLevel = Level.SEVERE;
                 if (exception instanceof NotConnectedException) {
                     logLevel = Level.FINE;
-                } else {
-                    logLevel = Level.SEVERE;
+                } else if (exception instanceof XMPPErrorException) {
+                    Condition condition = ((XMPPErrorException) exception).getStanzaError().getCondition();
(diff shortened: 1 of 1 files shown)


maniac103:
See https://github.com/openhab/openhab-addons/pull/12231#issuecomment-1031193915 for the rationale behind this.

smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
```diff
@@ -469,11 +470,14 @@ public void reload() throws NotLoggedInException, NotConnectedException, Interru
...
                 rosterState = RosterState.uninitialized;
-                Level logLevel;
+                Level logLevel = Level.SEVERE;
                 if (exception instanceof NotConnectedException) {
                     logLevel = Level.FINE;
-                } else {
-                    logLevel = Level.SEVERE;
+                } else if (exception instanceof XMPPErrorException) {
+                    StanzaError error = ((XMPPErrorException) exception).getStanzaError();
+                    if (error.getCondition() == Condition.feature_not_implemented) {
```
Flowdalic, line 478: It's a common error to assume that a server should return [feature-not-implemented](https://datatracker.ietf.org/doc/html/rfc6120#page-111) when a feature - here: the XMPP roster - is not implemented. However, if you carefully read RFC 6120, then you may notice that it probably should be [service-unavailable](https://datatracker.ietf.org/doc/html/rfc6120#page-121). I'd say we simply check for both.

Flowdalic (changes requested):

smack-im/src/main/java/org/jivesoftware/smack/roster/Roster.java
maniac103: Done.

PR #514, opened 2022-02-08
Add warning log to see what http connection is having trouble

Mtodiy:
During uploading some file with the HTTP File upload, there was some connection timeout that occurred. Afterward in the logging you are unable to see in the logging to what http connection the timeout occured. 

This pull request just adds one line of logging so you can better see to what http connection specifically is having problems.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/AbstractHttpUploadException.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/AbstractHttpUploadException.java
@@ -1,0 +1,101 @@
+/**
+ *
+ * Copyright 2022 Micha Kurvers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.httpfileupload;
+
+import java.io.IOException;
+import java.net.URL;
+
+import org.jivesoftware.smackx.httpfileupload.element.Slot;
+/**
+ * An exception class to provide additional information in case of exceptions during file uploading.
+ *
+ */
+public abstract class AbstractHttpUploadException extends IOException {
+
+    private static final long serialVersionUID = 1L;
+    private final long fileSize;
+    private final Slot slot;
+
+    protected AbstractHttpUploadException(long fileSize, Slot slot, String message) {
+        this(fileSize, slot, message, null);
+    }
+
+    protected AbstractHttpUploadException(long fileSize, Slot slot, String message, Throwable wrappedThrowable) {
+        super(message, wrappedThrowable);
+        this.fileSize = fileSize;
+        this.slot = slot;
+    }
+
+    public long getFileSize() {
+        return fileSize;
+    }
+
+    public URL getPutUrl() {
+        return slot.getPutUrl();
+    }
+
+    public Slot getSlot() {
+        return slot;
+    }
+
+    /**
+     * Exception thrown when http response returned after upload is not 200.
+     */
+    public static class HttpUploadErrorException extends AbstractHttpUploadException {
+
+        private static final long serialVersionUID = 8494356028399474995L;
+        private final int httpStatus;
+        private final String responseMsg;
+
+        public HttpUploadErrorException(int httpStatus, String responseMsg, long fileSize, Slot slot) {
+            super(fileSize, slot, "Error response " + httpStatus + " from server during file upload: "
+                            + responseMsg + ", file size: " + fileSize + ", put URL: "
+                            + slot.getPutUrl());
+            this.httpStatus = httpStatus;
+            this.responseMsg = responseMsg;
+        }
+
+        public int getHttpStatus() {
+            return httpStatus;
+        }
+
+        public String getResponseMsg() {
+            return responseMsg;
+        }
+
+    }
+
+    /**
+     * Exception thrown when an unexpected exception occurred during the upload.
+     */
+    public static class HttpUploadIOException extends AbstractHttpUploadException {
+
+        private static final long serialVersionUID = 5940866318073349451L;
+        private final IOException wrappedIOException;
+
+        public HttpUploadIOException(long fileSize, Slot slot, IOException cause) {
+            super(fileSize, slot, "Unexpected error occurred during file upload, file size: " + fileSize
+                            + ", put Url: " + slot.getPutUrl(), cause);
+            this.wrappedIOException = cause;
+        }
+
+        public IOException getCausingIOException() {
+            return this.wrappedIOException;
+        }
+
+    }
+}
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
@@ -49,6 +49,8 @@ import org.jivesoftware.smack.proxy.ProxyInfo;
 
 import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
 import org.jivesoftware.smackx.disco.packet.DiscoverInfo;
+import org.jivesoftware.smackx.httpfileupload.AbstractHttpUploadException.HttpUploadErrorException;
+import org.jivesoftware.smackx.httpfileupload.AbstractHttpUploadException.HttpUploadIOException;
 import org.jivesoftware.smackx.httpfileupload.UploadService.Version;
 import org.jivesoftware.smackx.httpfileupload.element.Slot;
 import org.jivesoftware.smackx.httpfileupload.element.SlotRequest;
@@ -494,10 +496,11 @@ public final class HttpFileUploadManager extends Manager {
             case HttpURLConnection.HTTP_NO_CONTENT:
                 break;
             default:
-                throw new IOException("Error response " + status + " from server during file upload: "
-                                + urlConnection.getResponseMessage() + ", file size: " + fileSize + ", put URL: "
-                                + putUrl);
+                throw new HttpUploadErrorException(status, urlConnection.getResponseMessage(), fileSize, slot);
             }
+        }
+        catch (IOException e) {
+            throw new HttpUploadIOException(fileSize, slot, e);
         }
         finally {
             urlConnection.disconnect();


smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -499,6 +499,9 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
                                 + putUrl);
             }
         }
+        catch (IOException e) {
+            LOGGER.log(Level.WARNING, String.format("Exception occurred while uploading to %s", putUrl.toString()));
```
Flowdalic, line 503: That would swallow the `IOException` and not throw it up to the caller. That does not appear to be sensible behavior for an library.

Mtodiy: You are right. Changed it to a simple catch rethrow, just adding the logging. I Amended the commit

Flowdalic: It is unusual to log an exception and throw it. I assume your motivation is to record `putUrl`, right? That's usually the callers job to log this when he catches the exception. Another option would be to add a specialized subclass of `IOException` that also contains the `putUrl`, say `HttpUploadIoExcetpion`. But I thing that would clutter the API with next to now benefit. But maybe I am misunderstand the problem you are trying to solve here?

Flowdalic: I just noticed that `putUrl` is not known to the caller. In that case, having an specialized exception `HttpUploadException` appears sensible. The exception should include `slot` and `fileSize`, as I think that is all contextual information that is sensible to include.

Mtodiy: Indeed, that was the motivation, since putUrl is never known to the caller in some setups it can be hard to debug what server we should investigate if the upload fails. 

I added the sensible custom exception., instead of rethrowing. Let me know what you think, cheers.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -494,11 +495,12 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
             case HttpURLConnection.HTTP_NO_CONTENT:
                 break;
             default:
-                throw new IOException("Error response " + status + " from server during file upload: "
-                                + urlConnection.getResponseMessage() + ", file size: " + fileSize + ", put URL: "
-                                + putUrl);
+                throw new HttpUploadException(status, urlConnection.getResponseMessage(), fileSize, putUrl);
```
Flowdalic, line 498: This should be an extra subclass of `HttpUploadException`, which itself, should become an abstract class.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/exception/HttpUploadException.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.net.URL;
+
+import org.jivesoftware.smackx.httpfileupload.element.Slot;
+
+/**
+ * An exception class to provide additional information in case of exceptions during file uploading.
+ *
+ */
+public class HttpUploadException extends IOException {
```
Flowdalic, line 29: Creating a new Java package for a single class seems like overkill. Just place it in o.j.smackx.httpfileupload. Also, what's the point if the constructors have parameters, but the arguments of the parameters are never accessible by the user? I think this class is missing some `private final` fields and the according getters.

Flowdalic (changes requested):

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/exception/HttpUploadException.java
Mtodiy: Fair enough, I made the class more accessible with appropriate getters

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
Mtodiy: Made the class abstract, and made two concrete and more specific exceptions

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -494,11 +497,12 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
             case HttpURLConnection.HTTP_NO_CONTENT:
                 break;
             default:
-                throw new IOException("Error response " + status + " from server during file upload: "
-                                + urlConnection.getResponseMessage() + ", file size: " + fileSize + ", put URL: "
-                                + putUrl);
+                throw new UnsuccessfulUploadException(status, urlConnection.getResponseMessage(), fileSize, putUrl);
```
Flowdalic, line 500: I am not super happy with the name of the exception, but you know: naming things is hard. But what about `HttpUploadErrorException`, since it contains an HTTP error code?

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -494,11 +497,12 @@ private void upload(InputStream iStream, long fileSize, Slot slot, UploadProgres
...
             }
         }
+        catch (IOException e) {
+            throw new UnexpectedUploadException(fileSize, slot, e);
```
Flowdalic, line 504: As above, I would name things exception `HttpUploadIoException`, since it basically wraps an `IOException`.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpUploadException.java
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Exception thrown when an unexpected exception occured during the upload.
+     */
+    public static class UnexpectedUploadException extends HttpUploadException {
+
+        private static final long serialVersionUID = 5940866318073349451L;
+        private final long fileSize;
+        private final Slot slot;
+
+        public UnexpectedUploadException(long fileSize, Slot slot, Throwable cause) {
```
Flowdalic, line 90: The `cause` parameter here should be of type `IOException`, since this is the only type we catch.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpUploadException.java
```diff
@@ -0,0 +1,106 @@
...
+
+
+        private static final long serialVersionUID = 8494356028399474995L;
+        private final int httpStatus;
+        private final String responseMsg;
+        private final long fileSize;
+        private final URL putUrl;
+
+        public UnsuccessfulUploadException(int httpStatus, String responseMsg, long fileSize, URL putUrl) {
+            super(String.format("Error response %d from server during file upload %s, file size: %d, put Url: %s",
```
Flowdalic, line 56: Note that this is a situation where you typically would not use `String.format()`, as you do not use any format specifiers. String concatenation using the `+` operator would just work fine here, would be more readable and more Java-idomatic.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpUploadException.java
```diff
@@ -0,0 +1,106 @@
...
+     * Exception thrown when an unexpected exception occured during the upload.
+     */
+    public static class UnexpectedUploadException extends HttpUploadException {
+
+        private static final long serialVersionUID = 5940866318073349451L;
+        private final long fileSize;
+        private final Slot slot;
+
+        public UnexpectedUploadException(long fileSize, Slot slot, Throwable cause) {
+            super(String.format("Unexpected error occurred during file upload, file size: %d, put Url: %s",
```
Flowdalic, line 91: See my remark about `String.format()` above.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpUploadException.java
```diff
@@ -0,0 +1,106 @@
...
+                    fileSize, slot.getPutUrl()), cause);
+            this.fileSize = fileSize;
+            this.slot = slot;
```
Flowdalic, line 94: I think it would be sensible to also save the causing `IOException` here, so we can provide a getter `IOException getCausingIoException()`, which returns the most narrow type: `IOException` in this case.

Flowdalic (changes requested):
We are getting close. Thanks for you work. Please find my remarks attached. Also note that I suggest that you squash the commits into one at the end.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpUploadException.java
```diff
@@ -0,0 +1,106 @@
...
+import java.net.URL;
+
+import org.jivesoftware.smackx.httpfileupload.element.Slot;
+
+
+/**
+ * An exception class to provide additional information in case of exceptions during file uploading.
+ *
+ */
+public abstract class HttpUploadException extends IOException {
```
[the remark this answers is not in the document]
Mtodiy: fair enough

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
Mtodiy: Sure, though I liked my functional descriptive ones, this is more factual indeed

Mtodiy:
Alright, made the amends. Looking forward to your feedback ;)

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/HttpFileUploadManager.java
```diff
@@ -57,7 +59,6 @@
 import org.jivesoftware.smackx.xdata.packet.DataForm;
 
 import org.jxmpp.jid.DomainBareJid;
-
```
Flowdalic, line 60: Unrelated whitespace change, please remove this.

Flowdalic (changes requested):
Some minor things left, then it's ready.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/AbstractHttpUploadException.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ */
+public abstract class AbstractHttpUploadException extends IOException {
+
+    private static final long serialVersionUID = 1L;
+    private final long fileSize;
+    private final Slot slot;
+
+    protected AbstractHttpUploadException(long fileSize, Slot slot, String message) {
+        super(message);
```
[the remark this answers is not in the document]
Mtodiy: resolved!

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/AbstractHttpUploadException.java
```diff
@@ -0,0 +1,103 @@
+/**
+ *
+ * Copyright the original author or authors
```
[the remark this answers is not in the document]
Mtodiy: Finally a place where I can legally just scribble my name onto!

Mtodiy:
Put the last changes in, cheers!

smack-experimental/src/main/java/org/jivesoftware/smackx/httpfileupload/AbstractHttpUploadException.java
```diff
@@ -0,0 +1,101 @@
+/**
+ *
+ * Copyright Micha Kurvers
```
Flowdalic, line 3: Please add the year to the copyright statement. See also the other copyright statements in Smack.

PR #515, opened 2022-02-12
Smack JingleReason#toXML must include both text and element

cmeng-git:
The JingleReason#toXML method must include both the optional text string and element XmlElement in the create process.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/jingle/element/JingleReason.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/jingle/element/JingleReason.java
@@ -132,7 +132,7 @@ public class JingleReason implements XmlElement {
      * @param reason the reason string that we'd like to transport in this packet extension, which may or
      * may not be one of the static strings defined here.
      * @param text an element providing human-readable information about the reason for the action or
-     * <tt>null</tt> if no such information is currently available.
+     * <code>null</code>  if no such information is currently available.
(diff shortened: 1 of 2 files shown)


cmeng-git:
I have already change <tt>null</tt> to <code>null</code> as per Smack check requirements. But still get rejected. Anything else I need to modifiy.

Flowdalic:
Thanks, from a quick look, there are at least two things you may want to change: First, this should target the 4.4 branch. Secondly, not all HTML tags are supported by javadoc. I the case where you used `tt`, you may want to use `{@link JingleReason}` instead.

smack-extensions/src/main/java/org/jivesoftware/smackx/jingle/element/JingleReason.java
```diff
@@ -155,6 +179,14 @@ public XmlStringBuilder toXML(XmlEnvironment enclosingXmlEnvironment) {
 
         xml.emptyElement(reason);
 
+        // add reason "text" if we have it
+        xml.optElement(TEXT_ELEMENT, text);
+
+        // add the extra element if it has been specified.
+        if (element != null) {
```
Flowdalic, line 186: This should be `xml.optAppend(element)` instead.

cmeng-git:
Closed pull request. Will try to do it in 4.4 branch

Flowdalic:
Please note that you don't have to close PRs to update the target branch.

cmeng-git:
Thank for the info. I am still learning all the github features.

PR #522, opened 2022-03-04
IQProvider correction

Neustradamus:
- Remove old IQProvider.java
- Rename IqProvider.java file to IQProvider.java
- Replace IqProvider -> IQProvider


Flowdalic:
`IqProvider` is determined to be the future abstract superclass for IQ providers, so renaming it to `IQProvider` is not an option. What somebody needs to do is to switch call usages of `IQProvider` to use `IqProvider` instead.

Neustradamus:
@Flowdalic: There is a problem not new:
- https://discourse.igniterealtime.org/t/problem-when-we-fork-the-github/87670

If you want I can rename the original IqProvider.java to  IqProvider2.java.

ibauersachs:
> What somebody needs to do is to switch call usages of `IQProvider` to use `IqProvider` instead.

@Flowdalic do you have a timeline for the 4.5 release? I might give this a try, but I'd need some time.

Flowdalic:
This PR is now obsolete due to #523.

PR #523, opened 2022-04-01
Replace legacy IQProvider with IqProvider

ibauersachs:
This is an attempt at solving the name conflict with the legacy `IQProvider` and the replacement `IqProvider`, see e.g. https://discourse.igniterealtime.org/t/iqprovider-similar-class-names-update-for-4-5/91469/2

All providers now use `IqProvider`, instead of the other way 'round as in #522. I looked at all `parse` methods to see if they move the parser to the end of the element, but lacking unit tests I'm not sure I really caught everything.

I was wondering if putting `ParserUtils.forwardToEndTagOfDepth(parser, initialDepth);` into `org.jivesoftware.smack.provider.IqProvider#parse(org.jivesoftware.smack.xml.XmlPullParser, org.jivesoftware.smack.packet.IqData, org.jivesoftware.smack.packet.XmlEnvironment)` just before the return would be an option? But I guess you had a reason not to do that in the first place.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/provider/BindIQProvider.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/provider/BindIQProvider.java
@@ -19,6 +19,7 @@ package org.jivesoftware.smack.provider;
 import java.io.IOException;
 
 import org.jivesoftware.smack.packet.Bind;
+import org.jivesoftware.smack.packet.IqData;
 import org.jivesoftware.smack.packet.XmlEnvironment;
 import org.jivesoftware.smack.xml.XmlPullParser;
 import org.jivesoftware.smack.xml.XmlPullParserException;
@@ -27,10 +28,10 @@ import org.jxmpp.jid.EntityFullJid;
 import org.jxmpp.jid.impl.JidCreate;
 import org.jxmpp.jid.parts.Resourcepart;
 
-public class BindIQProvider extends IQProvider<Bind> {
+public class BindIQProvider extends IqProvider<Bind> {
 
     @Override
-    public Bind parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException {
+    public Bind parse(XmlPullParser parser, int initialDepth, IqData iqData, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException {
         String name;
         Bind bind = null;
         outerloop: while (true) {
--- a/smack-core/src/main/java/org/jivesoftware/smack/provider/IQProviderInfo.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/provider/IQProviderInfo.java
@@ -28,7 +28,7 @@ import org.jivesoftware.smack.packet.IQ;
 public final class IQProviderInfo extends AbstractProviderInfo {
 
     /**
-     * Defines an IQ provider which implements the <code>IQProvider</code> interface.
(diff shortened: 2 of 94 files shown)


Flowdalic:
Thanks. +1

I did not had the time for a close look, but could you keep `IQProvider` and mark it deprecated with a TODO comment like
```
// TODO: Remove in Smack 4.6.
```
to give users a chance for a smooth transition and not have their projects fail to build at even more places if they switch to Smack 4.5.

ibauersachs:
Can we please agree on a compromise here? I've renamed `IQProvider` to `LegacyIQProvider` and added the deprecation notices. This way the git clone and build works again on Windows and Mac, and users that need to upgrade can run something along `sed -i s/IQProvider/LegacyIQProvider` before making the real migration.

damencho:
> Can we please agree on a compromise here? I've renamed `IQProvider` to `LegacyIQProvider` and added the deprecation notices. This way the git clone and build works again on Windows and Mac, and users that need to upgrade can run something along `sed -i s/IQProvider/LegacyIQProvider` before making the real migration.

👍

Flowdalic:
Well, I was once told at a certain open-source developers conference by someone that Smack to aggressively breaks API in newer versions (non-patchlevel versions, that is, patchlevel releases strive to be always API compatible and drop-in replacements).  :) That said, even though I personally would have kept `IQProvider` marked deprecated in Smack 4.5, I can live with your proposed compromise.

Flowdalic:
> But I guess you had a reason not to do that in the first place.

It is been a long time since I wrote that. I could imagine my thinking was that most providers would use the current idomatic parsing approach with `outerloop: while (true)` where the `break outerloop` is triggered by an end tag event of the correct depth. But even that, not having `IqProvider` forwarding the parser is merely a micro optimization. Hence I wouldn't mind if this was changed and this change results on less lines of code (which I assume it would).

ibauersachs:
> Well, I was once told at a certain open-source developers conference by someone that Smack to aggressively breaks API in newer versions (non-patchlevel versions, that is, patchlevel releases strive to be always API compatible and drop-in replacements). :) That said, even though I personally would have kept `IQProvider` marked deprecated in Smack 4.5, I can live with your proposed compromise.

I don't want to start a (new) SemVer debate, and in a normal case I'd fully agree to simply add `@deprecated`. Thanks for the compromise!

>> But I guess you had a reason not to do that in the first place.
>
> It is been a long time since I wrote that. I could imagine my thinking was that most providers would use the current idomatic parsing approach with outerloop: while (true) where the break outerloop is triggered by an end tag event of the correct depth. As a result, the superclass IqProivder would not need to forward the parser, as the parser would be already at the right position as per the XML Pull Parser calling convention. But even than, not having IqProvider forwarding the parser is merely a micro optimization. Hence I wouldn't mind if this was changed and this change results on less lines of code (which I assume it would).

I didn't count, but guessing from the changes yesterday more than half of the internal IqProviders don't use the `break outerloop` paradigm. So yes, there would be much less changes having the 'forwarder' in `IqProvider`. Plus, it'd be safer to merge since not all providers have (enough) unit tests to verify that my changes are correct.

skynet-gh:
Can we get a release with this change? I see `IQProvider.class` is still in 4.4.6, making it difficult to deal with on Windows with build systems that explode jars to the filesystem.

us4r93t3mnus:
> Can we get a release with this change? I see `IQProvider.class` is still in 4.4.6, making it difficult to deal with on Windows with build systems that explode jars to the filesystem.

Have you deal with this? Congratulations to the magician that created 2 classes with the same name!

PR #525, opened 2022-05-18
fix: Bosh configuration when bosh URI contains IP address.

damencho:
In a case where bosh url contains ip address in ConnectionConfiguration:
https://github.com/igniterealtime/Smack/blob/70abd8a18269c7fc0686ca75977fd48acf8d0a1b/smack-core/src/main/java/org/jivesoftware/smack/ConnectionConfiguration.java#L829
only hostAddress is set and host is null, which results an attempt to connect to https://null:443/http-bind.


Flowdalic:
Thanks for your contribution. :) This looks like a good candidate for the 4.4 branch of Smack. Would you mind updating the PR accordingly?

damencho:
Here it is: https://github.com/igniterealtime/Smack/pull/526

Flowdalic:
Merged #526

PR #535, opened 2022-08-01
Add partial support for XEP-0249 Direct MUC Invitations

MF1-MS:
This MR adds in partial support for [XEP-0249 Direct MUC Invitations](https://xmpp.org/extensions/xep-0249.html)

It exposes a method for a MUC to invite a user to the room, and adds a listener to the MultiUserChat listener to inform users of direct invitations they have received. 

A new UT class is added to test this function.

I found I could not build Smack without updating bouncy castle to 1.71 so have done so in this MR. Please advise how I would build without doing this if this is not desired.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
@@ -74,6 +74,7 @@ import org.jivesoftware.smackx.muc.MultiUserChatException.MucNotJoinedException;
 import org.jivesoftware.smackx.muc.MultiUserChatException.NotAMucServiceException;
 import org.jivesoftware.smackx.muc.filter.MUCUserStatusCodeFilter;
 import org.jivesoftware.smackx.muc.packet.Destroy;
+import org.jivesoftware.smackx.muc.packet.GroupChatInvitation;
 import org.jivesoftware.smackx.muc.packet.MUCAdmin;
 import org.jivesoftware.smackx.muc.packet.MUCInitialPresence;
 import org.jivesoftware.smackx.muc.packet.MUCItem;
@@ -1060,6 +1061,51 @@ public class MultiUserChat {
         messageBuilder.addExtension(mucUser);
 
         Message message = messageBuilder.build();
+        connection.sendStanza(message);
+    }
+
+    /**
+     * Invites another user to the room in which one is an occupant. In contrast
+     * to the method "invite", the invitation is sent directly to the user rather
+     * than via the chat room.  This is useful when the user being invited is
+     * offline, as otherwise the invitation would be dropped.
+     *
+     * @param address the user to send the invitation to
+     * @throws NotConnectedException if the XMPP connection is not connected.
+     * @throws InterruptedException if the calling thread was interrupted.
+     */
+    public void inviteDirectly(EntityBareJid address) throws NotConnectedException, InterruptedException {
+        inviteDirectly(address, null, null, false, null);
+    }
+
+     /**
+     * Invites another user to the room in which one is an occupant. In contrast
+     * to the method "invite", the invitation is sent directly to the user rather
+     * than via the chat room.  This is useful when the user being invited is
+     * offline, as otherwise the invitation would be dropped.
+     *
+     * @param address the user to send the invitation to
+     * @param reason the purpose for the invitation
+     * @param password specifies a password needed for entry
+     * @param continueAsOneToOneChat specifies if the groupchat room continues a one-to-one chat having the designated thread
+     * @param thread the thread to continue
+     * @throws NotConnectedException if the XMPP connection is not connected.
+     * @throws InterruptedException if the calling thread was interrupted.
+     */
+    public void inviteDirectly(EntityBareJid address, String reason, String password, boolean continueAsOneToOneChat, String thread)
+        throws NotConnectedException, InterruptedException {
+        // Add the extension for direct invitation
+        GroupChatInvitation invitationExt = new GroupChatInvitation(room,
+                                                                    reason,
+                                                                    password,
+                                                                    continueAsOneToOneChat,
+                                                                    thread);
+
+        Message message = connection.getStanzaFactory().buildMessageStanza()
+                .to(address)
+                .addExtension(invitationExt)
+                .build();
+
         connection.sendStanza(message);
     }
 
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
@@ -40,6 +40,7 @@ import org.jivesoftware.smack.XMPPConnection;
 import org.jivesoftware.smack.XMPPConnectionRegistry;
 import org.jivesoftware.smack.XMPPException.XMPPErrorException;
 import org.jivesoftware.smack.filter.AndFilter;
+import org.jivesoftware.smack.filter.ExtensionElementFilter;
 import org.jivesoftware.smack.filter.MessageTypeFilter;
 import org.jivesoftware.smack.filter.NotFilter;
 import org.jivesoftware.smack.filter.StanzaExtensionFilter;
@@ -57,6 +58,7 @@ import org.jivesoftware.smackx.disco.packet.DiscoverInfo;
 import org.jivesoftware.smackx.disco.packet.DiscoverItems;
 import org.jivesoftware.smackx.muc.MultiUserChatException.MucNotJoinedException;
 import org.jivesoftware.smackx.muc.MultiUserChatException.NotAMucServiceException;
+import org.jivesoftware.smackx.muc.packet.GroupChatInvitation;
 import org.jivesoftware.smackx.muc.packet.MUCInitialPresence;
 import org.jivesoftware.smackx.muc.packet.MUCUser;
 
@@ -139,6 +141,11 @@ public final class MultiUserChatManager extends Manager {
     private static final StanzaFilter INVITATION_FILTER = new AndFilter(StanzaTypeFilter.MESSAGE, new StanzaExtensionFilter(new MUCUser()),
                     new NotFilter(MessageTypeFilter.ERROR));
 
+    private static final StanzaFilter DIRECT_INVITATION_FILTER =
+        new AndFilter(StanzaTypeFilter.MESSAGE,
+                      new ExtensionElementFilter<GroupChatInvitation>(GroupChatInvitation.class),
+                      new NotFilter(MessageTypeFilter.ERROR));
+
     private static final ExpirationCache<DomainBareJid, DiscoverInfo> KNOWN_MUC_SERVICES = new ExpirationCache<>(
         100, 1000 * 60 * 60 * 24);
 
@@ -198,6 +205,33 @@ public final class MultiUserChatManager extends Manager {
             }
         };
         connection.addAsyncStanzaListener(invitationPacketListener, INVITATION_FILTER);
+
+        // Listens for all messages that include an XEP-0249 GroupChatInvitation extension and fire the invitation
+        // listeners
+        StanzaListener directInvitationStanzaListener = new StanzaListener() {
+            @Override
+            public void processStanza(Stanza stanza) {
+                final Message message = (Message) stanza;
+                GroupChatInvitation invite =
+                    stanza.getExtension(GroupChatInvitation.class);
+
+                // Fire event for invitation listeners
+                final MultiUserChat muc = getMultiUserChat(invite.getRoomAddress());
+                final XMPPConnection connection = connection();
+                final EntityJid from = message.getFrom().asEntityJidIfPossible();
+                if (from == null) {
+                    LOGGER.warning("Group Chat Invitation from non entity JID in '" + message + "'");
+                    return;
+                }
+                final String reason = invite.getReason();
+                final String password = invite.getPassword();
+                final MUCUser.Invite mucInvite = new MUCUser.Invite(reason, from, connection.getUser().asEntityBareJid());
+                for (final InvitationListener listener : invitationsListeners) {
+                    listener.invitationReceived(connection, muc, from, reason, password, message, mucInvite);
+                }
+            }
+        };
+        connection.addAsyncStanzaListener(directInvitationStanzaListener, DIRECT_INVITATION_FILTER);
 
         connection.addConnectionListener(new ConnectionListener() {
             @Override
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
@@ -69,6 +69,10 @@ public class GroupChatInvitation implements ExtensionElement {
     public static final QName QNAME = new QName(NAMESPACE, ELEMENT);
 
     private final EntityBareJid roomAddress;
+    private final String reason;
+    private final String password;
+    private final String thread;
+    private final boolean continueAsOneToOneChat;
 
     /**
      * Creates a new group chat invitation to the specified room address.
@@ -79,7 +83,67 @@ public class GroupChatInvitation implements ExtensionElement {
      * @param roomAddress the address of the group chat room.
      */
     public GroupChatInvitation(EntityBareJid roomAddress) {
+        this(roomAddress, null, null, false, null);
+    }
+
+    /**
+     * Creates a new group chat invitation to the specified room address.
+     * GroupChat room addresses are in the form <code>room@service</code>,
+     * where <code>service</code> is the name of group chat server, such as
+     * <code>chat.example.com</code>.
+     *
+     * @param roomAddress the address of the group chat room.
+     * @param reason the purpose for the invitation
+     * @param password specifies a password needed for entry
+     * @param continueAsOneToOneChat specifies if the groupchat room continues a one-to-one chat having the designated thread
+     * @param thread the thread to continue
+     */
+    public GroupChatInvitation(EntityBareJid roomAddress,
+                               String reason,
+                               String password,
+                               boolean continueAsOneToOneChat,
+                               String thread) {
         this.roomAddress = Objects.requireNonNull(roomAddress);
+        this.reason = reason;
+        this.password = password;
+        this.continueAsOneToOneChat = continueAsOneToOneChat;
+        this.thread = thread;
+    }
+
+    /**
+     * Returns the purpose for the invitation.
+     *
+     * @return the address of the group chat room.
+     */
+    public String getReason() {
+        return reason;
+    }
+
+    /**
+     * Returns the password needed for entry.
+     *
+     * @return the password needed for entry
+     */
+    public String getPassword() {
+        return password;
+    }
+
+    /**
+     * Returns the thread to continue.
+     *
+     * @return the thread to continue.
+     */
+    public String getThread() {
+        return thread;
+    }
+
+    /**
+     * Returns whether the groupchat room continues a one-to-one chat.
+     *
+     * @return whether the groupchat room continues a one-to-one chat.
+     */
+    public boolean continueAsOneToOneChat() {
+        return continueAsOneToOneChat;
     }
 
     /**
@@ -107,6 +171,13 @@ public class GroupChatInvitation implements ExtensionElement {
     public XmlStringBuilder toXML(org.jivesoftware.smack.packet.XmlEnvironment enclosingNamespace) {
         XmlStringBuilder xml = new XmlStringBuilder(this);
         xml.attribute("jid", getRoomAddress());
+        xml.optAttribute("reason", getReason());
+        xml.optAttribute("password", getPassword());
+        xml.optAttribute("thread", getThread());
+
+        if (continueAsOneToOneChat())
+            xml.optBooleanAttribute("continue", true);
+
         xml.closeEmptyElement();
         return xml;
     }
--- a/smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
+++ b/smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
@@ -1,0 +1,324 @@
+/**
+ *
+ * Copyright 2021-2022 Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.muc;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.jivesoftware.smack.DummyConnection;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.packet.Stanza;
+import org.jivesoftware.smack.test.util.SmackTestSuite;
+import org.jivesoftware.smack.test.util.WaitForPacketListener;
+import org.jivesoftware.smackx.muc.packet.GroupChatInvitation;
+import org.jivesoftware.smackx.muc.packet.MUCUser;
+import org.jivesoftware.smackx.muc.packet.MUCUser.Invite;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.jxmpp.jid.EntityBareJid;
+import org.jxmpp.jid.EntityFullJid;
+import org.jxmpp.jid.EntityJid;
+import org.jxmpp.jid.impl.JidCreate;
+import org.jxmpp.stringprep.XmppStringprepException;
+
+/**
+ * A test for for following features:
+ *    <li>Adds support for Direct MUC invitations (see <a href="https://xmpp.org/extensions/xep-0249.html">XEP-0249</a>), which allows offline users to be invited to group chats.</li>
+ * </ul>
+ */
+public class MultiUserChatTest extends SmackTestSuite {
+    private static final int RESPONSE_TIMEOUT_IN_MILLIS = 10000;
+
+    private DummyConnection connection;
+    private MultiUserChatManager multiUserChatManager;
+
+    @BeforeEach
+    public void setUp() throws Exception {
+        connection = new DummyConnection();
+        connection.connect();
+        connection.login();
+
+        multiUserChatManager = MultiUserChatManager.getInstanceFor(connection);
+    }
+
+    @AfterEach
+    public void tearDown() {
+        if (connection != null) {
+            connection.disconnect();
+            connection = null;
+        }
+    }
+
+    @Test
+    public void testInviteDirectly() throws Throwable {
+        EntityBareJid roomJid = JidCreate.entityBareFrom("room@example.com");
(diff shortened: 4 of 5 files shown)


smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -254,6 +264,44 @@ public void run() {
...
+        StanzaListener xep0249InvitationPacketListener = new StanzaListener() {
+            @Override
+            public void processStanza(Stanza packet) {
+                final Message message = (Message) packet;
+                ExtensionElement inviteExt =
+                    packet.getExtension(GroupChatInvitation.class);
+                GroupChatInvitation invite =
+                    inviteExt instanceof GroupChatInvitation ? (GroupChatInvitation) inviteExt : null;
+
+                if (invite != null && invite.getRoomAddress() != null) {
```
Flowdalic, line 279: Those checks are not needed. The `invite.getRoomAddress() != null` check is unnecessary,because the provider will not create GroupChatInviation instances without a room address, because the according element is required to have a 'jid' attribute.

Flowdalic (changes requested):
Thanks for your contribution. I have left some remarks inline.

I think the most important issue that this is currently somehow entangled with some arbitrary `MultiUserChat` instance, when its business logic should really go into a new class `DirectMucInviteManager`.

Flowdalic:
> I found I could not build Smack without updating bouncy castle to 1.71 so have done so in this MR. Please advise how I would build without doing this if this is not desired.

Bumping BC is fine. Since you give not details why you couldn't build without updating, I can't really help you there.

build.gradle
```diff
@@ -148,7 +148,7 @@ allprojects {
 		smackMinAndroidSdk = 19
 		junitVersion = '5.7.1'
 		commonsIoVersion = '2.6'
-		bouncyCastleVersion = '1.70'
+		bouncyCastleVersion = '1.71'
```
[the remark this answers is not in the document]
MF1-MS: When I try to build/test Smack at 1.70 at get the following stacktrace. The method it is trying to access was introduced in BC 1.71. 

```
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details

> Task :smack-openpgp:test

SecretKeyBackupHelperTest > createAndDecryptSecretKeyElementTest() FAILED
    java.lang.NoSuchMethodError: 'org.bouncycastle.asn1.x9.X9ECParametersHolder org.bouncycastle.asn1.x9.ECNamedCurveTable.getByOIDLazy(org.bouncycastle.asn1.ASN1ObjectIdentifier)'
        at org.bouncycastle.openpgp.PGPPublicKey.init(Unknown Source)
        at org.bouncycastle.openpgp.PGPPublicKey.<init>(Unknown Source)
        at org.bouncycastle.openpgp.PGPSecretKeyRing.<init>(Unknown Source)
        at org.bouncycastle.openpgp.PGPObjectFactory.nextObject(Unknown Source)
        at org.pgpainless.key.parsing.KeyRingReader.readSecretKeyRingCollection(KeyRingReader.java:246)
        at org.pgpainless.key.parsing.KeyRingReader.readSecretKeyRingCollection(KeyRingReader.java:221)
        at org.pgpainless.key.parsing.KeyRingReader.secretKeyRingCollection(KeyRingReader.java:74)
        at org.jivesoftware.smackx.ox.store.filebased.FileBasedOpenPgpKeyStore.readSecretKeysOf(FileBasedOpenPgpKeyStore.java:124)
        at org.jivesoftware.smackx.ox.store.abstr.AbstractOpenPgpKeyStore.getSecretKeysOf(AbstractOpenPgpKeyStore.java:146)
        at org.jivesoftware.smackx.ox.store.abstr.AbstractOpenPgpKeyStore.importSecretKey(AbstractOpenPgpKeyStore.java:162)
        at org.jivesoftware.smackx.ox.store.abstr.AbstractOpenPgpStore.importSecretKey(AbstractOpenPgpStore.java:131)
        at org.jivesoftware.smackx.ox.SecretKeyBackupHelperTest.createAndDecryptSecretKeyElementTest(SecretKeyBackupHelperTest.java:83)

29 tests completed, 1 failed
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -139,6 +142,11 @@ public static synchronized MultiUserChatManager getInstanceFor(XMPPConnection co
     private static final StanzaFilter INVITATION_FILTER = new AndFilter(StanzaTypeFilter.MESSAGE, new StanzaExtensionFilter(new MUCUser()),
                     new NotFilter(MessageTypeFilter.ERROR));
 
+    private static final StanzaFilter XEP_0249_INVITATION_FILTER =
+        new AndFilter(StanzaTypeFilter.MESSAGE,
+                      new StanzaExtensionFilter(GroupChatInvitation.ELEMENT, GroupChatInvitation.NAMESPACE),
```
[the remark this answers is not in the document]
MF1-MS: `ExtensionElementFilter` has a protected constructer and won't let me call it from here. Should it be public?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +207,8 @@ public void processStanza(Stanza packet) {
         };
         connection.addAsyncStanzaListener(invitationPacketListener, INVITATION_FILTER);
 
+        addXEP0249Listener(connection);
```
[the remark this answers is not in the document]
MF1-MS: The machinery to listen for invitations is not part of a `MultiUserChat` class, but part of the singleton `MultiUserChatManager` that manages a connection's MUCs. It has responsibility to listen for invitations and kick off having the desired XMPP client entity join the specified MUC room. 

The logic on a `MultiUserChat` instance scope is the ability to invite a client entity directly to that room, which also seems logical valid to me?

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -1063,6 +1065,29 @@ public void invite(MessageBuilder messageBuilder, EntityBareJid user, String rea
...
+     * Invites another user to the room in which one is an occupant. In contrast
+     * to the method "invite", the invitation is sent directly to the user rather
+     * than via the chat room.  This is useful when the user being invited is
+     * offline, as otherwise the invitation would be dropped.
+     *
+     * @param user the user to send the invitation to
+     * @throws NotConnectedException if the XMPP connection is not connected.
+     * @throws InterruptedException if the calling thread was interrupted.
+     */
+    public void inviteDirectly(EntityBareJid user) throws NotConnectedException, InterruptedException {
```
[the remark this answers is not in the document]
MF1-MS: True, I shall try to add soon.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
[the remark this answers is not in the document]
Flowdalic: It should be, fixed in master.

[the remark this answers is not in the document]
Flowdalic: It should, fixed in master.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,40 @@ public void processStanza(Stanza packet) {
...
+        // Listens for all messages that include an XEP-0249 GroupChatInvitation extension and fire the invitation
+        // listeners
+        StanzaListener directInvitationPacketListener = new StanzaListener() {
+            @Override
+            public void processStanza(Stanza packet) {
+                final Message message = (Message) packet;
+                GroupChatInvitation invite =
+                    packet.getExtension(GroupChatInvitation.class);
+
+                if (invite.getRoomAddress() != null) {
```
Flowdalic, line 218: This check is unnecessary, as it is impossible to create a `GroupChatInvitation` that hasn't set a room address.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,40 @@ public void processStanza(Stanza packet) {
...
+                    EntityBareJid mucJid = JidCreate.entityBareFromOrNull(invite.getRoomAddress());
+                    if (mucJid == null) {
```
Flowdalic, line 220: Those checks are now also unnecessary, as the room address is always an `EntityBareJid`.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,40 @@ public void processStanza(Stanza packet) {
...
+                        LOGGER.warning("Received invite to non bare JID: '" + message + "'");
+                        return;
+                    }
+                    // Fire event for invitation listeners
+                    final MultiUserChat muc = getMultiUserChat(mucJid);
+                    final XMPPConnection connection = connection();
+                    final MUCUser.Invite mucInvite = null;
+                    final EntityJid from = message.getFrom().asEntityJidIfPossible();
+                    if (from == null) {
+                        LOGGER.warning("Group Chat Invitation from non entity JID: '" + message.toXML() + "'");
```
Flowdalic, line 230: I am not super happy about that, as I believe it is likely that components, i.e., domain bare JIDs, could also send directed invites. I think we could change the signature of `InvitationsLIstener` to use just `Jid` for `inviter`.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,40 @@ public void processStanza(Stanza packet) {
...
+                if (invite.getRoomAddress() != null) {
+                    EntityBareJid mucJid = JidCreate.entityBareFromOrNull(invite.getRoomAddress());
+                    if (mucJid == null) {
+                        LOGGER.warning("Received invite to non bare JID: '" + message + "'");
+                        return;
+                    }
+                    // Fire event for invitation listeners
+                    final MultiUserChat muc = getMultiUserChat(mucJid);
+                    final XMPPConnection connection = connection();
+                    final MUCUser.Invite mucInvite = null;
```
Flowdalic, line 227: Now that is a problem: `invite` was previously always not null. This has the potential to break apps if they upgrade to a newer Smack version. I think you want to synthesize a `MUCUser.Invite` here.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,40 @@ public void processStanza(Stanza packet) {
...
+                    final EntityJid from = message.getFrom().asEntityJidIfPossible();
+                    if (from == null) {
+                        LOGGER.warning("Group Chat Invitation from non entity JID: '" + message.toXML() + "'");
+                        return;
+                    }
+                    final String reason = null;
```
Flowdalic, line 233: You should obtain `reason` and `password` from the xep249 directed user invite.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    @Test
+    public void shouldReceiveOfflineInvitation() throws XmppStringprepException {
+        EntityBareJid roomJid = JidCreate.entityBareFrom("room@example.com");
+        EntityFullJid inviterJid = JidCreate.entityFullFrom("inviter@example.com/user1");
+        EntityBareJid inviteeJid = JidCreate.entityBareFrom("invitee@example.com");
+
+        GroupChatInvitation groupChatInvitation = new GroupChatInvitation(JidCreate.entityBareFrom("room@example.com"));
+        Message sentMessage = MessageBuilder.buildMessage(UUID.randomUUID().toString())
```
Flowdalic, line 121:
```suggestion
        Message sentMessage = connection.getStanzaFactory().buildMessage()
```

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,211 @@
+/**
+ *
+ * Copyright 2021 Microsoft Corporation.
```
Flowdalic, line 3: Correct date? Should probably be at least 2021-2022.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,211 @@
...
+        assertTrue(updateRequestSent.get(), "Invite directly request not sent");
+    }
+
+    @Test
+    public void shouldReceiveOfflineInvitation() throws XmppStringprepException {
+        EntityBareJid roomJid = JidCreate.entityBareFrom("room@example.com");
+        EntityFullJid inviterJid = JidCreate.entityFullFrom("inviter@example.com/user1");
+        EntityBareJid inviteeJid = JidCreate.entityBareFrom("invitee@example.com");
+
+        GroupChatInvitation groupChatInvitation = new GroupChatInvitation(JidCreate.entityBareFrom("room@example.com"));
```
Flowdalic, line 120:
```suggestion
        GroupChatInvitation groupChatInvitation = new GroupChatInvitation(roomJid);
```

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,211 @@
...
+                .addExtension(groupChatInvitation)
+                .build();
+
+        // Prepare listener to receive a group invitation
+        GroupInvitationListener groupInvitationListener = new GroupInvitationListener() {
+            @Override
+            public void verifyInvitation(XMPPConnection conn, MultiUserChat room, EntityJid inviter,
+                                         String reason, String password, Message message, MUCUser.Invite invitation) {
+                // Check all parameters values.
+                assertSame(connection, conn);
```
Flowdalic, line 133: I fear those asserts won't work, as their exceptions will be not be thrown in the thread that invoked the unit test method.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+        // Simulate sending a message with a group invitation
+        connection.processStanza(sentMessage);
+
+        // Wait for the listener to be called or throw a timeout exception
+        groupInvitationListener.waitUntilInvocationOrTimeout();
+    }
+
+    private static class Responder extends Thread {
+        protected final AtomicBoolean running = new AtomicBoolean();
```
Flowdalic, line 153: This field isn't used anywhere, in fact the whole `Responder` class could be merged into `InvokeDirectlyResponder`. I assume this was potentially at some point in the past different, but right now it is not required in this form and hence only contributes to confusion, i.e., it makes the code harder to read.

Flowdalic (changes requested):

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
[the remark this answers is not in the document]
Flowdalic: My fault. I wasn't looking carefully enough and assumed this is in `MultiUserChat`.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
MF1-MS: I'm not sure I agree. The end user the invitation is from should be the same regardless of if they're using an XEP-0249 direct invitation or a mediated XEP-0045 invitation. If we are asserting a user must be an entity jid in the XEP-0045 case then I think it's very reasonable to assert the same here.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
MF1-MS: Hmmm you're right, it throws by then we have to wait for the latch to time out, and then we can check the stderr in the test report to diagnose the problem. Any suggestions?

```
> Task :smack-extensions:test

MultiUserChatTest > shouldReceiveOfflineInvitationWithAllOptionalAttributes() FAILED
    java.lang.IllegalStateException: Latch timed out before it reached zero
        at org.jivesoftware.smack.test.util.WaitForPacketListener.waitUntilInvocationOrTimeout(WaitForPacketListener.java:49)
        at org.jivesoftware.smackx.muc.MultiUserChatTest.shouldReceiveOfflineInvitationWithAllOptionalAttributes(MultiUserChatTest.java:248)
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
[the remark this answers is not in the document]
MF1-MS: Done.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
MF1-MS: Ah, I can have it save off an exception we check and rethrow in the main thread like the `InvokeDirectlyResponder`

MF1-MS:
OK sorry about the delay, that's amended now. I can squash the commits and amend the message if you're happy, but leaving as is for now for a cleaner diff. 

Are you happy with my other PRs too?:
https://github.com/igniterealtime/Smack/pull/537
https://github.com/igniterealtime/Smack/pull/538

build.gradle
```diff
@@ -316,7 +316,7 @@ tasks.withType(Javadoc) {
 		// fixtures, and we want to have mockito also available in
 		// test, so we use API here.
 		testFixturesApi "org.mockito:mockito-core:${mockitoVersion}"
-
```
Flowdalic, line 319: Unrelated whitespace change, please remove this from the PR (but feel encouraged to add it as new PR).

smack-core/src/testFixtures/java/org/jivesoftware/smack/test/util/WaitForPacketListener.java
```diff
@@ -43,7 +43,7 @@ public void waitAndReset() {
 
     public void waitUntilInvocationOrTimeout() {
         try {
-            boolean res = latch.await(300, TimeUnit.SECONDS);
+            boolean res = latch.await(10, TimeUnit.SECONDS);
```
Flowdalic, line 46: I don't see a reason why this should be lowered to 10 seconds. Especially since 10 seconds is easily in the area where heavily congested systems may delay a thread. I'd like to prefer the original value, we could, however talk about making this configurable. But then, as above, no as part of this PR.

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -1063,6 +1064,52 @@ public void invite(MessageBuilder messageBuilder, EntityBareJid user, String rea
...
+     *
+     * @param address the user to send the invitation to
+     * @param reason the purpose for the invitation
+     * @param password specifies a password needed for entry
+     * @param continue_bool specifies if the groupchat room continues a one-to-one chat having the designated thread
+     * @param thread the thread to continue
+     * @throws NotConnectedException if the XMPP connection is not connected.
+     * @throws InterruptedException if the calling thread was interrupted.
+     */
+    public void inviteDirectly(EntityBareJid address, String reason, String password, boolean continue_bool, String thread)
```
Flowdalic, line 1095: Smack uses camel case, as most of the Java ecosystem, not snake case. I also think we do not need to encode the type in the parameter name.
```suggestion
    public void inviteDirectly(EntityBareJid address, String reason, String password, boolean continue, String thread)
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
```diff
@@ -1063,6 +1064,52 @@ public void invite(MessageBuilder messageBuilder, EntityBareJid user, String rea
...
+        throws NotConnectedException, InterruptedException {
+        MessageBuilder messageBuilder = connection.getStanzaFactory().buildMessageStanza();
```
Flowdalic, line 1097: Use the builder pattern if it is available
```suggestion
        Message message = connection.getStanzaFactory()
            .buildMessageStanza()
            .to(address)
            .addExtension(invitation)
            .build();
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,33 @@ public void processStanza(Stanza packet) {
         };
         connection.addAsyncStanzaListener(invitationPacketListener, INVITATION_FILTER);
 
+        // Listens for all messages that include an XEP-0249 GroupChatInvitation extension and fire the invitation
+        // listeners
+        StanzaListener directInvitationPacketListener = new StanzaListener() {
+            @Override
+            public void processStanza(Stanza packet) {
```
Flowdalic, line 213: We try to phase out the term `packet` in Smack, as it is not XMPP-ish.
```suggestion
            public void processStanza(Stanza stanza) {
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -79,7 +83,67 @@ public class GroupChatInvitation implements ExtensionElement {
...
+     * @param roomAddress the address of the group chat room.
+     * @param reason the purpose for the invitation
+     * @param password specifies a password needed for entry
+     * @param continue_bool specifies if the groupchat room continues a one-to-one chat having the designated thread
+     * @param thread the thread to continue
+     */
+    public GroupChatInvitation(EntityBareJid roomAddress,
+                               String reason,
+                               String password,
+                               Boolean continue_bool,
```
Flowdalic, line 104: No reason to use a boxed type, also see above regarding snake case.
```suggestion
                               boolean continue,
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -69,6 +69,10 @@ public class GroupChatInvitation implements ExtensionElement {
     public static final QName QNAME = new QName(NAMESPACE, ELEMENT);
 
     private final EntityBareJid roomAddress;
+    private final String reason;
+    private final String password;
+    private final String thread;
+    private final Boolean continue_bool;
```
Flowdalic, line 75: No reason to use a boxed type, also see above regarding snake case.
```suggestion
    private final boolean continue;
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -79,7 +83,67 @@ public class GroupChatInvitation implements ExtensionElement {
...
+    public String getThread() {
+        return thread;
+    }
+
+    /**
+     * Returns whether the groupchat room continues a one-to-one chat.
+     *
+     * @return whether the groupchat room continues a one-to-one chat.
+     */
+    public Boolean getContinue() {
```
Flowdalic, line 145: I am not really happy about the method name, as it is not expressive. Also return type. Maybe
```suggestion
    public boolean continuesAsOneToOneChat() {
```

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/packet/GroupChatInvitation.java
```diff
@@ -107,6 +171,13 @@ public String getNamespace() {
     public XmlStringBuilder toXML(org.jivesoftware.smack.packet.XmlEnvironment enclosingNamespace) {
         XmlStringBuilder xml = new XmlStringBuilder(this);
         xml.attribute("jid", getRoomAddress());
+        xml.optAttribute("reason", getReason());
+        xml.optAttribute("password", getPassword());
+        xml.optAttribute("thread", getThread());
+
+        if (getContinue())
+            xml.optAttribute("continue", getContinue().toString());
```
Flowdalic, line 179: I think `XmlStringBuilder` provides `optBoolAttribute()`.

smack-extensions/src/test/java/org/jivesoftware/smackx/muc/MultiUserChatTest.java
```diff
@@ -0,0 +1,324 @@
...
+
+        if (groupInvitationListener.getError() != null) {
+            throw groupInvitationListener.getError();
+        }
+    }
+
+    @Test
+    public void shouldReceiveOfflineInvitationWithAllOptionalAttributes() throws XmppStringprepException, Throwable {
+        EntityBareJid roomJid = JidCreate.entityBareFrom("room@example.com");
+        String expected_reason = "reason";
```
Flowdalic, line 216: camel case, please

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChat.java
MF1-MS: `continue` is a Java keyword, so I'd rather not just call it that. Have amended to `continueAsOneToOne`

Flowdalic: +1, but maybe go all the way and use `continueAsOneToOneChat`.

MF1-MS: Yeh I'd gone back and done that on second thoughts 👍

MF1-MS:
> I am not sure if you re-invented `org.jivesoftware.util.Protocol` with `InvokeDirectlyResponder` in your tests (or similar already existing classes of Smack's test fixtures). Please check that you don't invent the wheel again, I wouldn't be surprised if Smack wouldn't already provide what you need for the tests. Unfortunately, I had no time to dig into this further.

I think these are suitably disparate. `org.jivesoftware.util.Protocol` (which seems to have an out of data docstring, but I found good usage examples in the Socks5 tests) seems focussed on enabling dialogues where Smack will be expecting responses to its requests, and verifying the requests Smack sent. The sent stanza we're verifying with `InvokeDirectlyResponder` is generated by Smack due to it conceptually receiving a message. 

`InvokeDirectlyResponder` simply builds off the `DummyConnection` test fixture, capturing an outgoing stanza and verifying it.

MF1-MS:
I can squash the commits and amend the message if you're happy, but leaving as is for now for a cleaner diff.

Are you happy with my other PRs too?:
https://github.com/igniterealtime/Smack/pull/537
https://github.com/igniterealtime/Smack/pull/538

smack-extensions/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatManager.java
```diff
@@ -199,6 +206,33 @@ public void processStanza(Stanza packet) {
...
+                final Message message = (Message) stanza;
+                GroupChatInvitation invite =
+                    stanza.getExtension(GroupChatInvitation.class);
+
+                // Fire event for invitation listeners
+                final MultiUserChat muc = getMultiUserChat(invite.getRoomAddress());
+                final XMPPConnection connection = connection();
+                final EntityJid from = message.getFrom().asEntityJidIfPossible();
+                if (from == null) {
+                    LOGGER.warning("Group Chat Invitation from non entity JID: '" + message.toXML() + "'");
```
Flowdalic, line 223: Do not spit out the whole XML of the causing stanza if it is not a debug message.
```suggestion
                    LOGGER.warning("Group Chat Invitation from non entity JID in '" + message + "'");
```

Flowdalic (changes requested):
LGTM, only one minor thing, then this is ready to go.

Please
1. squash the commits into one
2. Add an table row for xep249 to https://github.com/igniterealtime/Smack/blob/f6c85d9fb3a4d8baa96d48dc428b53974d9fdba5/smack-java8-full/src/main/java/org/jivesoftware/smackx/package-info.java#L348-L361
3. rebase on the current master
4.  mention "Fixes SMACK-932" in the commit message's body. Also

MF1-MS:
Should be good to merge now :)

Flowdalic:
The commit message's subject currently reads "Fixes SMACK-932.". That is not really helpful if I look at a commit log only consisting of the commit subject.

MF1-MS:
Ok, commit message re-arranged.

PR #537, opened 2022-08-15
Use XMPP connection as local socks5 address

MF1-MS:
The default local address is often just "the first address found in the list of addresses read from the OS" and this might mean an internal IP address that cannot reach external servers. So wherever possible use the same IP address being used to connect to the XMPP server because this local address has a better chance of being suitable.

This MR adds the above behaviour, and two UTs to test that we use the local XMPP connection IP when connected, and the previous behaviour when not.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
@@ -16,6 +16,7 @@
  */
 package org.jivesoftware.smack;
 
+import java.net.InetAddress;
 import java.util.concurrent.TimeUnit;
 
 import javax.xml.namespace.QName;
@@ -159,6 +160,14 @@ public interface XMPPConnection {
      * @return the full XMPP address of the user logged in.
      */
     EntityFullJid getUser();
+
+    /**
+     * Returns the local address currently in use for this connection, or <code>null</code> if
+     * this is invalid for the type of underlying connection.
+     *
+     * @return the local address currently in use for this connection
+     */
+    InetAddress getLocalAddress();
 
     /**
      * Returns the stream ID for this connection, which is the value set by the server
--- a/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
+++ b/smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
@@ -1938,4 +1938,20 @@ public class XMPPTCPConnection extends AbstractXMPPConnection {
         this.bundleAndDeferCallback = bundleAndDeferCallback;
     }
 
+
+    /**
+     * Returns the local address currently in use for this connection.
+     *
+     * @return the local address
+     */
+    @Override
+    public InetAddress getLocalAddress() {
+        final Socket socket = this.socket;
+        if (socket == null) return null;
+
+        InetAddress localAddress = socket.getLocalAddress();
+        if (localAddress.isAnyLocalAddress()) return null;
+
+        return localAddress;
+    }
 }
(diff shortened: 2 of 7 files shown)


smack-tcp/src/main/java/org/jivesoftware/smack/tcp/XMPPTCPConnection.java
```diff
@@ -1938,4 +1938,18 @@ public void setBundleandDeferCallback(BundleAndDeferCallback bundleAndDeferCallb
...
 
+
+    /**
+     * Returns the local address currently in use for this connection.
+     *
+     * @return the local address
+     */
+    @Override
+    public InetAddress getLocalAddress() {
+        InetAddress xmppAddress = (socket != null) ? socket.getLocalAddress() : null;
```
Flowdalic, line 1949: I think a cleaner and more readable way to write this method is

```
final Socket socket = this.socket
if (socket == null) return null;

InetAddress localAddress = socket.getLocalAddress()
if (localAddress.isAnyLocalAddress) return null;

return localAddress;
```

smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
```diff
@@ -141,6 +142,16 @@ public interface XMPPConnection {
      */
     EntityFullJid getUser();
 
+    /**
+     * Returns the local address currently in use for this connection, or <code>null</code> if
+     * this is invalid the type of underlying connection.
+     *
+     * @return the local address currently in use for this connection
+     */
+    default InetAddress getLocalAddress() {
```
Flowdalic, line 151: I tend to have this not come with a default implementation, as it makes it obvious to creators of XMPPConnection classes that there is some behavior that, even though not strictly required, they may want to implement.

Flowdalic:
> The default local address is often just "the first address found in the list of addresses read from the OS" and this might mean an internal IP address that cannot reach external servers. So wherever possible use the same IP address being used to connect to the XMPP server because this local address has a better chance of being suitable.
> 
> This MR adds the above behaviour, and two UTs to test that we use the local XMPP connection IP when connected, and the previous behaviour when not.

*That* is exactly what belongs in the commit message. Sadly the commit message right now does contain stuff that is just not true. ;)

smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
MF1-MS: OK, I see the benefit of that but the result of this is the default has been copied and pasted into `ModularXmppClientToServerConnection`, `XMPPBOSHConnection` and `DummyConnection`.

smack-extensions/src/main/java/org/jivesoftware/smackx/bytestreams/socks5/Socks5Proxy.java
```diff
@@ -158,8 +158,6 @@ protected Socks5Proxy(ServerSocket serverSocket) {
         startServerThread();
     }
 
-
```
Flowdalic, line 161: Please drop the whitespace only change.

Flowdalic (changes requested):
LGTM, only minor whitespace fix.

Please rebase on the current master and address the feedback. Thanks for your contribution and sorry that it took so long to review.

Flowdalic (changes requested):

smack-core/src/main/java/org/jivesoftware/smack/XMPPConnection.java
```diff
@@ -160,6 +161,14 @@ public interface XMPPConnection {
      */
     EntityFullJid getUser();
 
+    /**
+     * Returns the local address currently in use for this connection, or <code>null</code> if
+     * this is invalid the type of underlying connection.
```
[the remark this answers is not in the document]
MF1-MS: Thanks, fixed up.

PR #539, opened 2022-08-30
[muc] Fix removal of the MUC's main presence interceptor

tharkum:
On dinamically remove the last existed presence interceptor
we also should to remove the MUC's main presence interceptor from the connection.

Hot fix for SMACK-925.


Flowdalic:
Thanks, could you add a

Fixes: 60fee7b3181f ("[muc] Fix Presence interceptors")

git trailer to the commit message?

PR #542, opened 2023-01-19
LowLevelIntegrationTest#get*Connection() ignores descriptors

guusdk:
Low level Integration test, as implemented by `AbstractSmackLowLevelIntegrationTest` execute for each ConnectionDescriptor that is configured in `XmppConnectionManager`. However, their `getConnectedConnection()` and `getUnconnectedConnection()` methods will use the default connection descriptor. This can result in multiple executions of the same test, that all use the same connection descriptor.

This commit addresses the issue by allowing the to-be-used descriptor to be set prior to each test execution.


Flowdalic:
https://github.com/igniterealtime/Smack/pull/549 is my take on the problem.

Flowdalic:
Superseded by #549.

PR #546, opened 2023-01-30
Add null check for mucUser.

bgrozev:
XEP-0045 requires "unavailable" presence to contain extended presence, but Smack shouldn't throw an exception if it doesn't.

There's an explicit null check on line 260, so accessing the value without a check on 279 is inconsistent.


Flowdalic:
LGTM. This appears to be a good candidate for the 4.4 branch. If you think so too, then please consider changing the target branch of this PR to 4.4. Also friendly reminder regarding https://discourse.igniterealtime.org/t/thread-stuck-in-multiuserchat-enter-forever/92090/10?u=flow

bgrozev:
Rebased and changed target to the `4.4` branch.

PR #548, opened 2023-02-06
[core] short message if waitForClosingStreamTagFromServer() times out

brantsch:
Instead of a long message looking like an error in level info we just log a short and friendly message without a stack trace if waitForClosingStreamTagFromServer() experiences an entirely inconsequential timeout.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/AbstractXMPPConnection.java
@@ -1046,7 +1046,10 @@ public abstract class AbstractXMPPConnection implements XMPPConnection {
 
     protected final boolean waitForClosingStreamTagFromServer() {
         try {
-            waitForConditionOrThrowConnectionException(() -> closingStreamReceived, "closing stream tag from the server");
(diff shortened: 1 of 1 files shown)


Flowdalic:
So the motivation is to suppress the stacktrace for the `NoResponseException`? I know that some confuse log messages with a stacktrace with an error situation. But it is actually the log level that indicates the severity of the log message. And here it is just a 'info' level.  Furthermore, I prefer to provide the log subystem with as much contextual information as possible.

If a 'info' log-level stacktrace confuses your users, then you should configure your log system so that stacktraces are not print if the level is 'info'.

brantsch:
Thank you for taking the time to reply, even more so in a polite and helpful way.

> So the motivation is to suppress the stacktrace for the `NoResponseException`? I know that some confuse log messages with a stacktrace with an error situation. But it is actually the log level that indicates the severity of the log message. And here it is just a 'info' level. Furthermore, I prefer to provide the log subystem with as much contextual information as possible.
I agree. It is just slightly unusual to see stack traces in INFO.

> If a 'info' log-level stacktrace confuses your users, then you should configure your log system so that stacktraces are not print if the level is 'info'.
Thank you for this suggestion. It sounds like a good solution.

PR #555, opened 2023-02-26
Rename ELEMENT 'candidate-activated' to 'activated' per XEP-0260

cmeng-git:


Flowdalic:
Applied as ef0fc01505f8c6c1fd86201d8c48859a363d8605

PR #557, opened 2023-03-01
Add XEP-0070: Verifying HTTP Requests via XMPP support

cmeng-git:
Add XEP-0070: Verifying HTTP Requests via XMPP support


smack-core/src/main/java/org/jivesoftware/smack/packet/IQ.java
```diff
@@ -194,7 +194,6 @@ private void appendInnerXml(XmlStringBuilder xml) {
         if (type == Type.error) {
             // Add the error sub-packet, if there is one.
             appendErrorIfExists(xml);
-            return;
```
Flowdalic, line 197: Please see https://discourse.igniterealtime.org/t/smack-4-4-6-iq-error-stanza-generation-does-not-allow-adding-of-an-extension-element/92622/3?u=flow and adjust the code accordingly.

Flowdalic (changes requested):
I usually do not add new features to stable Smack branches, so please adjust the target branch to `master`.

smack-core/src/main/java/org/jivesoftware/smack/packet/IQ.java
Flowdalic: Strike that, I was being stupid.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpauthorizationrequest/HTTPAuthorizationRequestManager.java
```diff
@@ -0,0 +1,217 @@
...
+            FromTypeFilter.DOMAIN_BARE_JID
+    );
+
+    private static final Logger LOGGER = Logger.getLogger(HTTPAuthorizationRequestManager.class.getName());
+
+    private static final Map<XMPPConnection, HTTPAuthorizationRequestManager> INSTANCES = new WeakHashMap<>();
+
+    private final Set<HTTPAuthorizationRequestListener> incomingListeners = new CopyOnWriteArraySet<>();
+
+    private ConfirmExtension mConfirmExtension;
```
Flowdalic, line 74: The current design of the Manager is far from ideal, it is racy, allows at most one outstanding verification, and you never know what you accept or reject.

If I understand the protocol correctly, then the manager could simply provide a method
```
void acceptId(URI url, String method, String id)
```
while keeping a Set of all outstanding IDs that it will accept. That way, the listeners simply look if the incoming ID is in the set, and, if so, remove the ID from the set and accept, otherwise reject.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpauthorizationrequest/HTTPAuthorizationRequestManager.java
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * The HTTP Request manager for HTTP Request. The manager handles both the HTTP Request via:
+ * IQ : when the request is via EntityFullJid
+ * Message: When request is via EntityBareJid
+ *
+ * XEP-0070: Verifying HTTP Requests via XMPP (1.0.1 (2016-12-09))
+ *
+ * @see HTTPAuthorizationRequestListener on callback
+ */
+public final class HTTPAuthorizationRequestManager extends Manager {
```
Flowdalic, line 57: Smack more or less follows Google Java coding guidelines, so this should be
```suggestion
public final class HttpAuthorizationRequestManager extends Manager {
```

Flowdalic (changes requested):
First round of review. I think especially the manager design needs to be changed.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpauthorizationrequest/element/ConfirmExtension.java
```diff
@@ -0,0 +1,119 @@
...
+
+    public static final QName QNAME = new QName(NAMESPACE, ELEMENT);
+
+    public static final String ATTR_ID = "id";
+    public static final String ATTR_METHOD = "method";
+    public static final String ATTR_URL = "url";
+
+    private final String id;
+    private final String method;
+    private final String url;
```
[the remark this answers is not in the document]
cmeng-git: Can this variable be handled as a string all the ways. Otherwise need to do extract steps during xml phrase() and in toXML() method.
In aTalk app implementation, the value is used as string in prompting user to accept or reject the auth request.

smack-experimental/src/main/java/org/jivesoftware/smackx/httpauthorizationrequest/HTTPAuthorizationRequestManager.java
cmeng-git: Add the following to support multiple instances of auth requests:
a. private static final Map<String, Object> mAuthRequests = new HashMap<>()
b. public void acceptId(String id)
c. public void rejectId(String id)

stokito (approved):

smack-experimental/src/main/java/org/jivesoftware/smackx/httpauthorizationrequest/element/ConfirmExtension.java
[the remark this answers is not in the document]
Flowdalic: > The URI class is not intended to be in fields.

Care to elaborate on that?

cmeng-git:
Try to add test script to the package but having problem i.e.
> No tests found for given includes: [org.jivesoftware.smackx.httpauthorizationrequest.ConfirmElementTest.confirmElementSerializationTest](--tests filter)

Any advice how to resolve this problem.

PR #565, opened 2023-06-24
Migrate from libsignal-protocol-java to libsignal

vanitasvitae:
This PR moves away from `libsignal-protocol-java` which was archived in February of 2022 and instead makes use of `libsignal` which provides java bindings.

The migration would require a minimum Android API level bump, since the `libsignal` code now uses `java.util.Optional`, which is Java 8 (and some Android API level higher than the current min).

I have not done any interoperability- or integration testing yet.
It is also not yet clear, if the ecosystem (OMEMO next) will fork libsignal, or maintain a fork of libsignal-protocol-java instead, so consider this more a feasibility analysis for now :)

### DIFF
--- a/smack-omemo-signal/build.gradle
+++ b/smack-omemo-signal/build.gradle
@@ -6,8 +6,7 @@ dependencies {
 	api project(":smack-im")
 	api project(":smack-extensions")
 	api project(":smack-omemo")
-	implementation 'org.whispersystems:signal-protocol-java:2.6.2'
-
+	implementation 'org.signal:libsignal-client:0.26.0'
 	testFixturesApi(testFixtures(project(":smack-core")))
 	testImplementation project(path: ":smack-omemo", configuration: "testRuntime")
 }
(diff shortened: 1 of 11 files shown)


Flowdalic:
@vanitasvitae we are in the process of release the next version of Smack and it seems like this PR would be a good candidate for the next version. Could I bribe you into rebasing this PR onto the latest master while addressing the sensible suggestions @stokito gave? Thanks in advance :)

Flowdalic:
> I'll remove them to not spent a time even on reading them

I am sorry, but this is counter productive. The points you raised are valid and should be addressed before this is merged. For example, the new error prone version will flag the usages of `LinkedList` (rightly so, in most cases, if I may say).

Please restore your comments.

Flowdalic:
> [Undefined reference] org.jivesoftware.smackx.omemo.signal.(SignalOmemoRatchet.java:87)
>  >> java.util.Optional

```java
            if (!preKeyMessage.getPreKeyId().isPresent()) {
```

Android added `j.u.Optional` with API level 24. Smack is currently 23 and was, until recently 19 (IIRC). Looking at the Android API distribution charts, I think we can raise the API level once more to 24. Especially since I consider the availability of `Optional` desirable.

However, I am not yet sure why this check starts to fail now, in code that was not changed in this PR. Needs more investigation, but first lunch… ;)

vanitasvitae:
One further note: The PR still uses `libsignal-client:0:26.0`, which is over a year old as of now.
I don't have the time to check newer releases for breaking changes (e.g. pqc stuff) or important security updates right now, but it may be worthwhile for you to check, if Smack may benefit from a newer release of the library.

Flowdalic:
Please rebase, Smack's minimum Android SDK level is now 26. That should fix the compilation issue.

PR #566, opened 2023-09-29
[sinttest] Fix IBR usage (and fix typos)

guusdk:
Prior to this change, the Smack Integration Test are configured to use In-Band Registration when no admin credentials are provided. However, when trying to apply In-Band Registration, the (missing) admin credentials are used, leading to an error.

In this PR, usage of the admin credentials is prevented when IBR is in use.

Additionally, some typos in a log message and two exception messages are fixed in a separate commit.

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/XmppConnectionManager.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/XmppConnectionManager.java
@@ -285,7 +285,7 @@ public class XmppConnectionManager {
             if (unsuccessfullyDeletedAccountsCount == 0) {
                 LOGGER.info("Successfully deleted all created accounts ✔");
             } else {
-                LOGGER.warning("Could not delete all created accounts, " + unsuccessfullyDeletedAccountsCount + " remainaing");
+                LOGGER.warning("Could not delete all created accounts, " + unsuccessfullyDeletedAccountsCount + " remaining");
             }
         }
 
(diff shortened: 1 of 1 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/XmppConnectionManager.java
```diff
@@ -184,20 +184,18 @@ public static boolean removeConnectionDescriptor(String nickname) {
 
         switch (sinttestConfiguration.accountRegistration) {
         case serviceAdministration:
-        case inBandRegistration:
             accountRegistrationConnection = defaultConnectionDescriptor.construct(sinttestConfiguration);
             accountRegistrationConnection.connect();
             accountRegistrationConnection.login(sinttestConfiguration.adminAccountUsername,
```
Flowdalic, line 189: Couldn't we simply move this line into the `else` block below?

guusdk: Maybe? Having an if/else block in a switch/case that uses the same arguments smells. The fact that I have to think if it would be a suitable alternative makes me prefer my original solution: The code is a lot easier to understand, and isn't even any longer. That goes a long way in preventing the kind of bugs that this PR is fixing.

Flowdalic: This approach causes 8 insertions and 9 deletions (if you don't count the deletion of the empty line 194). The reduction in deletions comes due the removal of the if/else block, which accounts for 3 deleted lines.

However, it introduces some duplicate code. And the change itself is much more verbose as compared to https://github.com/igniterealtime/Smack/pull/573.

I prefer to avoid duplicate code and hence tend to to with https://github.com/igniterealtime/Smack/pull/573 over this.

But in any case, thanks for the bug report!

Flowdalic:
Closing in favor of https://github.com/igniterealtime/Smack/pull/573

PR #571, opened 2023-11-13
[websocket] Reduce fragility

guusdk:
Replace string-based comparison with a XML parsing when checking for 'open' and 'close' elements.


Flowdalic:
Thanks, see https://github.com/igniterealtime/Smack/commit/097ab20485c3644d9195581a931915696f96ab81 for the follow up commit.

PR #579, opened 2024-03-14
Sint tagging

guusdk:
A new annotation is introduced (`SpecificationReference`) that can be used to annotate a SINT test.

These properties are available in the annotation:
- `document`: Iidentifier for a specification document, such as 'RFC 6120' or 'XEP-0485'
- `section`: Identifier for a section (or paragraph), such as '6.2.1'
- `quote`: A quotation of relevant text from the section

The SINT execution framework is modified so that two new configuration options are available:
- `enabledSpecifications`
- `disabledSpecifications`

These operate on the value of the `document` property of the annotation. Their usage is comparable
to that of the pre-existing `enabledTests` and `disabledTest` configuration options.

Execution output now includes the document, section and quote that's on the annotated test, when
the test fails. This allows an end-user to easily correspond a test failure with a particular
specification.

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
@@ -44,6 +44,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -71,6 +73,7 @@ import org.igniterealtime.smack.inttest.Configuration.AccountRegistration;
 import org.igniterealtime.smack.inttest.annotations.AfterClass;
 import org.igniterealtime.smack.inttest.annotations.BeforeClass;
 import org.igniterealtime.smack.inttest.annotations.SmackIntegrationTest;
+import org.igniterealtime.smack.inttest.annotations.SpecificationReference;
 import org.reflections.Reflections;
 import org.reflections.scanners.MethodAnnotationsScanner;
 import org.reflections.scanners.MethodParameterScanner;
@@ -128,10 +131,21 @@ public class SmackIntegrationTestFramework {
         final int exitStatus;
         if (failedTests > 0) {
             LOGGER.warning("💀 The following " + failedTests + " tests failed! 💀");
+            final SortedSet<String> bySpecification = new TreeSet<>();
             for (FailedTest failedTest : testRunResult.failedIntegrationTests) {
                 final Throwable cause = failedTest.failureReason;
                 LOGGER.log(Level.SEVERE, failedTest.concreteTest + " failed: " + cause, cause);
+                if (failedTest.concreteTest.method.isAnnotationPresent(SpecificationReference.class)) {
+                    final String specificationReference = getSpecificationReference(failedTest.concreteTest.method);
+                    if (specificationReference != null) {
+                        bySpecification.add("- " + specificationReference + " (as tested by '" + failedTest.concreteTest + "')");
+                    }
+                }
             }
+            if (!bySpecification.isEmpty()) {
+                LOGGER.log(Level.SEVERE, "The failed tests correspond to the following specifications:" + System.lineSeparator() + String.join(System.lineSeparator(), bySpecification));
+            }
+
             exitStatus = 2;
         } else {
             LOGGER.info("All possible Smack Integration Tests completed successfully. \\o/");
@@ -147,6 +161,24 @@ public class SmackIntegrationTestFramework {
         }
 
         System.exit(exitStatus);
+    }
+
+    private static String getSpecificationReference(Method method) {
+        final SpecificationReference spec = method.getDeclaringClass().getAnnotation(SpecificationReference.class);
+        if (spec == null || spec.document().isBlank()) {
+            return null;
+        }
+        String line = spec.document().trim();
+
+        final SmackIntegrationTest test = method.getAnnotation(SmackIntegrationTest.class);
+        if (!test.section().isBlank()) {
+            line += " section " + test.section().trim();
+        }
+        if (!test.quote().isBlank()) {
+            line += ":\t\"" + test.quote().trim() + "\"";
+        }
+        assert !line.isBlank();
+        return line;
     }
 
     public SmackIntegrationTestFramework(Configuration configuration) {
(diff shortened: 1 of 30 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/annotations/SpecificationReference.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return a document identifier
+     */
+    String document();
+
+    /**
+     * Unique identifier for a section (or paragraph) of the document referenced by {@link #document()}, such as '6.2.1'.
+     *
+     * @return a document section identifier
+     */
+    String section() default "";
```
Flowdalic, line 47: Is it possible to make the default `null` instead of the empty string (here and for 'quote')? That would allow us to reject annotations which contain the empty string as value.

smack-integration-test/src/main/java/org/jivesoftware/smackx/caps/EntityCapsTest.java
```diff
@@ -91,6 +92,7 @@ public void tearDown() throws NotConnectedException, InterruptedException {
     }
 
     @SmackIntegrationTest
+    @SpecificationReference(document = "XEP-0115")
```
Flowdalic, line 95: I am using this code line as example.

I would split this annotation into two: `SpecificationDocumentReference` and `SpecificationReference`. The former `SpecificationDocumentReference` should be used on the class level, while the `SpecificationReference` should be used on the method level. `SpecificationDocumentReference` only takes a required document attribute.

The idea is that you only need to add `SpecificationDocumentReference` to sinttest classes. This has the advantage of reducing boilerplate *and* being less prone to forgetting correctly tagging tests newly added to a class.

I first though that we could remove the `document` property from `SpecificationReference` then. Not sure if there are reasonable cases where you have a method belonging to a sinttest class of XEP-A which is actually testing XEP-B. However, I could image that this case will come up, especially if you have cross cutting concerns between XEPs. We could cross that bridge when we come to it.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/annotations/SpecificationReference.java
guusdk: I don't think that's possible in Java.

smack-integration-test/src/main/java/org/jivesoftware/smackx/caps/EntityCapsTest.java
guusdk: I prefer to keep the reference to the document and the reference to the section within the document in one annotation. That makes the annotation self-sufficient: that's easier to parse, for both humans as well as in code.

It is also _less_ error prone: when keeping document & section together, any error most likely introduces a _missing_ reference, but when splitting them up, it is a lot easier to create an _incorrect_ reference (by mis-matching document and section). That's generally a worse.

Flowdalic: > I prefer to keep the reference to the document and the reference to the section within the document in one annotation.

I prefer splitting this to avoid redundant information.

I also noticed that we could probably, and should if possible, re-use the `@SmackIntegrationTest` annotation. So `@SpecificationReference` could become an annotation only containing `document` which must only be used on classes, not methods.

guusdk: Both changes have now been applied. I want my dislike for this approach on the record though :)

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -128,10 +131,21 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
         final int exitStatus;
         if (failedTests > 0) {
             LOGGER.warning("💀 The following " + failedTests + " tests failed! 💀");
+            final SortedSet<String> bySpecification = new TreeSet<>();
             for (FailedTest failedTest : testRunResult.failedIntegrationTests) {
                 final Throwable cause = failedTest.failureReason;
                 LOGGER.log(Level.SEVERE, failedTest.concreteTest + " failed: " + cause, cause);
+                if (failedTest.concreteTest.method.isAnnotationPresent(SpecificationReference.class)) {
+                    final String specificationReference = getSpecificationReference(failedTest);
```
Flowdalic, line 139: I'd simply set `getSpecificationReference()` return `null` if no specification references where found in the annotation(s). This saves us one level of indentation.

Also, we should assume that this method never returns the empty string. Either `null` or the non-empty String.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -149,6 +163,22 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
         System.exit(exitStatus);
     }
 
+    private static String getSpecificationReference(FailedTest failedTest) {
+        final SmackIntegrationTest test = failedTest.concreteTest.method.getAnnotation(SmackIntegrationTest.class);
+        final SpecificationReference spec = failedTest.concreteTest.method.getAnnotation(SpecificationReference.class);
```
Flowdalic, line 168: Mustn't this be retrieved from the class declaring the method?

Also, if not existent, then early exit returning `null`.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -149,6 +163,22 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
...
+        String line = "";
+        if (spec.document() != null && !spec.document().isBlank()) {
```
Flowdalic, line 170: We could skip the `null` check here and below, couldn't we?

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -149,6 +163,22 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
         System.exit(exitStatus);
     }
 
+    private static String getSpecificationReference(FailedTest failedTest) {
```
Flowdalic, line 166: This should not take `FailedTest` as argument, but `Method`.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -392,6 +422,28 @@ private void runTests(Set<Class<? extends AbstractSmackIntTest>> classes)
                     it.remove();
                     continue;
                 }
+
+                final String specification;
+                if (method.isAnnotationPresent(SpecificationReference.class)) {
```
Flowdalic, line 427: Simply re-use `getSpecificationReference()` here

Flowdalic (changes requested):

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
guusdk: This only gets the specification's document, not the entire thing. This entire block, however, should be moved up and operate on a class level, instead of a method level.

guusdk:
I believe I've processed all review comments.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -149,6 +163,22 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
...
     }
 
+    private static String getSpecificationReference(Method method) {
+        final SmackIntegrationTest test = method.getAnnotation(SmackIntegrationTest.class);
+        final SpecificationReference spec = method.getDeclaringClass().getAnnotation(SpecificationReference.class);
+        if (spec == null || spec.document().isBlank()) {
+            return null;
+        }
+        String line = spec.document().trim();
+        if (test.section() != null && !test.section().isBlank()) {
```
Flowdalic, line 173: As far as I can tell, this `null`-check and the one below are not necessary because the annotation type element values can not be `null`. Personally, I would drop those checks, but I leave that up to you.

Flowdalic (changes requested):
Looking good, only two minors things.

PR #580, opened 2024-03-21
Debugger context

guusdk:
This PR contains two related commits:
1. Introduction of a static 'context' in `SmackDebugger`, which is a simple static map.
2. Modification of the SINT framework to store the test that's currently being executed in this new context.

This mechanism allows SmackDebuggers to act in a way specific to the test that is being executed. An immediate use for this is having a log-generating debugger that writes XMPP traffic to a different file for each test.

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
@@ -87,6 +87,8 @@ public class SmackIntegrationTestFramework {
 
     public static boolean SINTTEST_UNIT_TEST = false;
 
+    private static ConcreteTest TEST_UNDER_EXECUTION;
+
     protected final Configuration config;
 
     protected TestRunResult testRunResult;
@@ -242,6 +244,10 @@ public class SmackIntegrationTestFramework {
         }
 
         return testRunResult;
+    }
+
+    public static ConcreteTest getTestUnderExecution() {
+        return TEST_UNDER_EXECUTION;
     }
 
     @SuppressWarnings({"Finally"})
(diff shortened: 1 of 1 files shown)


smack-core/src/main/java/org/jivesoftware/smack/debugger/SmackDebugger.java
```diff
@@ -41,6 +43,8 @@
  */
 public abstract class SmackDebugger {
 
+    private static final Map<String, String> CONTEXT = new HashMap<>();
```
Flowdalic, line 46: I'd rather use a list that is used like a stack.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -676,15 +677,19 @@ private PreparedTest(AbstractSmackIntTest test, List<ConcreteTest> concreteTests
...
         public void run() throws InterruptedException, XMPPException, IOException, SmackException {
             try {
+                SmackDebugger.putInContext("sint.test", test.getClass().getSimpleName());
+
                 // Run the @BeforeClass methods (if any)
                 executeSinttestSpecialMethod(beforeClassMethod);
 
                 for (ConcreteTest concreteTest : concreteTests) {
+                    SmackDebugger.putInContext("sint.concreteTest", concreteTest.toString());
                     runConcreteTest(concreteTest);
```
Flowdalic, line 687:
```suggestion
SinttestDebugger.pushContext(concreteTest.toString());
try {
                    runConcreteTest(concreteTest);
} finally {
    SinttestDebugger.popContext();
}
```

Flowdalic (changes requested):
I do think there is no need to modify anything outside of smack-integration-test for this approach, therefore we should not touch code, for example, in smack-core.

PR #582, opened 2024-04-04
sint: add assertion implementation to SimpleResultSyncPoint

guusdk:
By using an assertion, a test that uses a SimpleResultSyncPoint can now fail in a way equal to other test failures. This allows for failure messages to be defined.

Example usage:
```java
SimpleResultSyncPoint.assertSuccess(activeSyncPoint, timeout,
    "Expected " + conTwo.getUser() + " to receive an 'active' chat state from " +
     conOne + " (but they did not).");
```

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
@@ -16,8 +16,6 @@
  */
 package org.igniterealtime.smack.inttest.util;
 
-import org.junit.jupiter.api.Assertions;
-
 public class SimpleResultSyncPoint extends ResultSyncPoint<Void, Exception> {
 
     public void signal() {
@@ -30,15 +28,5 @@ public class SimpleResultSyncPoint extends ResultSyncPoint<Void, Exception> {
 
     public void signalFailure(String failureMessage) {
         signal(new Exception(failureMessage));
-    }
-
-    public static void assertSuccess(SimpleResultSyncPoint resultSyncPoint, long timeout, String message) throws InterruptedException {
-        try {
-            resultSyncPoint.waitForResult(timeout);
-        } catch (InterruptedException e) {
-            throw e;
-        } catch (Exception e) {
-            Assertions.fail(message, e);
-        }
     }
 }
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
@@ -18,10 +18,10 @@ package org.igniterealtime.smack.inttest.util;
 
 import org.junit.jupiter.api.Assertions;
 
-public class SimpleResultSyncPoint extends ResultSyncPoint<Boolean, Exception> {
+public class SimpleResultSyncPoint extends ResultSyncPoint<Void, Exception> {
 
     public void signal() {
-        signal(Boolean.TRUE);
+        signal((Void) null);
     }
 
     public void signalFailure() {
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
@@ -16,6 +16,8 @@
  */
 package org.igniterealtime.smack.inttest.util;
 
+import org.junit.jupiter.api.Assertions;
+
 public class SimpleResultSyncPoint extends ResultSyncPoint<Boolean, Exception> {
 
     public void signal() {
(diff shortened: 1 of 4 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
```diff
@@ -29,4 +31,14 @@ public void signalFailure() {
     public void signalFailure(String failureMessage) {
         signal(new Exception(failureMessage));
     }
+
+    public static void assertSuccess(SimpleResultSyncPoint ResultSyncPoint, long timeout, String message) throws InterruptedException {
+        try {
+            ResultSyncPoint.waitForResult(timeout);
```
Flowdalic, line 37: Drop the unnecessary `ResultSyncPoint` qualification. Furthermore, return the result of  `waitForResult()` in the method. There is no need to swallow it.

```suggestion
            return waitForResult(timeout);
```

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
```diff
@@ -29,4 +31,14 @@ public void signalFailure() {
     public void signalFailure(String failureMessage) {
         signal(new Exception(failureMessage));
     }
+
+    public static void assertSuccess(SimpleResultSyncPoint ResultSyncPoint, long timeout, String message) throws InterruptedException {
```
Flowdalic, line 35: I think it would probably be more consistent to call this method `assertResult()` or `assertHasResult()`. The result must not always indicate success.

Flowdalic (changes requested):
Is there any reason we want this method in `SimpleResultSyncPoint` and not directly in `ResultSyncPoint`?

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
Flowdalic: Thinking of it, this is basically, and therefore should become, `ResultSyncPoint.waitForResult()` with a second parameter `String message`.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
guusdk: `waitForResult()` is not a static method. It's invoked on the argument. The argument name should not start with a capital letter, which I've now fixed.

Convention for assertions implementations is that they either throw an `AssertionException` in case of failure, or do nothing (see `org.junit.jupiter.api.Assertions` for example). I've never seen an assertion that returns a value. That would allow for funky usage patterns (and possibly, confusing ones). Let's not break convention.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
guusdk: I think that's the difference between `ResultSyncPoint` and `SimpleResultSyncPoint`. In the 'simple' version, any result _does_ indicate success. Failures are represented by an Exception being thrown.

The `assertResult()` (or possibly `assertResultNotNull()`, `assertResultNull()` and `assertResultEquals()`) would make for nice candidates in the 'non-simple' version of the sync point implementation. Pragmatically, I've hardly seen these be used, so I was trying to prevent adding unneeded code in this PR.

guusdk:
> Is there any reason we want this method in `SimpleResultSyncPoint` and not directly in `ResultSyncPoint`?

As I hinted at in the inline feedback, the `SimpleResultSyncPoint`-based assertion is a lot simpler than the other one: it only knows of a success and failure outcome.

`ResultSyncPoint` on the other hand deals with return values. Appropriate assertions for those would include null and equals checks. As they're rarely used, they're of lower priority to me. I do not want to create them if I can help it, or do that as a separate PR if we must have them.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/util/SimpleResultSyncPoint.java
Flowdalic: > In the 'simple' version, any result _does_ indicate success.

No, `SimpleResultSyncPoint` returns a Boolean result, which very well could indicate an expected failure result.

If you really want to go with an junit-assert-like API, then have a look at https://github.com/igniterealtime/Smack/pull/583

guusdk: Thought: what if we'd change `SimpleResultSyncPoint` to extend from `ResultSyncPoint<Void, Exception>` instead of `ResultSyncPoint<Boolean, Exception>`?

guusdk: I've just pushed a commit that does exactly that. It was deceptively easy, requiring only code changes in places that are, as far as I can tell, unreachable code bits anyway (which in itself is smelly).

guusdk: I've now pulled in the API changes from #583 

Pondering if it'd be good to rename the method to `assertHasResult` or `assertNotNullResult` to allow for later additions like `assertResultEquals` - far from a hill for me to die on though.

Flowdalic:
Closing in favor of #583.

PR #586, opened 2024-04-10
[sinttest] Refactoring XEP-0045 test for default roles

guusdk:
Order of events in the test is made more predictable by explicitly waiting for all users to have joined the room, prior to checking their role.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
@@ -25,7 +25,9 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import java.util.Set;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
 
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.XMPPException;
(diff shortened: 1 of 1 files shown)


smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatRolesAffiliationsPrivilegesIntegrationTest.java
```diff
@@ -794,23 +811,26 @@ public void mucTestDefaultRoleForAffiliationInMembersOnlyRoom() throws Exception
...
             mucAsSeenByOne.grantMembership(conThree.getUser().asBareJid());
+            mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
 
             mucAsSeenByTwo.join(nicknameTwo);
             mucAsSeenByThree.join(nicknameThree);
-            mucAsSeenByOne.grantAdmin(conTwo.getUser().asBareJid());
-            adminResultSyncPoint.waitForResult(timeout);
-            assertEquals(mucAsSeenByOne.getOccupantsCount(), 3, "Unexpected occupant count in room " + mucAddress);
+
+            assertResult(allOccupantsDetectedSyncPoint, "Expected " + conTwo.getUser() + " to observe all of these occupants in room " + mucAddress + ", but not all of them appear to be in: " + expectedOccupants.stream().map(Object::toString).collect(Collectors.joining(", ")));
```
Flowdalic, line 833: It's conOne that does the observing, no?

```suggestion
            assertResult(allOccupantsDetectedSyncPoint, "Expected " + conOne.getUser() + " to observe all of these occupants in room " + mucAddress + ", but not all of them appear to be in: " + expectedOccupants.stream().map(Object::toString).collect(Collectors.joining(", ")));
```

guusdk: Absolutely. I actually intended that, but `con<tab>` apparently added the wrong reference, without me noticing.

PR #588, opened 2024-04-11
[sinttest] Allow custom processing of test run result

guusdk:
This adds a new configuration option, `testRunResultProcessors`, that allows a user to customize the way the results of a test run is processed.

By default, the pre-exising printing-to-stderr is used.

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.lang.reflect.Method;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -31,6 +32,7 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import javax.net.ssl.SSLContext;
 
@@ -131,6 +133,8 @@ public final class Configuration {
 
     public final CompatibilityMode compatibilityMode;
 
+    public final List<? extends SmackIntegrationTestFramework.TestRunResultProcessor> testRunResultProcessors;
+
     private Configuration(Configuration.Builder builder) throws KeyManagementException, NoSuchAlgorithmException {
         service = Objects.requireNonNull(builder.service,
                         "'service' must be set. Either via 'properties' files or via system property 'sinttest.service'.");
@@ -203,6 +207,7 @@ public final class Configuration {
 
         this.dnsResolver = builder.dnsResolver;
         this.compatibilityMode = builder.compatibilityMode;
+        this.testRunResultProcessors = builder.testRunResultProcessors;
     }
 
     public boolean isAccountRegistrationPossible() {
@@ -262,6 +267,8 @@ public final class Configuration {
         private DnsResolver dnsResolver = DnsResolver.minidns;
 
         private CompatibilityMode compatibilityMode = CompatibilityMode.standardsCompliant;
+
+        private List<? extends SmackIntegrationTestFramework.TestRunResultProcessor> testRunResultProcessors;
 
         private Builder() {
         }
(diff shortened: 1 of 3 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
```diff
@@ -473,6 +480,16 @@ public Builder setCompatibilityMode(String compatibilityModeString) {
             return setCompatibilityMode(compatibilityMode);
         }
 
+        public Builder setTestRunResultProcessors(String testRunResultProcessorsString) {
+            if (testRunResultProcessorsString == null) {
```
Flowdalic, line 484: The default-value logic shouldn't be hidden in a method called `setTestRunResultProcessors`, see also comment below.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
```diff
@@ -550,6 +567,7 @@ public static Configuration newConfiguration(String[] testPackages)
 
         builder.setCompatibilityMode(properties.getProperty("compatibilityMode"));
 
+        builder.setTestRunResultProcessors(properties.getProperty("testRunResultProcessors"));
```
Flowdalic, line 570: Add the default-value logic here. That is, if getProperty() does not return something, set the default value.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -114,54 +114,65 @@ public static void main(String[] args) throws IOException, KeyManagementExceptio
...
 
         if (config.debuggerFactory instanceof EnhancedDebugger) {
             EnhancedDebuggerWindow.getInstance().waitUntilClosed();
         }
 
+        final int exitStatus = testRunResult.failedIntegrationTests.isEmpty() ? 0 : 2;
         System.exit(exitStatus);
     }
 
+    public static class ConsoleTestRunResultProcessor implements TestRunResultProcessor {
```
Flowdalic, line 133: It's just co-incidental that the log messages appear on the console. But all this processor does is to feed the JUL API the test run result. Lets have the name reflect hat.

```suggestion
    public static class JulTestRunResultProcessor implements TestRunResultProcessor {
```

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/SmackIntegrationTestFramework.java
```diff
@@ -665,6 +676,10 @@ private static Exception throwFatalException(Throwable e) throws Error, NoRespon
         return (Exception) e;
     }
 
+    public interface TestRunResultProcessor {
```
Flowdalic, line 679: We probably could (and should) annotate this with `@FunctionalInterface`. Not sure if we already do this in Smack, but it is certainly not wrong doing so.

Flowdalic (changes requested):
Thanks for your contribution. Had time to give it some thought. Only minor things to fix.

guusdk:
Applied all review feedback.

Flowdalic:
Squash please.

guusdk:
Squashed.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
```diff
@@ -473,6 +480,15 @@ public Builder setCompatibilityMode(String compatibilityModeString) {
             return setCompatibilityMode(compatibilityMode);
         }
 
+        public Builder setTestRunResultProcessors(String testRunResultProcessorsString) {
+            if (testRunResultProcessorsString == null) {
```
Flowdalic, line 484: Calling a public API like this with `null` should usually throw an exception. It is not defined that calling this method with `null` actually does and therefore it is most likely a sign of a programming error, e.g., the user wanting to set a result processor for somehow it ended up being `null`.

The same argument can be made for the empty string.

Therefore,  use `StringUtils.requireNonNullNorEmpty(testRunResultProcessorString)` here.

Flowdalic (changes requested):
One minor thing, then its good to go.

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/Configuration.java
guusdk: I'm happy to change this, but it's a pattern that's used throughout this implementation. Do you want me to change it everywhere?

Flowdalic: Uh, fair point. Thanks for pointing this out. There is probably more to it that I don't remember.

In any case, if this is going to get changed, it will be done outside of this PR for all methods.

PR #590, opened 2024-04-26
[sinttest] Expand list of default test packages

guusdk:
The default set of packages that is scanned for integration tests are referencing jivesoftware. This commit adds igniterealtime-oriented package names.


guusdk:
I wonder if the alternative of scanning _all_ of the classpath by default, optionally limited by a provided config value, would be a bad thing.

Flowdalic:
> I wonder if the alternative of scanning _all_ of the classpath by default, optionally limited by a provided config value, would be a bad thing.

Classpaths can become very large, I'd rather limit it.

PR #591, opened 2024-04-30
[sinttest] Add optional version to specification reference

guusdk:
Some specifications are versioned. XEPs, for example, typicaly are. It is useful to annotate an implementation with the specific version of the specification that is being tested.


Fishbowler:
I love this idea! I hadn't considered that if someone inserts a new section between s2 and s3 that it'd throw off everything.

PR #592, opened 2024-05-02
[sinttest] Add versions for tests

guusdk:
Adds a version attribute to the annotations that point at the specification that's used in an integration test.


guusdk:
I've added a bit of a rationale in the commit message for each version number that I choose (which _mostly_ are the most recent versions of the XEP). Feel free to squash-merge this if you don't think this information is valuable to retain.

PR #595, opened 2024-05-10
[sinttest] prevent NullPointerException in RosterIntegrationTest

guusdk:
When listening for a presence stanza, assuming that every stanza will have a 'status' can cause NullPointerExceptions to be logged. This commit prevents that.


Flowdalic:
Thanks.

This was fixed with 51cd3f9b85491763e96b1b5d07a76c69541fb2b8.

PR #596, opened 2024-05-23
[muc] changeSubject errors should be propagated

guusdk:
When Smack requests a subject change of a MUC, an error returned by the server (eg: 'forbidden') should be propagated (as suggested by the pre-existing javadoc).


guusdk:
This fixes the problem reported in https://discourse.igniterealtime.org/t/multiuserchat-changesubject-times-out-without-auth/

Flowdalic:
LGTM

But could you change the type of  the argument of the `StanzaIdFilter` constructor from `Stanza`, to `StanzaView`? This would avoid introducing an new local variable `stanza` in the method and is something that we should do in any case (the stanza id is immutable, even in stanza builders). Bonus points if this is done in an individual commit.

guusdk:
I tried, but apparently the ID is not set yet without building the stanza.

Flowdalic:
Fixed with https://github.com/igniterealtime/Smack/commit/147071ff643971b4d873ddbdb16e347e4e353652

PR #597, opened 2024-06-07
[sinttest] Add new config option: 'host'

guusdk:
An optional configuration option for the Smack Integration Test framework has been added that allows one to bypass DNS when resolving a host for the XMPP domain that is the subject of the test.

The `host` option can be used with IP addresses (eg: `-Dsinttest.host=127.0.0.1`) and DNS names (eg: `-Dsinttest.host=example.org`).

The new option was tested with:
- not being defined (making sure it's optional)
- a reachable IP address
- an unreachable IP address
- a reachable DNS name
- an non-resolvable DNS name

### DIFF
--- a/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/package-info.java
+++ b/smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/package-info.java
@@ -80,6 +80,10 @@
  * <td>XMPP service to run the tests on</td>
  * </tr>
  * <tr>
+ * <td>host</td>
+ * <td>IP address or DNS name of the XMPP service to run the tests on</td>
+ * </tr>
+ * <tr>
  * <td>serviceTlsPin</td>
  * <td>TLS Pin (used by <a href="https://github.com/Flowdalic/java-pinning">java-pinning</a>)</td>
  * </tr>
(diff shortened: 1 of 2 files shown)


smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/package-info.java
```diff
@@ -80,6 +80,10 @@
  * <td>XMPP service to run the tests on</td>
  * </tr>
  * <tr>
+ * <td>Name (DNS name or IP address) of the host providing the XMPP service.</td>
```
Flowdalic, line 83: That seems wrong, shouldn't there be the name of the configuration option?

```suggestion
 * <td>host</td>
```

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/package-info.java
```diff
@@ -80,6 +80,10 @@
...
+ * <td>XMPP service to run the tests on</td>
```
Flowdalic, line 84:
```suggestion
 * <td>IP address or DNS name of the XMPP service to run the tests on</td>
```

Flowdalic (changes requested):
I think the javadoc table entry may needs a little tweaking. Otherwise, lgtm. But hadn't had my morning coffee yet. :)

smack-integration-test/src/main/java/org/igniterealtime/smack/inttest/package-info.java
guusdk: Thanks, got this change fixed up with the previous one.

PR #600, opened 2024-06-14
[core] Unique names for threads of Smack Cached Executor service

guusdk:
Having unique names for threads of Smack's own executor service can help with debugging issues.


Flowdalic:
I am a bit hesitant to add an atomic counter to thread creating. But maybe I am missing the motivation.

Looking at thread dumps, one sees

```
"Smack Cached Executor" #19
"Smack Cached Executor" #22 
"Smack Cached Executor" #26 
```
so yes, the thread names are not unique, but every thread has a unique ID. I never felt the need to make the thread names unique in such cases, so I am wondering about the motivation for doing so.

Flowdalic:
Thanks for elaborating on it. I'd like to keep it as is. At least for now.

PR #601, opened 2024-06-14
[sint] Refactor test to ensure stanza order

guusdk:
The test that's modified in this commit asserts that upon MUC join, stanzas are received in a particular order.

The previous implementation depended on several event listeners (one for presence, one for messages) that did not always fire in the same order in which the corresponding stanzas arrived. This made the approach unsuitable to reliably test the order in which stanzas arrive.

This commit stops using Smack's MUC API when trying to collect the order in which stanzas arrive. Instead, it joins a chatroom and listens for its stanzas using basic stanza handling. As this uses exactly one stanza listener, that's guaranteed to be invoked in order of stanza arrival, any synchronicity issue in the previous implementation no longer applies.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOccupantIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOccupantIntegrationTest.java
@@ -33,8 +33,10 @@ import java.util.stream.Collectors;
 import org.jivesoftware.smack.PresenceListener;
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.XMPPException;
+import org.jivesoftware.smack.filter.FromMatchesFilter;
 import org.jivesoftware.smack.packet.Message;
 import org.jivesoftware.smack.packet.Presence;
+import org.jivesoftware.smack.packet.Stanza;
 import org.jivesoftware.smack.packet.StanzaError;
 import org.jivesoftware.smack.sm.predicates.ForEveryMessage;
 import org.jivesoftware.smack.util.StringUtils;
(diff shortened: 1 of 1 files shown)


guusdk:
@Fishbowler this intends the flappy test results that we've been observing.

Fishbowler (approved):

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/MultiUserChatOccupantIntegrationTest.java
```diff
@@ -101,41 +101,35 @@ public void mucJoinEventOrderingTest() throws Exception {
...
-        mucAsSeenByTwo.addSubjectUpdatedListener((subject, from) -> {
-            results.add(subject);
-            subjectResultSyncPoint.signal(subject);
-        });
+        }, FromMatchesFilter.create(mucAddress));
 
         try {
-            Presence reflectedJoinPresence = mucAsSeenByTwo.join(nicknameTwo);
-            results.add(reflectedJoinPresence.getFrom()); // Self-presence should be second
+            // Join a MUC room, without invoking the MUC implementation in Smack that would set up 'competing' stanza
```
Flowdalic, line 113: Are you sure that you can't keep using `mucAsSeenByTwo.join(nicknameTwo)` here? I wonder what you mean by 'competing stanza listeners.

The key here seems to be that we now use `XMPPConnectino.addStanzaListener()` and not the MUC listeners anymore. But this should also work if we keep using `mucAsSeenByTwo.join(nicknameTwo)`. At least, I don't see why it should break.

guusdk: The essence of the trouble that we had with this test was apparent parallel execution of stanzas related to MUC when using Smack's MUC API. 

In my refactoring, I didn't distinguish between the MUC listeners and the listeners that are internal to the MUC implementation. Instead, I went with the approach that avoided that completely. That is guaranteed to avoid the issue (now, and in potential future changes). It also adds to the consistency of this particular implementation (in that it uses the low-level API everywhere).

PR #602, opened 2024-06-14
SMACK-943: Prevent duplicate processing of mediated invitations

guusdk:
When a MUC mediated invitation includes the `<x xmlns='jabber:x:conference'>` element (for backwards compatibility), then Smack should not fire the 'direct invitation' event listener. The value for 'inviter' that's used to invoke that listener will be false.


Flowdalic:
Merged as https://github.com/igniterealtime/Smack/commit/98dbc0ee2eb1e15ae83c7ed9025a9b3075670bb1

PR #606, opened 2024-07-03
[sinttest] XEP-0373 Integration Tests should clean-up

guusdk:
After test execution, the OpenPGP for XMPP integration tests should clean up the data published via PEP. This prevents these tests from interfering with other tests.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/ox/OXSecretKeyBackupIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/ox/OXSecretKeyBackupIntegrationTest.java
@@ -38,6 +38,7 @@ import org.jivesoftware.smackx.ox.exception.MissingUserIdOnKeyException;
 import org.jivesoftware.smackx.ox.exception.NoBackupFoundException;
 import org.jivesoftware.smackx.ox.store.definition.OpenPgpStore;
 import org.jivesoftware.smackx.ox.store.filebased.FileBasedOpenPgpStore;
+import org.jivesoftware.smackx.ox.util.OpenPgpPubSubUtil;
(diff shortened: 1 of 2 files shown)


guusdk:
I could test only `OXInstantMessagingIntegrationTest` as Openfire is not suitable to run the other test. The fix is very similar though.

I ran into this while creating tests that interacted with service discovery. When executed after the tests modified by this PR, they suddenly saw disco#items (which was unexpected/undesirable).

smack-integration-test/src/main/java/org/jivesoftware/smackx/ox/OXSecretKeyBackupIntegrationTest.java
```diff
@@ -122,38 +123,45 @@ public void test() throws InvalidAlgorithmParameterException, NoSuchAlgorithmExc
...
+            assertArrayEquals(beforeSec.getEncoded(), afterSec.getEncoded());
 
-        PGPPublicKeyRing afterPub = afterStore.getPublicKeyRing(alice, keyFingerprint);
-        assertNotNull(afterPub);
-        assertArrayEquals(beforePub.getEncoded(), afterPub.getEncoded());
+            PGPPublicKeyRing afterPub = afterStore.getPublicKeyRing(alice, keyFingerprint);
+            assertNotNull(afterPub);
+            assertArrayEquals(beforePub.getEncoded(), afterPub.getEncoded());
+        } finally {
+            Thread.sleep(10000);
```
Flowdalic, line 161: Could those `Thread.sleep()`s be dropped?

guusdk: :facepalm:

PR #607, opened 2024-07-05
SMACK-945: MUC Item actor's nick to XML

guusdk:
If an actor's nick is set in `MUCItem` this value should be added to the XMPP representation of the instance.

Note that I've not tested this change.

It is worth considering adding more changes to Smack related to this. Where it now exposes `MUCItem#actor`, `MUCItem#actorNick` should possibly be exposed too.


Flowdalic (changes requested):
Good find! :)

Could you drop the SMACK-945 reference from the commit's subject, replace it with '[muc]'  and add "Fixes SMACK-945" to the commit's body? Thanks

PR #608, opened 2024-07-13
[sint] MUC tests should test for room creation permission

guusdk:
When the service under test does not allow SINT's test users to create a room, the tests should not fail. They should be skipped instead.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
@@ -18,6 +18,7 @@
 package org.jivesoftware.smackx.muc;
 
 import java.util.List;
+import java.util.logging.Level;
 
 import org.jivesoftware.smack.SmackException;
 import org.jivesoftware.smack.XMPPException;
@@ -55,10 +56,34 @@ public abstract class AbstractMultiUserChatIntegrationTest extends AbstractSmack
 
         List<DomainBareJid> services = mucManagerOne.getMucServiceDomains();
         if (services.isEmpty()) {
-            throw new TestNotPossibleException("No MUC (XEP-45) service found");
+            throw new TestNotPossibleException("No MUC (XEP-0045) service found");
         }
 
-        mucService = services.get(0);
+        DomainBareJid needle = null;
+        for (final DomainBareJid service : services) {
+            MultiUserChat multiUserChat = null;
+            try {
+                final String roomNameLocal = String.join("-", "smack-inttest-abstract", testRunId, StringUtils.insecureRandomString(6));
+                final EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from(roomNameLocal), service.getDomain());
+                multiUserChat = mucManagerOne.getMultiUserChat(mucAddress);
+                createMuc(multiUserChat, Resourcepart.from("test"));
+                needle = service;
+                break;
+            } catch (XMPPException.XMPPErrorException | MultiUserChatException.MucAlreadyJoinedException |
+                     MultiUserChatException.MissingMucCreationAcknowledgeException | XmppStringprepException |
+                     MultiUserChatException.NotAMucServiceException e) {
+                LOGGER.log(Level.FINER, "Error creating test MUC", e);
+            } finally {
+                if (needle != null) {
+                    tryDestroy(multiUserChat);
+                }
+            }
+        }
+
+        if (needle == null) {
+            throw new TestNotPossibleException("No MUC (XEP-0045) service found that allows test users to create a new room.");
+        }
+        mucService = needle;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -55,10 +55,30 @@ public AbstractMultiUserChatIntegrationTest(SmackIntegrationTestEnvironment envi
...
+                final EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from(roomNameLocal), service.getDomain());
+                final MultiUserChat multiUserChat = mucManagerOne.getMultiUserChat(mucAddress);
+                createMuc(multiUserChat, Resourcepart.from("test"));
+                tryDestroy(multiUserChat);
+                needle = service;
+                break;
+            } catch (XMPPException.XMPPErrorException | MultiUserChatException.MucAlreadyJoinedException |
+                     MultiUserChatException.MissingMucCreationAcknowledgeException | XmppStringprepException |
+                     MultiUserChatException.NotAMucServiceException e) {
+                // Ignore, try next service.
```
Flowdalic, line 74: Always log, with a level of FINE or lower, never silently ignore.
```suggestion
               LOGGER.log(Level.FINER, "Error creating test MUC", e);
```

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -55,10 +55,30 @@ public AbstractMultiUserChatIntegrationTest(SmackIntegrationTestEnvironment envi
...
 
-        mucService = services.get(0);
+        DomainBareJid needle = null;
+        for (final DomainBareJid service : services) {
+            try {
+                final String roomNameLocal = String.join("-", "smack-inttest-abstract", testRunId, StringUtils.insecureRandomString(6));
+                final EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from(roomNameLocal), service.getDomain());
+                final MultiUserChat multiUserChat = mucManagerOne.getMultiUserChat(mucAddress);
+                createMuc(multiUserChat, Resourcepart.from("test"));
+                tryDestroy(multiUserChat);
```
Flowdalic, line 68: I'd rather put this tryDestroy() in a `finally` block and let it throw instead of catching its exceptions.

Flowdalic (changes requested):
Idea is sensible, please find my comments inline.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
guusdk: That's actually one of my own rules, that I broke for no apparent reason. Fixed.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
guusdk: I've moved that to a `finally` block, to be executed only when there was a successful creation.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -55,10 +56,34 @@ public AbstractMultiUserChatIntegrationTest(SmackIntegrationTestEnvironment envi
...
+                final String roomNameLocal = String.join("-", "smack-inttest-abstract", testRunId, StringUtils.insecureRandomString(6));
+                final EntityBareJid mucAddress = JidCreate.entityBareFrom(Localpart.from(roomNameLocal), service.getDomain());
+                multiUserChat = mucManagerOne.getMultiUserChat(mucAddress);
+                createMuc(multiUserChat, Resourcepart.from("test"));
+                needle = service;
+                break;
+            } catch (XMPPException.XMPPErrorException | MultiUserChatException.MucAlreadyJoinedException |
+                     MultiUserChatException.MissingMucCreationAcknowledgeException | XmppStringprepException |
+                     MultiUserChatException.NotAMucServiceException e) {
+                LOGGER.log(Level.FINER, "Error creating test MUC", e);
```
Flowdalic, line 75: Writing the comment "below" in MultiUserChatLowLevelIntegrationTest made me realize that we can and should be more be more precise here. Only catch `XMPPErrorException` here, and only continue the loop, i.e., do not rethrow if the error condition is `not-allowed`. I think this is the common condition that is returned if the user is not allowed to create MUC. I wonder if, depending on the exact reason why a user is not allowed to create a room, other errror conditions could also be returned. XEP-0045 does not seem to talk about it, but it feels like a possibility.

Flowdalic (changes requested):
This makes me wonder if we should specify a "dry-run" MUC join operation as extension protocol to XEP-0045, so that clients can check if an operation would be successful without having to resort to doing the real thing.

In any case, we can probably simplify the code by not catching most exceptions that are currently listed in the catch condition, and further filter down on the XMPP errror condition.

guusdk:
I'm sure if any of the requested changes are needed, or are an improvement to be honest. Frankly, I'm getting quite demotivated. Please go ahead and take ownership of this PR.

Flowdalic:
Done, merged as https://github.com/igniterealtime/Smack/commit/17d9b742fc81398e453f01ff829fff14352add05

PR #609, opened 2024-07-13
[muc] Invoke ParticipantStatusListener after revokation of membership

guusdk:
When an occupant gets its membership revoked in an members-only room, the appropriate method of registered ParticipantStatusListeners should be invoked.


Flowdalic (commented):
LGTM, please squash

guusdk:
Squashed - any reason why you choose not to do that through GitHub's UI, by the way?

Flowdalic:
Simply to encourage (and teach) best-practices.

PR #611, opened 2024-08-12
build.gradle: set miniDnsVersion version to the latest 1.0.x

stokito:
The minidns dependency use the range to pick the latest version in range of 1.0.x.
This caused some problems in the Spark that I fixed https://github.com/igniterealtime/Spark/pull/866
Anyway, the minidns now released an alpha version of the 1.1 so it shouldn't be any new versions in the 1.0.x series. So here I changed the range to the latest 1.0.5.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -146,7 +146,7 @@ allprojects {
 		// - https://issues.apache.org/jira/browse/MNG-6232
 		// - https://issues.igniterealtime.org/browse/SMACK-858
 		jxmppVersion = '[1.0.0, 1.0.999]'
-		miniDnsVersion = '[1.0.0, 1.0.999]'
+		miniDnsVersion = '1.0.5'
 		smackMinAndroidSdk = 19
 		junitVersion = '5.7.1'
 		commonsIoVersion = '2.6'


Flowdalic:
Thanks for your contribution.

However, I wonder if the real problem isn't Maven bailing out. I've looked at https://github.com/igniterealtime/Spark/pull/866#issue-2459432316 where you write

> The Maven goes into the maven-snapshots-repo and tries to resolve the 1.0.3-SNAPSHOT which is not exists there but it was referenced from some other dependency.

So this issue seems to be that something is referencing 1.0.3-SNAPSHOT. 

Wouldn't identifying the causing dependency and fixing it be the real fix?

Smack (and all my other Java projects) use ranged dependencies, if the dependency I guranteed to not break API withing the ranged version specification. This is true for MiniDNS, and, fwiw, JXMPP. This always to consume patch-level releases which often carry important fixes.

If you want to pin a dependency to a certain version, then this is usually possible with the help of the build system. I know that Gradle supports this and would be surprised if Maven did not.

With the information I have at the moment, I am unfortunately inclined to reject this change.

But thanks again for submitting it!

guusdk:
For what it's worth, I'm frequently hit by this issue (unable to download version 1.0.3-SNAPSHOT of minidns) outside of Spark. Just refreshing the Intellij workspace of a project that uses Smack can trigger it. It's a major pain.

guusdk:
I suspect that contributing factors include:
- the Sonatype repository metadata referencing the version `1.0.3-SNAPSHOT` in https://oss.sonatype.org/content/repositories/snapshots/org/minidns/minidns-core/maven-metadata.xml while it does not actually provide any content for that version (note that it's missing in the directory listing of https://oss.sonatype.org/content/repositories/snapshots/org/minidns/minidns-core/
- the IDE trying to loading many/all (?) poms for versions that match the range as defined in Smack (see screenshot)

image

guusdk:
A complicating factor is that the repository that is misbehaving (Sonatype's snapshots repo) is the same one as where Smack publishes its snapshots. If we want to consume Smack's snapshots, you will want to enable the repository.

guusdk:
I have been talking to Sonatype's support engineers. Per my request, they have modified the repository to remove the faulty reference to the non-existing 1.0.3-SNAPSHOT artifact. Hopefully, this once and for all resolves this problem.

I'm not sure if this faulty metadata gets cached in your local .m2 repository, or, if it does, if the changes from upstream automatically bubble down to your local copy. It may be needed to manually remove `org/minidns/` from your local repository for the fix to become available to you.

guusdk:
I'm not a fan of ranged versions either, but I'll leave that up to @Flowdalic

Flowdalic:
> I'm not a fan of ranged versions either, but I'll leave that up to @Flowdalic

Another advantage of ranged version is that dependent releases do not require a Smack release. For example, Smack uses two libraries I also maintain, jxmpp and MiniDNS. Not using ranged dependencies would mean that for every patch-level release of those libraries I would need to create a Smack release for Smack to benefit from the patch-level release.

So, yeah, I am really sold on ranged versions.

This PR is also just an indirect solution for the problem the PR tries to solve. It seems now that the right solution was deployed, thanks to Guus.

PR #612, opened 2024-08-12
Fix gradle plugins

stokito:
The build.gradle uses a a plugin `org.kordamp.gradle:clirr-gradle-plugin:0.2.2` that is very outdated. It comes from the `dl.bintray.com/content/aalmiray/kordamp` repo that is even not works anymore.
I tried to repair it and upgrade to latest version but that was too complicated for me. Anyway, the plugin currently is disabled. So instead I removed it for now and if anyone can restore it and enable then you may check history.
Check it's documentation https://plugins.gradle.org/plugin/org.kordamp.gradle.clirr

The `biz.aQute.bnd.builder` plugin was added in an old Gradle style so I migrated it to the new and declared in the `plugins`. I also updated it from v6.0.0 to v.6.4.0.

The `jcenter()` repo doesn't exists anymore and I removed it. The `https://plugins.gradle.org/m2/` replaced with `gradlePluginPortal()`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,11 +1,6 @@
 buildscript {
     repositories {
-        jcenter()
-		maven { url 'https://plugins.gradle.org/m2/' }
-        maven { url 'https://dl.bintray.com/content/aalmiray/kordamp' }
-    }
-    dependencies {
-		classpath "biz.aQute.bnd:biz.aQute.bnd.gradle:6.0.0"
+		gradlePluginPortal()
     }
 }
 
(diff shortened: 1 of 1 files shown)


Flowdalic:
Thanks for your contribution.

I always hoped to re-enabled clirr, since the functionality it provided was great. But it seems about time that we drop it.

Also kudos for making the changes in two commits. +1

LGTM

Flowdalic (approved):

PR #620, opened 2024-08-28
Updated: Changed version of jbosh for bosh component module, replaced…

DudkaDima:
… version 0.9.1 to 0.9.2

Thank you for your contribution!

Before creating a Pull Request, please make sure to
* read https://github.com/igniterealtime/Smack/wiki/Guidelines-for-Smack-Developers-and-Contributors
* run `gradle check` successfully in order to make sure that your code does not break any JUnit tests and is conform to the projects code style.
* (if applicable) mention any Jira issue codes (eg. `SMACK-XXX`) in the *body* of your commit message (not the header), so that Jira automatically links the PR to the issue.
* squash your commits if possible/sensible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Smack Changelog
 
+# 4.4.9 -- 2024-08-28
+
+### Changes
+
+Changed jbosh library version from 0.9.1 to 0.9.2 
+
 # 4.4.8 -- 2024-04-02
 
 ### Improvement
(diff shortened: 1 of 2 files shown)


Flowdalic:

However, that change seems unnecessary. Smack is declaring the bosh dependency using ranged dependencies, which include 0.9.2.

Your build system should automatically pick the latest viable jbosh version, which would be 0.9.2 at the time of writing this. If it does not, then you should check your build and/or potentially force the usage of 0.9.2 via your build system.

PR #628, opened 2024-09-13
[sint] Add connection pre- and postconditions

guusdk:
This adds verification that before and after a test execution of AbstractSmackIntegrationTest instances, the connections that are used are in the expected state.

When these connections are _not_ in such a state, then it is likely that one of the tests contains a bug that will affect execution of other tests. Therefor, a breaking IOException is thrown when that's detected.


Flowdalic:
Thanks, but this seems a little bit convoluted and does not mention the disconnected connections in the exception message.

My alternative proposal: https://github.com/igniterealtime/Smack/pull/645

guusdk:
I've got no strong preference for a particular implementation either way.

PR #630, opened 2024-09-18
[sint] Fix compatibility with Smack 4.5.0-beta2

guusdk:
Due to a change in Smack 4.5.0-beta2, test execution of (all) SINT tests is aborted when `FormTest` is executed.

It appears that Smack now has more strict argument validation when setting thread IDs on message stanzas. This validation should not fail for the tests that are shipped with Smack.


guusdk:
The build failure of this PR does not seem to be related to the changes in this PR.

PR #631, opened 2024-09-19
[ide] Improve junit5-platform hack (for IntelliJ)

guusdk:
The build definition uses a hack to make Eclipse play nice(r) with JUnit 5. This seems to cause issues for IntelliJ, that fails to properly setup a project, citing many of these errors:

```
Could not find org.junit.platform:junit-platform-launcher:.
```

Usage of the `.` in that error made me wonder if this is a version resolution issue. After adding an explicit version range (that covers basically any version of the dependency) the problem in Intellij goes away.


Flowdalic:
Closing in favor of https://github.com/igniterealtime/Smack/pull/632

PR #644, opened 2024-11-04
[test] Prevent hard-coded path

guusdk:
The hardcoded path used in this test conflicts with a pre-existing path by another user/process. This commit adds a bit of randomness to the directory name, similar to that used in `OXInstantMessagingManagerTest`

Fixes SMACK-955

### DIFF
--- a/smack-openpgp/src/test/java/org/jivesoftware/smackx/ox/SecretKeyBackupHelperTest.java
+++ b/smack-openpgp/src/test/java/org/jivesoftware/smackx/ox/SecretKeyBackupHelperTest.java
@@ -26,6 +26,7 @@ import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
 
 import org.jivesoftware.smack.test.util.SmackTestSuite;
+import org.jivesoftware.smack.util.StringUtils;
 
 import org.jivesoftware.smackx.ox.crypto.PainlessOpenPgpProvider;
 import org.jivesoftware.smackx.ox.element.SecretkeyElement;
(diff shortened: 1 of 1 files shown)


smack-openpgp/src/test/java/org/jivesoftware/smackx/ox/SecretKeyBackupHelperTest.java
```diff
@@ -51,7 +52,7 @@ public class SecretKeyBackupHelperTest extends SmackTestSuite {
     private static final File basePath;
 
     static {
-        basePath = new File(org.apache.commons.io.FileUtils.getTempDirectory(), "ox_secret_keys");
+        basePath = new File(org.apache.commons.io.FileUtils.getTempDirectory(), "ox_secret_keys_" + StringUtils.randomString(10));
```
Flowdalic, line 55: Let's do the idiomatic thing:

```suggestion
        basePath = Files.createTempDirectory("ox_secret_keys").toFile();
```

Flowdalic (changes requested):

smack-openpgp/src/test/java/org/jivesoftware/smackx/ox/SecretKeyBackupHelperTest.java
akrherz: Could we please move this PR along so that we can get ignite's bamboo jobs green again, thanks.

guusdk:
Florian, please adjust this as you see fit.

Flowdalic:
Fixed with https://github.com/igniterealtime/Smack/commit/2e4178d8d6eab2f0673082bd90a45556965b74ce

PR #647, opened 2025-01-30
Implement XEP-0444: Message Reactions in Smack

ismael221:
This commit adds support for XEP-0444 (Message Reactions) in Smack. Key changes include:

- Added ReactionsManager to handle reactions, including adding, removing, and listening for reactions on messages.
- Introduced ReactionsElement and Reaction classes to represent the <reactions> element and individual emoji reactions.
- Added ReactionsFilter to detect messages containing reactions.
- Implemented ReactionRestrictions to manage restrictions like max reactions per user and allowed emojis.
- Integrated reaction restrictions with XMPP service discovery.
- Added ReactionsListener for applications to handle incoming reactions.
- Included unit tests to verify functionality.

This enables emoji reactions in XMPP messages, with support for restrictions and service discovery.

Related: XEP-0444 (https://xmpp.org/extensions/xep-0444.html)

Thank you for your contribution!

Before creating a Pull Request, please make sure to
* read https://github.com/igniterealtime/Smack/wiki/Guidelines-for-Smack-Developers-and-Contributors
* run `gradle check` successfully in order to make sure that your code does not break any JUnit tests and is conform to the projects code style.
* (if applicable) mention any Jira issue codes (eg. `SMACK-XXX`) in the *body* of your commit message (not the header), so that Jira automatically links the PR to the issue.
* squash your commits if possible/sensible.

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsListener.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsListener.java
@@ -1,0 +1,31 @@
+/**
+ *
+ * Copyright 2025 Ismael Nunes Campos
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.reactions;
+
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.reactions.element.ReactionsElement;
+
+public interface ReactionsListener {
+
+
+    /**
+     * Listener method that gets called when a {@link Message} containing a {@link ReactionsElement} is received.
+     *
+     * @param message message
+     */
+    void onReactionReceived(Message message, ReactionsElement reactionsElement);
+}
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
@@ -1,0 +1,357 @@
+/**
+ *
+ * Copyright 2025 Ismael Nunes Campos
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.reactions;
+
+import org.jivesoftware.smack.AsyncButOrdered;
+import org.jivesoftware.smack.ConnectionCreationListener;
+import org.jivesoftware.smack.Manager;
+import org.jivesoftware.smack.SmackException;
+import org.jivesoftware.smack.XMPPConnection;
+import org.jivesoftware.smack.XMPPConnectionRegistry;
+import org.jivesoftware.smack.XMPPException;
+import org.jivesoftware.smack.filter.AndFilter;
+import org.jivesoftware.smack.filter.StanzaFilter;
+import org.jivesoftware.smack.filter.StanzaTypeFilter;
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smack.packet.Stanza;
+import org.jivesoftware.smack.packet.XmlElement;
+import org.jivesoftware.smackx.disco.ServiceDiscoveryManager;
+import org.jivesoftware.smackx.disco.packet.DiscoverInfo;
+import org.jivesoftware.smackx.reactions.element.Reaction;
+import org.jivesoftware.smackx.reactions.element.ReactionsElement;
+import org.jivesoftware.smackx.reactions.filter.ReactionsFilter;
+import org.jivesoftware.smackx.xdata.FormField;
+import org.jivesoftware.smackx.xdata.TextSingleFormField;
+import org.jivesoftware.smackx.xdata.form.Form;
+import org.jivesoftware.smackx.xdata.packet.DataForm;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collectors;
+
+import org.jxmpp.jid.BareJid;
+import org.jxmpp.jid.EntityBareJid;
+
+/**
+ * Manages reactions in the XMPP protocol. This class allows adding, removing, and listening for reactions
+ * on messages, as well as managing restrictions on the number of reactions per user and allowed emojis.
+ * It also allows propagating these restrictions to other clients via XMPP service discovery.
+ *
+ * This class is based on the XEP-0444 extension protocol for reactions.
+ *
+ * @author Ismael Nunes Campos
+ *
+ * @see <a href="https://xmpp.org/extensions/xep-0444.html">XEP-0444 Message Reactions</a>
+ * @see ReactionsElement
+ * @see Reaction
+ */
+public final class ReactionsManager extends Manager {
+
+    private static final Map<XMPPConnection, ReactionsManager> INSTANCES = new WeakHashMap<>();
+
+
+    static {
+        XMPPConnectionRegistry.addConnectionCreationListener(new ConnectionCreationListener() {
+            @Override
+            public void connectionCreated(XMPPConnection connection) {
+                getInstanceFor(connection);
+            }
+        });
+    }
+
+    private static final String REACTIONS_RESTRICTIONS_NAMESPACE = "urn:xmpp:reactions:0:restrictions";
+    private final Set<ReactionsListener> listeners = new CopyOnWriteArraySet<>();
+    private final AsyncButOrdered<BareJid> asyncButOrdered = new AsyncButOrdered<>();
+    private final  StanzaFilter reactionsElementFilter = new AndFilter(StanzaTypeFilter.MESSAGE,ReactionsFilter.INSTANCE);
+
+
+    /**
+     * Constructs an instance of the reactions manager and add ReactionsElement to disco features.
+     *
+     * @param connection The XMPP connection used by the manager.
+     */
+    public ReactionsManager(XMPPConnection connection) {
+        super(connection);
+        connection.addAsyncStanzaListener(this::reactionsElementListener,reactionsElementFilter);
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection);
+        sdm.addFeature(ReactionsElement.NAMESPACE);
+    }
+
+    /**
+     * Listener method for reactions elements in XMPP messages. This method is invoked when a new
+     * stanza (message) is received and attempts to extract a {@link ReactionsElement} from the message.
+     * If the element is found, it notifies the registered reaction listeners.
+     *
+     * @param packet The received XMPP stanza (message).
+     */
+    public  void reactionsElementListener(Stanza packet){
+        Message message = (Message) packet;
+        ReactionsElement reactionsElement = ReactionsElement.fromMessage(message);
+
+        if (reactionsElement != null){
+            notifyReactionListeners(message,reactionsElement);
+        }
+
+    }
+
+    /**
+     * Notifies all registered reaction listeners that a new reaction has been received. This method
+     * performs the notification in an ordered, asynchronous manner to ensure listeners are notified in
+     * the order that they were added.
+     *
+     * @param message The XMPP message that contains the reactions.
+     * @param reactionsElement The {@link ReactionsElement} containing the reactions.
+     */
+    public  void notifyReactionListeners(Message message, ReactionsElement reactionsElement) {
+        for (ReactionsListener listener : listeners) {
+            asyncButOrdered.performAsyncButOrdered(message.getFrom().asBareJid(), () -> {
+                listener.onReactionReceived(message, reactionsElement);
+            });
+        }
+    }
+
+
+    /**
+     * Retrieves the instance of the ReactionsManager for the given XMPP connection.
+     *
+     * @param connection The XMPP connection.
+     * @return The ReactionsManager instance for the connection.
+     */
+    public static synchronized ReactionsManager getInstanceFor(XMPPConnection connection) {
+        ReactionsManager reactionsManager = INSTANCES.get(connection);
+
+        if (reactionsManager == null) {
+            reactionsManager = new ReactionsManager(connection);
+            INSTANCES.put(connection, reactionsManager);
+        }
+        return reactionsManager;
+    }
+
+    /**
+     * Checks whether the user supports reactions.
+     *
+     * @param jid The JID of the user.
+     * @return {@code true} if the user supports reactions, otherwise {@code false}.
+     * @throws XMPPException.XMPPErrorException If an XMPP error occurs.
+     * @throws SmackException.NotConnectedException If the connection is not established.
+     * @throws InterruptedException If the operation is interrupted.
+     * @throws SmackException.NoResponseException If no response is received from the server.
+     */
+    public boolean userSupportsReactions(EntityBareJid jid) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException,
+                    InterruptedException, SmackException.NoResponseException {
+        return ServiceDiscoveryManager.getInstanceFor(connection()).supportsFeature(jid,ReactionsElement.NAMESPACE);
+    }
+
+    /**
+     * Checks whether the server supports reactions.
+     *
+     * @return {@code true} if the server supports reactions, otherwise {@code false}.
+     * @throws XMPPException.XMPPErrorException If an XMPP error occurs.
+     * @throws SmackException.NotConnectedException If the connection is not established.
+     * @throws InterruptedException If the operation is interrupted.
+     * @throws SmackException.NoResponseException If no response is received from the server.
+     */
+    public boolean serverSupportsReactions()
+                    throws XMPPException.XMPPErrorException, SmackException.NotConnectedException, InterruptedException,
+                    SmackException.NoResponseException {
+        return ServiceDiscoveryManager.getInstanceFor(connection())
+                        .serverSupportsFeature(ReactionsElement.NAMESPACE);
+    }
+
+    /**
+     * Adds reactions to a message.
+     *
+     * @param message           The message builder where the reactions will be added.
+     * @param emojis            The list of emojis to be added as reactions.
+     * @param originalMessageId The ID of the original message being reacted to.
+     * @param restrictions      The reaction restrictions such as max reactions per user and allowed emojis.
+     * @throws IllegalArgumentException If the reactions exceed the allowed limit or if any emoji is not allowed.
+     */
+    public static void addReactionsToMessage(Message message, List<String> emojis,
+                    String originalMessageId, ReactionRestrictions restrictions){
+        List<Reaction> reactions = new ArrayList<>();
+
+        if (restrictions != null) {
+
+            if (emojis.size() > restrictions.getMaxReactionsPerUser()) {
+                throw new IllegalArgumentException("Exceeded maximum number of reactions per user");
+            }
+
+
+            for (String emoji : emojis) {
+                if (!restrictions.getAllowedEmojis().contains(emoji)) {
+                    throw new IllegalArgumentException("Emoji " + emoji + " is not allowed");
+                }
+            }
+        }
+
+
+        for (String emoji : emojis) {
+            Reaction reaction = new Reaction(emoji);
+            reactions.add(reaction);
+        }
+
+        ReactionsElement reactionsElement = new ReactionsElement(reactions, originalMessageId);
+
+        message.addExtension(reactionsElement);
+
+    }
+
+    /**
+     * Adds a reactions' listener.
+     *
+     * @param listener The reactions listener to be added.
+     */
+    public synchronized void addReactionsListener(ReactionsListener listener){
+        listeners.add(listener);
+    }
+
+    /**
+     * Removes a reactions listener.
+     *
+     * @param listener The reactions listener to be removed.
+     */
+    public synchronized void removeReactionsListener(ReactionsListener listener){
+        listeners.remove(listener);
+    }
+
+
+    /**
+     * Creates a form for reaction restrictions, including the max number of reactions per user
+     * and the list of allowed emojis.
+     *
+     * @param maxReactionsPerUser The maximum number of reactions allowed per user.
+     * @param allowedEmojis The list of allowed emojis.
+     * @return The reaction restrictions form.
+     */
+    public static DataForm createReactionRestrictionsForm(int maxReactionsPerUser, List<String> allowedEmojis) {
+
+        DataForm.Builder builder = DataForm.builder();
+        builder.setFormType(String.valueOf(DataForm.Type.result));
+        builder.addField(
+                        FormField.buildHiddenFormType("urn:xmpp:reactions:0:restrictions")
+        );
+        builder.addField(
+                        FormField.builder("max_reactions_per_user").setValue(String.valueOf(maxReactionsPerUser))
+                                        .build()
+        );
+
+        FormField.Builder<TextSingleFormField, TextSingleFormField.Builder> allowlistFieldBuilder = FormField.builder("allowlist");
+        for (String emoji : allowedEmojis) {
+            Reaction reaction = new Reaction(emoji);
+            FormField.builder("value").setValue((CharSequence) reaction);
+        }
+        builder.addField(allowlistFieldBuilder.build());
+
+        return builder.build();
+    }
+
+    /**
+     * Advertises reaction restrictions to a given XMPP server.
+     *
+     * @param connection The XMPP connection.
+     * @param maxReactionsPerUser The maximum number of reactions allowed per user.
+     * @param allowedEmojis The list of allowed emojis.
+     */
+    public void advertiseReactionRestrictions(XMPPConnection connection, int maxReactionsPerUser, List<String> allowedEmojis) {
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection);
+        DataForm restrictionsForm = createReactionRestrictionsForm(maxReactionsPerUser, allowedEmojis);
+        sdm.addExtendedInfo(restrictionsForm);
+
+        sdm.addFeature(ReactionsElement.NAMESPACE);
+    }
+
+    /**
+     * Retrieves the reaction restrictions for a given user.
+     *
+     * @param jid The JID of the user.
+     * @return The reaction restrictions for the user.
+     * @throws XMPPException.XMPPErrorException If an XMPP error occurs.
+     * @throws SmackException.NotConnectedException If the connection is not established.
+     * @throws InterruptedException If the operation is interrupted.
+     * @throws SmackException.NoResponseException If no response is received from the server.
+     */
+    public ReactionRestrictions getReactionRestrictions(EntityBareJid jid) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException,
+                    InterruptedException, SmackException.NoResponseException {
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection());
+        DiscoverInfo discoverInfo = sdm.discoverInfo(jid);
+
+
+        for (XmlElement extension : discoverInfo.getExtensions()) {
+            if (extension instanceof DataForm) {
+                DataForm dataForm = (DataForm) extension;
+                FormField formTypeField = dataForm.getField("FORM_TYPE");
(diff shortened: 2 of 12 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,357 @@
...
+     * and the list of allowed emojis.
+     *
+     * @param maxReactionsPerUser The maximum number of reactions allowed per user.
+     * @param allowedEmojis The list of allowed emojis.
+     * @return The reaction restrictions form.
+     */
+    public static DataForm createReactionRestrictionsForm(int maxReactionsPerUser, List<String> allowedEmojis) {
+
+        DataForm.Builder builder = DataForm.builder();
+        builder.setFormType(String.valueOf(DataForm.Type.result));
```
Flowdalic, line 248: That `String.valueOf` can probably be dropped.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,357 @@
...
+
+    }
+
+    /**
+     * Adds a reactions' listener.
+     *
+     * @param listener The reactions listener to be added.
+     */
+    public synchronized void addReactionsListener(ReactionsListener listener){
+        listeners.add(listener);
```
Flowdalic, line 224: We usually expose the return value of `add()`

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    /**
+     * Removes a reactions listener.
+     *
+     * @param listener The reactions listener to be removed.
+     */
+    public synchronized void removeReactionsListener(ReactionsListener listener){
+        listeners.remove(listener);
```
Flowdalic, line 233: We usually expose the return value of `remove()`

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,357 @@
...
+        );
+        builder.addField(
+                        FormField.builder("max_reactions_per_user").setValue(String.valueOf(maxReactionsPerUser))
+                                        .build()
+        );
+
+        FormField.Builder<TextSingleFormField, TextSingleFormField.Builder> allowlistFieldBuilder = FormField.builder("allowlist");
+        for (String emoji : allowedEmojis) {
+            Reaction reaction = new Reaction(emoji);
+            FormField.builder("value").setValue((CharSequence) reaction);
```
Flowdalic, line 260: This cast to `CharSequence` probably does not what you want.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/provider/ReactionsElementProvider.java
```diff
@@ -0,0 +1,52 @@
...
+public class ReactionsElementProvider extends ExtensionElementProvider<ReactionsElement> {
+    @Override
+    public ReactionsElement parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException, SmackParsingException, ParseException {
+        String id = parser.getAttributeValue(null, "id");
+        List<Reaction> reactions = new ArrayList<>();
+
+        while (true) {
+            XmlPullParser.Event tag = parser.next();
+
+            if (tag == XmlPullParser.Event.END_ELEMENT && parser.getName().equals(ReactionsElement.ELEMENT)) {
```
Flowdalic, line 41: Please follow the idiom in the other Providers where we break out of the lop if `END_ELEMENT` and depth are correct.

smack-experimental/src/test/java/org/jivesoftware/smackx/reactions/ReactionTest.java
```diff
@@ -0,0 +1,164 @@
...
+        MessageBuilder messageBuilder = StanzaBuilder.buildMessage();
+        Message message = messageBuilder
+                        .setBody("Hello")
+                        .ofType(Message.Type.chat)
+                        .to("teste@domain.com")
+                        .build();
+
+        ReactionsManager.addReactionsToMessage(message, emojis, messageId, null);
+
+        ReactionsElement reactionsElement = (ReactionsElement) message.getExtensionElement(ReactionsElement.ELEMENT, ReactionsElement.NAMESPACE);
```
Flowdalic, line 108: Casts are bad, better use the type safe approach (may need adjustments)
```suggestion
        ReactionsElement reactionsElement = message.getExtensionElement(ReactionsElement.class);
```

smack-experimental/src/test/java/org/jivesoftware/smackx/reactions/ReactionTest.java
```diff
@@ -0,0 +1,164 @@
...
+        ReactionsElement reactionsElement = new ReactionsElement(reactions, messageId);
+        message.addExtension(reactionsElement);
+
+        // Act: Call the listener
+        XMPPConnection connection = mock(XMPPConnection.class); // Mock of XMPP connection
+        ReactionsManager reactionsManager = new ReactionsManager(connection);
+        reactionsManager.reactionsElementListener(message);
+
+        // Assertions: Ensure that the message contains the reactions element
+        assertNotNull(message.getExtensionElement(ReactionsElement.ELEMENT, ReactionsElement.NAMESPACE));
```
Flowdalic, line 144:
```suggestion
        assertNotNull(message.getExtensionElement(ReactionsElement.class));
```

Flowdalic (changes requested):
Thanks for your contribution. Gave your PR a quick review from the top of my head. There is probably more.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
ismael221: I couldn't drop it, as it requires a String

`   public Builder setFormType(String formType) {
            FormField formField = FormField.buildHiddenFormType(formType);
            return addField(formField);
        }`

ismael221:
Hi Flowdalic! I've made all the requested changes from the previous review and addressed the feedback provided. Additionally, the code passed the gradlew check after the updates. Could you please review the PR again? Thank you for your attention!"

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/element/Reaction.java
```diff
@@ -0,0 +1,106 @@
...
+ * is used to handle the individual reaction data, including the emoji and its associated message.
+ * It is an extension element used in XMPP messages.
+ *
+ * @see ExtensionElement
+ * @see Message
+ */
+public class Reaction implements ExtensionElement {
+
+    public static final String ELEMENT = "reaction";
+    public static final String NAMESPACE = StreamOpen.CLIENT_NAMESPACE;
```
Flowdalic, line 36: That's not correct, the namespace of `<reaction/>` is not `StreamOpen.CLIENT_NAMESPACE`.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/element/Reaction.java
```diff
@@ -0,0 +1,106 @@
...
+        return xml;
+    }
+
+    /**
+     * Retrieves the Reaction extension from an XMPP message.
+     *
+     * @param message The XMPP message from which to extract the reaction.
+     * @return The Reaction extension from the message, or {@code null} if not present.
+     */
+    public static Reaction fromMessage(Message message) {
```
Flowdalic, line 103: I don't think we need this method.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/element/ReactionsElement.java
```diff
@@ -0,0 +1,123 @@
...
+     * @return The XML string builder with the XML representation of the element.
+     */
+    @Override
+    public XmlStringBuilder toXML(XmlEnvironment xmlEnvironment) {
+        XmlStringBuilder xml = new XmlStringBuilder(this);
+        xml.attribute("id", id);
+
+        xml.rightAngleBracket();
+        for (Reaction reaction : reactions) {
+            if (reaction != null) {
```
Flowdalic, line 106: Shouldn't be possible that `reactions` contains `null` values and this should be an invariant.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/filter/ReactionsFilter.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Message Reactions filter class.
+ *
+ * @see <a href="http://xmpp.org/extensions/xep-0313.html">XEP-0444: Message
+ *      Reactions</a>
+ * @author Ismael Nunes Campos
+ *
+ */
+public final class ReactionsFilter extends StanzaExtensionFilter {
```
Flowdalic, line 32: I don't see the value of this class if the same could be achieved by `ExtensionElementFilter(ReactionsElement.class)`.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/provider/ReactionsElementProvider.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public ReactionsElement parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException, SmackParsingException, ParseException {
+        String id = parser.getAttributeValue(null, "id");
+        List<Reaction> reactions = new ArrayList<>();
+
+        outerloop: while (true) {
+            XmlPullParser.Event tag = parser.next();
+
+            if (tag == XmlPullParser.Event.START_ELEMENT) {
```
Flowdalic, line 54: switch/case here please

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/provider/ReactionsElementProvider.java
```diff
@@ -0,0 +1,69 @@
...
+                if (parser.getName().equals(Reaction.ELEMENT)) {
+
+                    String emoji = parser.nextText();
+                    reactions.add(new Reaction(emoji));
+                }
+            } else if (tag == XmlPullParser.Event.END_ELEMENT) {
+                if (parser.getName().equals(ReactionsElement.ELEMENT) && parser.getDepth() == initialDepth) {
```
Flowdalic, line 61: The `parser.getName().equals(ReactionsElement.ELEMENT)` is unnecessary

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * @throws InterruptedException If the operation is interrupted.
+     * @throws SmackException.NoResponseException If no response is received from the server.
+     */
+    public ReactionRestrictions getReactionRestrictions(EntityBareJid jid) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException,
+                    InterruptedException, SmackException.NoResponseException {
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection());
+        DiscoverInfo discoverInfo = sdm.discoverInfo(jid);
+
+
+        for (XmlElement extension : discoverInfo.getExtensions()) {
```
Flowdalic, line 306: Use `getExtensions(DataForm.class)`, this also avoids the instanceof check in the line below

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * Retrieves the reaction restrictions for a given user.
+     *
+     * @param jid The JID of the user.
+     * @return The reaction restrictions for the user.
+     * @throws XMPPException.XMPPErrorException If an XMPP error occurs.
+     * @throws SmackException.NotConnectedException If the connection is not established.
+     * @throws InterruptedException If the operation is interrupted.
+     * @throws SmackException.NoResponseException If no response is received from the server.
+     */
+    public ReactionRestrictions getReactionRestrictions(EntityBareJid jid) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException,
```
Flowdalic, line 300: `EntityBareJId` here is wrong, you want to be able to query all XMPP addresses here (at least, that is my interpretation of the XEP).

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * @param connection The XMPP connection.
+     * @param maxReactionsPerUser The maximum number of reactions allowed per user.
+     * @param allowedEmojis The list of allowed emojis.
+     */
+    public void advertiseReactionRestrictions(XMPPConnection connection, int maxReactionsPerUser, List<String> allowedEmojis) {
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection);
+        DataForm restrictionsForm = createReactionRestrictionsForm(maxReactionsPerUser, allowedEmojis);
+        sdm.addExtendedInfo(restrictionsForm);
+
+        sdm.addFeature(ReactionsElement.NAMESPACE);
```
Flowdalic, line 287: This is already done in the class' constructor, why do it again here?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
Flowdalic: That's because `DataForm.Type.result` is the wrong value. Look at XEP-0440 again what the right form type is.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * @param allowedEmojis The list of allowed emojis.
+     * @return The reaction restrictions form.
+     */
+    public DataForm createReactionRestrictionsForm(int maxReactionsPerUser, List<String> allowedEmojis) {
+
+        DataForm.Builder builder = DataForm.builder();
+        builder.setFormType(String.valueOf(DataForm.Type.result));
+
+        builder.addField(
+                        FormField.builder("max_reactions_per_user").setValue(String.valueOf(maxReactionsPerUser))
```
Flowdalic, line 261: I see `FormField.Builder.setValue(int)` in Smack, so the `String.valueOf` appears unnecessary

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+                                        .build()
+        );
+
+        FormField.Builder<TextSingleFormField, TextSingleFormField.Builder> allowlistFieldBuilder = FormField.builder("allowlist");
+        for (String emoji : allowedEmojis) {
+            Reaction reaction = new Reaction(emoji);
```
Flowdalic, line 267: Why do you wrap the emoji in a `Reaction` just to extract the emoji again one line later?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+     * to the original message identified by its ID. It ensures that the number of reactions and the emojis used comply
+     * with the given restrictions.</p>
+     *
+     * @param emojis              The list of emojis to be added as reactions.
+     * @param originalMessageId   The ID of the original message being reacted to. This ID will be used for the {@link ReactionsElement}.
+     * @param restrictions        The reaction restrictions, such as the maximum number of reactions per user and allowed emojis.
+     * @return                   A message builder containing the reactions to the original message.
+     * @throws IllegalArgumentException If the number of reactions exceeds the allowed limit or if any emoji is not allowed.
+     */
+    public MessageBuilder createMessageWithReactions(List<String> emojis, String originalMessageId, ReactionRestrictions restrictions) {
```
Flowdalic, line 191: Any reason why this method is not `static`?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+        public int getMaxReactionsPerUser() {
+            return maxReactionsPerUser;
+        }
+
+        /**
+         * Retrieves the list of allowed emojis.
+         *
+         * @return The list of allowed emojis.
+         */
+        public List<String> getAllowedEmojis() {
```
Flowdalic, line 360: This should return a `Set` for fast lookup

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+        }
+        return null;
+    }
+
+    /**
+     * Represents the reaction restrictions for a user or XMPP server.
+     */
+    public static class ReactionRestrictions {
+        private final int maxReactionsPerUser;
+        private final List<String> allowedEmojis;
```
Flowdalic, line 333: This should be a `Set`

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+    public ReactionRestrictions getReactionRestrictions(EntityBareJid jid) throws XMPPException.XMPPErrorException, SmackException.NotConnectedException,
+                    InterruptedException, SmackException.NoResponseException {
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection());
+        DiscoverInfo discoverInfo = sdm.discoverInfo(jid);
+
+
+        for (XmlElement extension : discoverInfo.getExtensions()) {
+            if (extension instanceof DataForm) {
+                DataForm dataForm = (DataForm) extension;
+                FormField formTypeField = dataForm.getField("FORM_TYPE");
```
Flowdalic, line 309: Use `DataForm.getFormType()` to select the right form

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/element/Reaction.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Represents a reaction in the form of an emoji to be associated with a message in XMPP. This class
+ * is used to handle the individual reaction data, including the emoji and its associated message.
+ * It is an extension element used in XMPP messages.
+ *
+ * @see ExtensionElement
+ * @see Message
+ */
+public class Reaction implements ExtensionElement {
```
Flowdalic, line 33: This should probably extend `NamedElement`

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    /**
+     * Listener method for reactions elements in XMPP messages. This method is invoked when a new
+     * stanza (message) is received and attempts to extract a {@link ReactionsElement} from the message.
+     * If the element is found, it notifies the registered reaction listeners.
+     *
+     * @param packet The received XMPP stanza (message).
+     */
+    public void reactionsElementListener(Stanza packet) {
```
Flowdalic, line 106: Why is this method `public`?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+        Message message = (Message) packet;
+        ReactionsElement reactionsElement = ReactionsElement.fromMessage(message);
+        if (reactionsElement != null) {
```
Flowdalic, line 109: That should never happen, otherwise, the filter wouldn't work properly.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+    private final StanzaFilter reactionsElementFilter = new AndFilter(StanzaTypeFilter.MESSAGE, ReactionsFilter.INSTANCE);
+
+    /**
+     * Constructs an instance of the reactions manager and add ReactionsElement to disco features.
+     *
+     * @param connection The XMPP connection used by the manager.
+     */
+    public ReactionsManager(XMPPConnection connection) {
+        super(connection);
+        connection.addAsyncStanzaListener(this::reactionsElementListener, reactionsElementFilter);
```
Flowdalic, line 94: This stanza listener shouldn't only be setup if there are actual reaction listeners.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+
+    /**
+     * Notifies all registered reaction listeners that a new reaction has been received. This method
+     * performs the notification in an ordered, asynchronous manner to ensure listeners are notified in
+     * the order that they were added.
+     *
+     * @param message The XMPP message that contains the reactions.
+     * @param reactionsElement The {@link ReactionsElement} containing the reactions.
+     */
+    public void notifyReactionListeners(Message message, ReactionsElement reactionsElement) {
```
Flowdalic, line 122: Why is this method `public`? Why does it exist at all if there is only exactly one call site?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+        for (ReactionsListener listener : listeners) {
+            asyncButOrdered.performAsyncButOrdered(message.getFrom().asBareJid(), () -> {
```
Flowdalic, line 124: Using `asyncButOrdered` here is too late, since the initial listener was already async the messages for a particular jid could be already out of order.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
```diff
@@ -0,0 +1,364 @@
...
+    /**
+     * Constructs an instance of the reactions manager and add ReactionsElement to disco features.
+     *
+     * @param connection The XMPP connection used by the manager.
+     */
+    public ReactionsManager(XMPPConnection connection) {
+        super(connection);
+        connection.addAsyncStanzaListener(this::reactionsElementListener, reactionsElementFilter);
+        ServiceDiscoveryManager sdm = ServiceDiscoveryManager.getInstanceFor(connection);
+        sdm.addFeature(ReactionsElement.NAMESPACE);
```
Flowdalic, line 96: I don't think it is sensible to announce the feature right away. What if the Smack user did not setup a listener or isn't even aware of reactions? Then Smack would still announce the feature signalling to others that I can understand reactions.
A typical Smack idiom would be to only start announcing the feature once a reaction listener has been setup.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsListener.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jivesoftware.smackx.reactions.element.ReactionsElement;
+
+public interface ReactionsListener {
+
+    /**
+     * Listener method that gets called when a {@link Message} containing a {@link ReactionsElement} is received.
+     * @param reactionsElement reactionsElement
+     * @param message message
+     */
+    void onReactionReceived(Message message, ReactionsElement reactionsElement);
```
Flowdalic, line 30: Let's make the API of `ReactionsListener` more user friendly. The user is primary interested in two aspects: the ID of the message that was reacted upon, and the reactions itself. Everything else is secondary.
```suggestion
    void onReactionReceived(String id, List<String> reactions, ReactionsElement reactionsElement, Message message);
```

Also make sure to document what `message` is. Is it the Message the reaction arrived with, or is the the message on which was reacted on?

smack-experimental/src/test/java/org/jivesoftware/smackx/reactions/ReactionTest.java
```diff
@@ -0,0 +1,169 @@
...
+     * <p>
+     * This test validates the creation of a `ReactionsElement` from an XML
+     * and checks if the message ID and emoji reactions are correctly extracted.
+     *
+     * @throws XmlPullParserException If an error occurs during XML parsing.
+     * @throws IOException If an I/O error occurs during parsing.
+     * @throws SmackParsingException If a failure occurs while parsing the XML.
+     */
+    @Test
+    void testReaction() throws XmlPullParserException, IOException, SmackParsingException {
```
Flowdalic, line 68: All parsing related tests should be parameterized:
```
    @ParameterizedTest
    @EnumSource(SmackTestUtil.XmlPullParserKind.class)
```

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/provider/ReactionsElementProvider.java
```diff
@@ -0,0 +1,69 @@
...
+ * @see <a href="http://xmpp.org/extensions/xep-0444.html">XEP-0444: Message
+ *       Reactions</a>
+ *
+ *  @author Ismael Nunes Campos
+ */
+public class ReactionsElementProvider extends ExtensionElementProvider<ReactionsElement> {
+
+    @Override
+    public ReactionsElement parse(XmlPullParser parser, int initialDepth, XmlEnvironment xmlEnvironment) throws XmlPullParserException, IOException, SmackParsingException, ParseException {
+        String id = parser.getAttributeValue(null, "id");
```
Flowdalic, line 48:
```suggestion
        String id = parser.getAttributeValue("id");
```

smack-experimental/src/test/java/org/jivesoftware/smackx/reactions/ReactionTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        // Simulate a message with reactions
+        Message message = StanzaBuilder.buildMessage().build();
+        ReactionsElement reactionsElement = new ReactionsElement(reactions, messageId);
+        message.addExtension(reactionsElement);
+
+        // Act: Call the listener
+        XMPPConnection connection = mock(XMPPConnection.class); // Mock of XMPP connection
+        ReactionsManager reactionsManager = new ReactionsManager(connection);
+        reactionsManager.reactionsElementListener(message);
```
Flowdalic, line 139: How does any of this test the reactions element listener?

smack-experimental/src/test/java/org/jivesoftware/smackx/reactions/ReactionTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        ReactionsElement reactionsFromMessage = ReactionsElement.fromMessage(message);
+        // Assertions: Ensure that the message contains the reactions element
+        assertNotNull(reactionsFromMessage);
+
+    }
+
+    @Test
+    void testXMLParsing() {
```
Flowdalic, line 148: This does not test parsing, this tests XML serialization.

Flowdalic (changes requested):

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/element/Reaction.java
ismael221: I used this because the XEP specification doesn't have a namespace, but if I don't add a namespace, the tests fail and I can't parse it. In this case, should I correct the implementation to not use ReactionElement or is there another namespace to put in cases that are empty? because even an empty string is not accepted

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
ismael221: I don't think that I fully understand what you mean

ismael221: Would this be a good approach?


`  public ReactionsManager(XMPPConnection connection) {
        super(connection);

        connection.addAsyncStanzaListener(new StanzaListener() {
            @Override public void processStanza(Stanza packet)
                            throws SmackException.NotConnectedException, InterruptedException,
                            SmackException.NotLoggedInException {
                final Message message = (Message) packet;
                final ReactionsElement reactionsElement = ReactionsElement.fromMessage(message);
                final Set<Reaction> reactionsSet = reactionsElement.getReactions();
                final Set<String> reactionsStrings = new HashSet<>();
                
                for (Reaction reaction : reactionsSet) {
                    reactionsStrings.add(reaction.getEmoji());
                }

                asyncButOrdered.performAsyncButOrdered(message.getFrom().asBareJid(), () -> {
                    for (ReactionsListener l: listeners){
                        l.onReactionReceived(message.getFrom().asBareJid().toString(),reactionsStrings,reactionsElement,message);
                    }
                });
            }
        }, REACTIONS_FILTER);

        serviceDiscoveryManager = ServiceDiscoveryManager.getInstanceFor(connection);
    }`

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsListener.java
Flowdalic: Why was this marked as resolved if it wasn't resolved?

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
Flowdalic: > Would this be a good approach?

No, only add the stanza listener once the first reaction listener is added, and remove the stanza listener once the last reaction listener got removed.

smack-experimental/src/main/java/org/jivesoftware/smackx/reactions/ReactionsManager.java
Flowdalic: Why did you mark things as resolved if they are not?

ismael221:
Ready for review

PR #648, opened 2025-02-18
 Implement XEP-0461: Message Replies

ismael221:
This commit adds support for XEP-0461 (Message Replies) in Smack. Key changes include:

    Introduced ReplyManager to handle message replies, including adding, removing, and listening for replies.
    Added ReplyElement  class to represent the reply element.
    Implemented ReplyFilter to detect messages containing replies.
    Integrated reply handling with XMPP service discovery.
    Added ReplyListener for applications to handle incoming replies.
    Included unit tests to verify functionality.

This enables message reply functionality in XMPP, with support for service discovery.

Related: XEP-0461 (https://xmpp.org/extensions/xep-0461.html)

### DIFF
--- a/smack-experimental/src/main/java/org/jivesoftware/smackx/reply/ReplyListener.java
+++ b/smack-experimental/src/main/java/org/jivesoftware/smackx/reply/ReplyListener.java
@@ -1,0 +1,33 @@
+/**
+ *
+ * Copyright 2025 Ismael Nunes Campos
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smackx.reply;
+
+import org.jivesoftware.smack.packet.Message;
+import org.jivesoftware.smackx.reply.element.ReplyElement;
+
+public interface ReplyListener {
+
+    /**
+     * Listener method that gets called when a {@link Message} containing a {@link ReplyElement} is received.
+     *
+     * @param message message
+     * @param reply Reply element
+     * @param replyBody body that is marked as reply
+     */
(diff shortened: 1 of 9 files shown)


smack-experimental/src/main/java/org/jivesoftware/smackx/reply/ReplyListener.java
```diff
@@ -0,0 +1,34 @@
...
+public interface ReplyListener {
+
+    /**
+     * Listener method that gets called when a {@link Message} containing a {@link ReplyElement} is received.
+     *
+     * @param message message
+     * @param reply Reply element
+     * @param replyBody body that is marked as reply
+     */
+    void onReplyReceived(Message message, ReplyElement reply, String replyBody);
```
stokito, line 32: What is a usage scenario for the listener?

smack-experimental/src/main/java/org/jivesoftware/smackx/reply/element/ReplyElement.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jivesoftware.smack.packet.XmlEnvironment;
+import org.jivesoftware.smack.util.XmlStringBuilder;
+
+
+public class ReplyElement implements ExtensionElement {
+
+    public static final String NAMESPACE = "urn:xmpp:reply:0";
+    public static final String ELEMENT = "reply";
+
+    private final String replyTo;
```
stokito, line 30: This filed can be a FullJid or for 1:1 chats it can be a BareJid. Also thei field can be null. I'm not sure why to bother to set the field if we always can get the author of the original comment by the message ID. But this is a question to the XEP, just saying.

stokito (approved):

smack-experimental/src/main/java/org/jivesoftware/smackx/reply/element/ReplyElement.java
ismael221: Good point. According to the XEP the field may contain a FullJid or BareJid and it may also be absent. Here we only expose the value if it is present in the message stanza.

smack-experimental/src/main/java/org/jivesoftware/smackx/reply/ReplyListener.java
ismael221: The listener is triggered when an incoming message contains a ReplyElement.
A common scenario is messaging applications where users reply to a specific message. The listener allows the application to extract the original referenced message and display the reply context in the chat UI, similar to how replies work in apps like WhatsApp or Telegram.

PR #657, opened 2025-08-22
[sinttest] Fix config for custom debugger

guusdk:
The provided `debuggerString` value is split into a implementation identifier, and options.

The implementation identifier part (instead of the whole `debuggerString` value) should be used to instantiate a custom class. This obviously was intended, as the `debuggerOptions` value would otherwise be unusable.


guusdk:
Rebased, fixed commit message.

PR #666, opened 2025-09-02
[sinttest] Don't assume MUC service supports config options

guusdk:
Prior to this commit, executing MUC tests could cause the test execution to crash with an IllegalStateException "No field named [...]".

With this commit, the corresponding test instead is marked as 'not possible'.

### DIFF
--- a/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
+++ b/smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
@@ -190,10 +190,14 @@ public abstract class AbstractMultiUserChatIntegrationTest extends AbstractSmack
      * semi-anonymous (a value of "moderators"), or fully anonymous (a value of "none", not shown here).
      * </blockquote>
      */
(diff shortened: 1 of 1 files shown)


guusdk:
I have wrapped `MucConfigurationNotSupportedException` in `TestNotPossibleException` to have a uniform way of reporting the issue (other tests do the same). This leads to a proper aggregate in the test run results.

smack-integration-test/src/main/java/org/jivesoftware/smackx/muc/AbstractMultiUserChatIntegrationTest.java
```diff
@@ -207,29 +211,42 @@ static void createNonAnonymousMuc(MultiUserChat muc, Resourcepart resourceName)
...
+    static void createPasswordProtectedMuc(MultiUserChat muc, Resourcepart resourceName, String password) throws SmackException.NoResponseException, XMPPException.XMPPErrorException, InterruptedException, MultiUserChatException.MucAlreadyJoinedException, SmackException.NotConnectedException, MultiUserChatException.MissingMucCreationAcknowledgeException, MultiUserChatException.NotAMucServiceException, TestNotPossibleException {
         muc.create(resourceName);
         Form configForm = muc.getConfigurationForm();
         FillableForm answerForm = configForm.getFillableForm();
-        answerForm.setAnswer("muc#roomconfig_passwordprotectedroom", true);
+
+        if (answerForm.hasField("muc#roomconfig_passwordprotectedroom")) {
+            // Prosody doesn't seem to use this field. Setting a room secret is likely 'enough'.
```
Flowdalic, line 234: After reading xep45, this seems like a bug in prosody to me.

Until its fixed, it should be guarded by a newly introduced 'prosody' `CompatilibytMode` (akin the two existing ones `standardsCompliant` and `ejabberd`.

PR #675, opened 2025-10-03
xdata: fix submit of cancel form

guusdk:
A cancel form isn't a fillable form, leading to exceptions. This commit works around the problem by directly sending the form, rather than wrapping it in a FillableForm first.


guusdk:
The build failure relates to a Checkstyle validation that relates to the order of import statements. After an extra attempt, I have no desire to spend more time into figuring out what the desired order of those statements is.

Flowdalic:
Closing in favor of https://github.com/igniterealtime/Smack/commit/fd2da2795ffb52f443e9ecfb1020ba19c2486002

PR #676, opened 2025-10-03
[pubsub] Detection of Pub/Sub service based on required feature

guusdk:
To detect a Pub/Sub service, Smack asserts that a service that advertises a pubsub identity also advertises a feature.

The feature that was used is contentious: although it is used in examples of XEP-0060, it is not used in normative text (nor is it defined as a known feature with the XSF Registrar).

In this commit, that feature is replaced by the 'publish' feature, which is one of two features that are required by the XEP to be supported.

fixes SMACK-957


guusdk:
Using the `publish` feature instead of the `subscribe` feature (the other one that's mandatory) isn't completely arbitrary. My rationale is that when a user is interested to find _any_ pub/sub service, it is more likely to be used to publish something than to subscribe to something. Given that both are mandatory, the difference may not be significant. Checking for _both_ may be a suitable alternative to this.

guusdk:
I have now redone this commit to depend on `subscribe` rather than `publish`. The `publish` feature is defined as mandatory in by section 10 of XEP-0060, but is described as being optional in section 7.1.1. There is no such ambiguity around `subscribe`, making it a safer choice for our purpose.

guusdk:
After my last change, it was pointed out that the `subscribe` feature can be considered to be optional, too. That disqualifies that feature as one to search for, too.

Perhaps the best approach is to _only_ search for an identity, and not a feature. That appears to be most compatible with the current version of the XEP.

PR #692, opened 2026-01-21
XEP-0478: Stream Limits Advertisement: initial support

stokito:
[XEP-0478: Stream Limits Advertisement](https://xmpp.org/extensions/xep-0478.html)

I added the parsing of the `<limits>`:
```xml
<limits xmlns='urn:xmpp:stream-limits:0'>
  <max-bytes>262144</max-bytes>
  <idle-seconds>840</idle-seconds>
</limits>
```
It doesn't affect any logic, I'm not sure where to put the limits check.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
@@ -1,0 +1,90 @@
+/*
+ *
+ * Copyright © 2014-2026 Florian Schmaus
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jivesoftware.smack.packet;
+
+import javax.xml.namespace.QName;
+
+import org.jivesoftware.smack.util.XmlStringBuilder;
+
+/**
+ * Limits feature.
+ * For any XMPP stream, there is an "initiating entity" (a client or server) and a "responding entity" that they are connecting to.
+ * The responding entity advertises its limits in the <code>&lt;stream:features/&gt;</code> element that it sends at the start of the stream.
+ * <a href="https://xmpp.org/extensions/xep-0478.html">XEP-0478: Stream Limits Advertisement</a>
+ */
+public class Limits implements ExtensionElement {
+    public static final String ELEMENT = "limits";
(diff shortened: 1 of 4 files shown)


smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
```diff
@@ -0,0 +1,90 @@
+/*
+ *
+ * Copyright © 2014-2026 Florian Schmaus
```
Flowdalic, line 3: I don't think I have the copyright on this code, have I?

smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
```diff
@@ -0,0 +1,90 @@
...
+    public static final QName QNAME = new QName(NAMESPACE, ELEMENT);
+
+    /**
+     * The maximum size of any first-level stream elements (including stanzas),
+     * in bytes the announcing entity is willing to accept.
+     * Guidance on acceptable limits is provided in RFC 6120 section 13.12.
+     * If the responding entity is unable to determine its limits, this child can be absent.
+     * Element: <code>&lt;max-bytes/&gt;</code>. Type UnsignedInt.
+     */
+    public final int maxBytes;
```
Flowdalic, line 41: Use `UInt32` for `xs:unsignedInt`.

```suggestion
    public final UInt32 maxBytes;
```

smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * The number of seconds without any traffic from the initiating entity after which
+     * the server may consider the stream idle, and either perform liveness checks
+     * (using e.g. Stream Management (XEP-0198) or XMPP Ping (XEP-0199)) or terminate the stream.
+     * Guidance on handling idle connections is provided in RFC 6120 section 4.6.
+     * If the responding entity is unable to determine its limits, this child can be absent.
+     * Element: <code>&lt;idle-seconds/&gt;</code>. Type UnsignedInt.
+     */
+    public final int idleSeconds;
```
Flowdalic, line 51:
```suggestion
    public final UInt32 idleSeconds;
```

smack-core/src/main/java/org/jivesoftware/smack/packet/Limits.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public int getIdleSeconds() {
+        return idleSeconds;
+    }
+
+    @Override
+    public XmlStringBuilder toXML(XmlEnvironment enclosingNamespace) {
+        XmlStringBuilder xml = new XmlStringBuilder(this);
+        xml.rightAngleBracket();
+        if (maxBytes != 0) {
```
Flowdalic, line 80: Use

`xml.optElement()`

now, since we denote the absence of the element with `null`.

Flowdalic (changes requested):
XEP-0478 is an experimental extension and hence does not belong in `smack-core`, please move the code into `smack-experimental` and refactor it accordingly.

stokito:
I put it into the `core` because the stream features are parsed there and not sure how to make it to use classes from extensions. I'll back to the PR later, making it draft for now.

PR #693, opened 2026-01-27
[core] Presence: add isActive() to check if user is online and can answer

stokito:
The Spark has many places where we need to determine if user is currently online i.e. its status is available or free for chat.
There is even a separate method for this:
https://github.com/igniterealtime/Spark/blob/a64185ad3b4527f764dff680686fc262af48cac1/core/src/main/java/org/jivesoftware/spark/PresenceManager.java#L148
Still there are places in code where to detect it a user is active manually like this:
```
usersPresence.isAvailable() && !usersPresence.isAway()
```

I would like to propose a separate method `isActive` to simplify usage.

### DIFF
--- a/smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
+++ b/smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
@@ -107,7 +107,8 @@ public final class Presence extends MessageOrPresence<PresenceBuilder>
      * equivalent to <code>getType() == Presence.Type.available</code>. Note that even
      * when the user is available, their presence mode may be {@link Mode#away away},
      * {@link Mode#xa extended away} or {@link Mode#dnd do not disturb}. Use
-     * {@link #isAway()} to determine if the user is away.
+     * {@link #isAway()} to determine if the user is away, or {@link #isActive()}
+     * to determine if user is available an not away.
      *
      * @return true if the presence type is available.
      */
@@ -127,6 +128,19 @@ public final class Presence extends MessageOrPresence<PresenceBuilder>
      */
     public boolean isAway() {
         return type == Type.available && (mode == Mode.away || mode == Mode.xa || mode == Mode.dnd);
+    }
+
+    /**
+     * Returns true if the presence signals availability ({@link #isAvailable()}) and no away mode ({@link #isAway()}). Otherwise, returns false.
+     * mode is {@link Mode#available available}, or {@link Mode#chat ready to chat}. False will be returned when the type or mode
+     * is any other value, including when the presence type is unavailable (offline).
+     * This is a convenience method equivalent to
+     * <code>isAvailable() &amp;&amp; !isAway()</code>.
+     *
+     * @return true if the presence type is available and the presence mode is available or chat.
+     */
+    public boolean isActive() {
+        return isAvailable() && !isAway();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
```diff
@@ -107,7 +107,8 @@ public Presence(Presence other) {
      * equivalent to <code>getType() == Presence.Type.available</code>. Note that even
      * when the user is available, their presence mode may be {@link Mode#away away},
      * {@link Mode#xa extended away} or {@link Mode#dnd do not disturb}. Use
-     * {@link #isAway()} to determine if the user is away.
+     * {@link #isAway()} to determine if the user is away, or {@link #isActive()}
+     * to determine if user is active and can answer immideatelly.
```
Flowdalic: Let's not make promises that we can't keep :)

```suggestion
     * to determine if user is available an not away.
```

smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
```diff
@@ -129,6 +130,19 @@ public boolean isAway() {
...
+     * Returns true if the presence type is {@link Type#available available} and the presence
+     * mode is {@link Mode#available available}, or {@link Mode#chat ready to chat}. False will be returned when the type or mode
+     * is any other value, including when the presence type is unavailable (offline).
+     * This is a convenience method equivalent to
+     * <code>type == Type.available &amp;&amp; (mode == Mode.available || mode == Mode.chat)</code>.
+     *
+     * @return true if the presence type is available and the presence mode is available or chat.
+     */
+    public boolean isActive() {
+        return type == Type.available && (mode == Mode.available || mode == Mode.chat);
```
Flowdalic: This can be shortend:

```suggestion
        return isAvailable() && !isAway()
```

smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
```diff
@@ -129,6 +130,19 @@ public boolean isAway() {
         return type == Type.available && (mode == Mode.away || mode == Mode.xa || mode == Mode.dnd);
     }
 
+    /**
+     * Returns true if the presence type is {@link Type#available available} and the presence
+     * mode is {@link Mode#available available}, or {@link Mode#chat ready to chat}. False will be returned when the type or mode
+     * is any other value, including when the presence type is unavailable (offline).
+     * This is a convenience method equivalent to
+     * <code>type == Type.available &amp;&amp; (mode == Mode.available || mode == Mode.chat)</code>.
```
Flowdalic:
```suggestion
     * <code>isAvailable() &amp;&amp; !isAway()</code>.
```

smack-core/src/main/java/org/jivesoftware/smack/packet/Presence.java
```diff
@@ -129,6 +130,19 @@ public boolean isAway() {
         return type == Type.available && (mode == Mode.away || mode == Mode.xa || mode == Mode.dnd);
     }
 
+    /**
+     * Returns true if the presence type is {@link Type#available available} and the presence
```
Flowdalic:
```suggestion
     * Returns true if the presence signals availability ({@link #isAvailable()}) and no away mode ({@link #isAway()}). Otherwise, returns false.
```

Flowdalic (changes requested):

stokito:
I applied your suggestions and squashed into a single commit.

PR #694, opened 2026-01-29
[extensions] MultiUserChatManager.getRoomsHostedBy() don't fail if so…

stokito:
…me room JIDs are broken

If some room in a response has invalid JID then the entire call fails with an exception. Instead we may show a warning and continue process other rooms. The HostedRoom constructor trows an exception without the JID so it's complicated to trace it. Instead use asEntityBareJidOrThrow() that will include the JID to a message.

*Motivation*
This fixes a real problem in Spark: When getting list of rooms from the `chat.yax.im` it returns many room jids like `670@chat.yax.im` but one room is just `chat.yax.im` i.e. it has a jid without `@`.

### DIFF
--- a/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/HostedRoom.java
+++ b/smack-extensions/src/main/java/org/jivesoftware/smackx/muc/HostedRoom.java
@@ -16,8 +16,6 @@
  */
 package org.jivesoftware.smackx.muc;
 
-import org.jivesoftware.smack.util.Objects;
-
 import org.jivesoftware.smackx.disco.packet.DiscoverItems;
 
 import org.jxmpp.jid.EntityBareJid;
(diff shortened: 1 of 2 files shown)


Flowdalic:
That's the wrong abstraction layer. Otherwise we would need to do that in many more places in Smack.

If you want to be more relaxed with JID parsing, then setup jxmpp with a more relaxed stringprep implementation and use the according jxmpp context in Smack.

stokito:
The PR doesn't fix the parsing of the room jid, but makes it resilient for errors. If we have one broken room this won't break the entire room list call.

Flowdalic:
> The PR doesn't fix the parsing of the room jid, but makes it resilient for errors. If we have one broken room this won't break the entire room list call.

Yes, that is right. But again, if you want weaker JID verification use an appropriate jxmpp context.

PR #695, opened 2026-01-29
[extensions] MucConfigFormManager: support persistentroom and roomdesc

stokito:
This is often used in `muc#roomconfig` "basic" fields. At least they are used in the Spark.

https://xmpp.org/extensions/xep-0045.html


Flowdalic (changes requested):
LGTM, but please back out the integration test change.

We can add it later, but then with different strings for room name and description.

PR #697, opened 2026-02-19
Make ModularXmppClientToServerConnection non-final

MarcelHeckel:
We need to override the method `AbstractXMPPConnection.parseAndProcessStanza()` to add some functionality to the `XmlPullParser`. This is possible with `XMPPBOSHConnection` and `XMPPTCPConnection`. 

The same flexibility is needed  in `ModularXmppClientToServerConnection`, but it is final, so we cannot override the method. Personally, I see no good reason why this class is `final`.


Flowdalic:
Thanks for your contribution. Much appreciated.

> We need to override the method `AbstractXMPPConnection.parseAndProcessStanza()` to add some functionality to the `XmlPullParser`.

Could you elaborate a bit on what you are trying to achieve?

MarcelHeckel:
The main problem is that we can have large stanzas combined with slow internet connections.
This causes `getLastStanzaReceived()` to be too far in the past and the `PingManager` to close the connection. 
Therefore, we have an `XmlPullParser` implementation that delegates to the real `XmlPullParser` and sets `lastStanzaReceived` to the current timestamp every time an XML element is read. This way, the `PingManager` will not disconnect the connections, as the tags of the large stanza occur much more frequently than the end of the stanza.

Flowdalic:
Seems what you really need is `XMPPConnection.getLastDataReceived()` and `PingManager` using that. I'll have a look.

MarcelHeckel:
@Flowdalic I could make a implementation of `getLastDataReceived()` based on our solution, if you like.

Flowdalic:
> @Flowdalic I could make a implementation of `getLastDataReceived()` based on our solution, if you like.

See https://github.com/igniterealtime/Smack/pull/698

MarcelHeckel:
Looks good!

